PR #8, opened 2017-09-15
Remove unnecessary setting of element in results cache array

papiomytoglou:
Remove unnecessary setting of element in results cache array in the scenario the value has already been calculated:

https://github.com/ta4j/ta4j/blob/a9786edf0b5a92a353b65d52599dea10ea8ef715/ta4j/src/main/java/eu/verdelhan/ta4j/indicators/CachedIndicator.java#L103-L107

The statement in line 107 should be included in the body of the if clause so that the element in `resultInnerIndex` is not overwritten by itself when `result` is not null.


team172011:
that  really make sense. Good discovery :+1:

PR #18, opened 2017-09-17
establish changelog, add note in PR template

martinklepsch:
I've always been missing a CHANGELOG from this project. Sometimes releases just break everything (such as the recent BaseTimeSeries/TimeSeries split) and the only way to figure it out is by reading commit messages. Breakage is part of life and that's ok but we should be communicating it properly.

📺 &nbsp; [Related talk that I think presents some nice food for thought.](https://www.youtube.com/watch?v=oyLBGkS5ICk)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,4 +6,4 @@ Changes proposed in this pull request:
 - 
 - 
 
-
+- [ ] added an entry to the unreleased section of `CHANGES.md` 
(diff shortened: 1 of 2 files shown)


dpalic:
I have added yesterday independently in the wiki: https://github.com/ta4j/ta4j/wiki/How-to-contribute

martinklepsch:
There also is a contributing section in the README already. We could put your text there to have things in one place?

I've worked on a few projects where the separate versioning of wiki and repo caused them to get out of sync. Because of that I'm slightly tending towards putting more stuff in the repo and less into the wiki. 

What do you think about the addition of a changelog and this PR specifically?

team172011:
lets merge?

Issue #13, opened 2017-09-16
moving all packages to org.ta4j


dpalic:
I think we should move also all packages to org.ta4j to make the new library also findable on the long run via the webpage.
Also this makes the statement clearer, that the lib is living independently of a person.
I would suggest to switch these with version 0.11. What do you think?

team172011:
Yes, i think so, too. If we do so i would prefer org.ta4j.ta4j for this repository to keep open the possibility to create further repositories under this organization.

dpalic:
have not got your point. you want to have all packages to start with

    org.ta4j.ta4j

correct?
If so I would rather prefer something like 

    org.ta4j.lib

or similar. Maybe we find something more matching.
Wondering for your request. Are you planning to move your personal ta4j projects to this org?

team172011:
I mean this repository under org.ta4j.ta4j and other possible repositories than under org.ta4j.other. 
Depending on the long term goals of this organization (that we have first to define) i could imagine starting some new once like org.ta4j.charting or org.ta4j.api ...

martinklepsch:
How about org.ta4j.core to avoid the name doubling?

team172011:
Exactly that would be my suggestion! 
But for the first the name of this repository should stay "ta4j" to avoid missunderstandings. Maybe we should establish the time frame of the "transitional period "  from than on we start renaming and established our release process?

dpalic:
org.ta4j.core

Sounds good!

I think we should deliver a transition release to move to the new api packages. And this release contains only this change. We should also post this on the README.md along with multiple more resources.

team172011:
should we do this in release 0.10 or create an extra transition-release?

dpalic:
I am not sure, but since we are a 0.x version I think we can risk the bad responses of the persons who have to modify their code.

We should definitively add this as a big note into the Changelog file!

team172011:
finished in #52

PR #22, opened 2017-09-19
Feature/rename org ta4j and add DMI indicators

team172011:
Enhanced #27, #13.

- change package description to org.ta4j.core
- customize pom xml for new structure and maven
- Added DirectionalMovementUpIndicator (DMI+) and test classes 
- Added DirectionalMovementDownIndicator (DMI-) and test classes
- New package for ADX, DMI+ and DMI-


ta4j/pom.xml
```diff
@@ -1,11 +1,11 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
-        <groupId>eu.verdelhan</groupId>
+        <groupId>org.ta4j</groupId>
         <artifactId>ta4j-parent</artifactId>
         <version>0.10-SNAPSHOT</version>
     </parent>
-    <artifactId>ta4j</artifactId>
+    <artifactId>core</artifactId>
```
mdeverdelhan: Should rather be "ta4j-core", right?

martinklepsch: Is it bad practice to have a generic `artifactId` like `core` if the `groupId` sufficiently describes where it belongs?

mdeverdelhan: Maybe not. But it is not common. See:
```
<dependency>
    <groupId>org.apache.commons</groupId>
    <artifactId>commons-math3</artifactId>
    <version>3.6.1</version>
</dependency>
```
or:
```
<dependency>
    <groupId>com.google.code.gson</groupId>
    <artifactId>gson</artifactId>
    <version>2.8.1</version>
</dependency>
```

mdeverdelhan: And it seems weird if the parent is still "ta4j-parent".

team172011: Yes, i think it should not be ta4j-parent in the parent, but ta4j. I like also the suggestion to describe this repository with ta4j-core, because it is more understandable regarding the old description.
so we would have

->org.ta4j
--> org.ta4j.ta4j-core
--> org.ta4j.ta4j-examples

what do you think?

mdeverdelhan: I would have kept the parent as "ta4j-parent" because IMO it avoids ambiguity.
We would have:
groupId: org.ta4j
artifactIds:
-> ta4j-parent (i.e. we kept the "parent" word to avoid you directly add it as a dependency)
--> ta4j-core
--> ta4j-examples
--> ta4j-whatever

team172011:
also DMI+/DMI- inidcator and new package for ADX indicators merged into this branch, because i have accidentila create the feature/branch for here

Issue #9, opened 2017-09-15
Feature request - support for bars and ticks.


magick93:
Hello

It would be great if this product supported ticks, in their traditional sense, and bars.

A tick can be thought of as a moment in time - and in the case of FX - is a price change in the currency pair, whereas a bar is a timespan, typically of 1m, 5m, 15m, 30... 1h, 4h, 8h, 1d etc.

team172011:
Hi,

using the current Tick interface to represent a "FX-Tick" just with endTime is not possible?

magick93:
I dont know if its possible.

The point is, the name is incorrect.

A tick cannot  have an open, high, low and close.

In the same way a photo captures a moment in time, whereas a video captures a time span.

It wouldnt make sense to say use a photo that plays for 1 hour.

A tick is also used to represent the minimum upward or downward movement. 

There is no concept of timespan in the industry definition of the term 'tick'.

martinklepsch:
While this has been a longstanding confusion for people coming to this library it hasn't been recognized as something that really needs to be changed yet.

I think with the new org and potentially a more streamlined development process this probably should be considered again.

team172011:
Sure, for my part the current `Tick ` interface can be renamed in Bar. But the request from @magick93 to support ticks without open, high, low and close will be problematically because the most indicators of this library expect those fields

martinklepsch:
I guess there could be `Tick` and `Bar` and means to convert multiple `Tick`s into a `Bar`. Indicators etc would continue to always require bars. Not 100% sure if this should be part of ta4j but might be worth exploring in a separate repo for starters.

team172011:
I agree, a fx-tick would have no use in this repository. If you want to store just fx data you can do
`BaseTick fxTick = BaseTick(endTime,  Decmial.NaN, Decmial.NaN, Decmial.NaN, closePrice, Decmial.NaN)` and have to make sure that you just use indicator that just need closePrice for calculation. 

Renaming Tick to Bar would be useful plan, it think.

magick93:
> But the request from @magick93 to support ticks without open, high, low and close will be problematically because the most indicators of this library expect those fields

A bar is an aggregation of tick values. Indicators take bars. And a bar is an aggregation of ticks.

team172011:
Prototype implementation in branch https://github.com/ta4j/ta4j/tree/Prototype/Decimal-Interface-And_Columnar-TimeSeries (#88) renames **Tick** to **Bar**

PR #25, opened 2017-09-20
Fix AroonDownIndicator, AroonUpIndicator and NaN value support

### LINKED ISSUE #21: Handle Decimal.NaN values (especially in Aroon Up/Down indicator)
Regarding an old issue: https://github.com/mdeverdelhan/ta4j/issues/187 i think there is a important question still unanswered.

- **How to handle `NaN `values?**

**In my opinion an indicator should return `NaN `if the input values are `NaN`.** 

For example the `AroonUpIndicator `and `AroonDownIndicator `returns zero instead of `NaN`. But zero is a admissible value for this indicator and would be interpreted totally different as a "Not a number" information.

team172011:
Fixes #21.

- AroonUpIndicator calculates correct values and supports NaN values
- AroonDownIndicator calculates correct values and supports NaN values
- HighestValueIndicator jumps over NaN if they are at current index
- LowestValueIndicator jumps over NaN if they are at current index

**Aroon Indicators** calculate now correct values.
Also work related the **LowestValueIndicator and HighestValueIndicator** "jump" over `NaN `values if they are at the current index. (like the have done it before if they where within the time frame intervall)

E.g.
`p(1,2,3,4,NaN)` ->` LowestValueIndicator(p, 5)` would return `1` and not `NaN`
`p(1,2,3,4,NaN)` -> `HighestValueIndicator(p, 5)` would return `4` and not `NaN`


nimo23:
AroonUp/Down" had used the **"ClosePriceIndicator"** in the previous version. This is not indeed bad: user can react only after getting the last tick with its close price. So reacting by AroonUp/Down can limit the risk for ups and downs when using the nearest price (which is the "close price"). I find, it is a reasonable tweak to use LowestClosePriceIndicator instead of MinPriceIndicator (the same goes for HighestClosePriceIndicator and MinPriceIndicator in ArronUp). Maybe the author did use the ClosePriceIndicator with a same reason. To leave the tweak within the next version, we can add a optional constructor where user can decide which PriceIndicator to use:


```
/**
 * Aroon down indicator.
 * <p>
 */
public class AroonDownIndicator extends CachedIndicator<Decimal> {

    private final int timeFrame;

    private final LowestValueIndicator lowestPriceIndicator;
    private final Indicator<Decimal> price;
    
    public AroonDownIndicator(TimeSeries series, Indicator<Decimal> price, int timeFrame) {
    	super(series);
        this.timeFrame = timeFrame;
        this.price = price;

        // + 1 needed for last possible iteration in loop
        lowestPriceIndicator = new LowestValueIndicator(price, timeFrame+1);
    }

    /**
     * default constructor with MinPriceIndicator
     * 
     * @param series
     * @param timeFrame
     */
    public AroonDownIndicator(TimeSeries series, int timeFrame) {
        this(series, new MinPriceIndicator(series), timeFrame);
    }

    @Override
        if (getTimeSeries().getTick(index).getMaxPrice().isNaN())
            return Decimal.NaN;

        // Getting the number of ticks since the lowest close price
        int endIndex = Math.max(0,index - timeFrame);
        int nbTicks = 0;
        for (int i = index; i > endIndex; i--) {
            if (price.getValue(i).isEqual(lowestPriceIndicator.getValue(index))) {
                break;
            }
            nbTicks++;
        }
        
        return Decimal.valueOf(timeFrame - nbTicks).dividedBy(Decimal.valueOf(timeFrame)).multipliedBy(Decimal.HUNDRED);
    }
}
```

The same goes for "Aroon up indicator."

Please review this version and if reasonable, change your pull request.

team172011:
The Formula says "highest high" and "highest low" at all sources i know. Do you know any provider that calculations are based on close price?

"using the nearest price (which is the "close price")"

Do not know what you mean. If the user has the close price he also has the min and max price of the tick?

nimo23:
> If the user has the close price he also has the min and max price of the tick?

yes, but reacting after close price is faster than reacting after high price, high price can happen within the duration of the tick. Close Price is at the end of the duration. Please let the user decide which price to take within the indicator. I dont see any cons to provide this with a optional constructor. The default constructor, as you see, does use the higest/lowesValueIndicator. There are other tweaks around this lib, for example RAVI (the official indicator says, it should use ABS, however ta4j does not use abs, which is ok).

nimo23:
For example: Imagine AroonUp=100 and "highest close price" is used instead of the "highest high price", then user can be for sure that when he decide to buy/sell, the last price is the highest price within the period. Using "highest high price" instead, this information is lost.

team172011:
> yes, but reacting after close price is faster than reacting after high price, high price can happen within the duration of the tick. Close Price is at the end of the duration and there can be a time lag between the high price and the close price

I think you are wrong with this assessment. If you want to trade and analyse you have to define a time period or the "tick size". Close (latest), high and low price are information that are **all at exact the same time available** and that is if the period ends (e.g. weekly tick means that close, high and low are available on friday after close of the maket and not one second earlier for any price type).

In my opinion this lib should provide indicators in its general form. For Aroon Up/Down it is definitly based on high and low prices. If the user needs customization he can do this in his own source code, if we start adding additional uncommon constructors to every indicator like here it becomes more and more difficult to maintain them all.

nimo23:
`Close (latest), high and low price are information that are all at exact the same time available`

yes, this is not the problem! I have a well formed Tick (Bar). Look at Moving Average or other indicators (ROC,..), these indicators ALL have an **additional constructor so the user can set the kind of price to use** (MaxPriceIndicator, MedianPriceIndicator, MinPriceIndicator, OpenPriceIndicator, TypicalPriceIndicator). For Indicators which only uses one "kind of price", it is a **benefit to fine tune the indicator.** For example, in traderview I can select which kind of price to use in the calculation for specific indicators (open, high, low or close). So it is a common thing for all indicators where only one kind of price is used.

`if we start adding additional uncommon constructors to every indicator like here it becomes more and more difficult to maintain them all.`

Yes, you are right. It would only be the case in these indicators where only one kind of price is used (for example, all moving averages, roc,..). Feel free, to disable this.

nimo23:
RSIIndicator is another example, it has the option **wich kind of price to use** for averageGain/averageLoss. So you see, the "kind of price"-selection is already in a wide range of common ta4j-indicators.

nimo23:
another example is in CoppockCurveIndicator with a constructor to enable "kind of price"-selection.

```
@param indicator the indicator (usually close price)
public CoppockCurveIndicator(Indicator<Decimal> indicator,..}
```

The benefit to have this is that we can use this indicator where only closePrice is available, for example, for a typical "Tick" https://github.com/ta4j/ta4j/issues/9

`BaseTick fxTick = BaseTick(endTime, Decmial.NaN, Decmial.NaN, Decmial.NaN, closePrice, Decmial.NaN)`

AroonUp/Down and all other indicators is also usable for fxTick if "kind of price"-selection exist.

The origin author has done good by providing this in the lib.

Issue #27, opened 2017-09-22
Values of ADX/DMI+/DMI- incorrect? 


nimo23:
I compared the values of ADX/DMI+/DMI- with 2 other chart-software and it seems either of one returns wrong values. So please check out, if ta4j really returns the right values.

For example,

ta4j returns this (I rounded the returned values):

```
AverageDirectionalMovementIndicator=1.4131
AverageDirectionalMovementUpIndicator=0.1937
AverageDirectionalMovementDownIndicator=0.1808
DirectionalMovementIndicator timeFrame: 12=3.4463
DirectionalMovementUpIndicator=0.04
DirectionalMovementDownIndicator=0.0
```

But the Chart-Software with the same timeframe on the same "tick" returns these values:

```
ADX=35,126
PDI=13,323
MDI=36,644
```

And other ticks have also completly different values. So I am in doubt, which returns the right values. Please double-check this out.

team172011:
Can you give a concrete example, please? Which time frame and other parameter do you use on which data? What is your Chart-Software you compare to?

nimo23:
I used this to check the values:

```
this.adx = new AverageDirectionalMovementIndicator(series, 12);
this.adxUp = new AverageDirectionalMovementUpIndicator(series, 12);
this.adxDown = new AverageDirectionalMovementDownIndicator(series, 12);
this.dmi = new DirectionalMovementIndicator(series, 12);
this.dmiUp = new DirectionalMovementUpIndicator(series);
this.dmiDown = new DirectionalMovementDownIndicator(series);
```

and compared the values with the values of the chart from http://www.teletrader.com/ (it is free to use).

nimo23:
Select stock, go to tab "chart"->Directional Movement System.

team172011:
Yes, there seems to be a bug in calculation. But i have to take a deeper look later

team172011:
I have checked the calculations and they are all correct. They are based on the procedure on [StockCharts](http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:average_directional_index_adx)
If you want to get the **Directional Movement System** like on [teletrader](http://www.teletrader.com/coca-cola-bottling-co-consolidated/stocks/chart/tts-832065) you have to calculate the PDI and MDI based on `AverageTrueRangeIndicator `and the corresponding `AverageDirectionalMovementUp/DownIndicator`:

````
//series = coke, daily, 2017-07-18 - 2017-09-21
// ADX
[...]

PR #31, opened 2017-09-23
Added DirectionalMovementPlus and -MinusIndicator to new package for adx

team172011:
in issue #27 it became clear that adx/admi was a little bit hard to use
because of the an alternative calculation.
I have not changed the calculation but added two new indicators with
test clases, that are usually used in combination with the adx.
Additional the established indicators were moved to the new package
"adx" to make clear where to find them and how they are to use in this
library

Fixes #.

- 
- 
- 


ta4j/src/main/java/org/ta4j/core/indicators/adx/DirectionalMovementMinusIndicator.java
```diff
@@ -0,0 +1,60 @@
...
+    public DirectionalMovementMinusIndicator(TimeSeries series, int timeFrame) {
+        super(series);
+        this.timeFrame = timeFrame;
+        this.averageDirectionalMovementDownIndicator = new AverageDirectionalMovementDownIndicator(series, timeFrame);
+        this.trueRangeIndicator = new AverageTrueRangeIndicator(series, timeFrame);
+    }
+
+    @Override
+    protected Decimal calculate(int index) {
+        return (averageDirectionalMovementDownIndicator.getValue(index).dividedBy(trueRangeIndicator.getValue(index))).multipliedBy(Decimal.HUNDRED);
```
nimo23, line 53: No need for the outer enclosing brackets.

ta4j/src/main/java/org/ta4j/core/indicators/adx/DirectionalMovementMinusIndicator.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private final AverageDirectionalMovementDownIndicator averageDirectionalMovementDownIndicator;
+    private final AverageTrueRangeIndicator trueRangeIndicator;
+    private final int timeFrame;
+
+    public DirectionalMovementMinusIndicator(TimeSeries series, int timeFrame) {
+        super(series);
+        this.timeFrame = timeFrame;
+        this.averageDirectionalMovementDownIndicator = new AverageDirectionalMovementDownIndicator(series, timeFrame);
+        this.trueRangeIndicator = new AverageTrueRangeIndicator(series, timeFrame);
```
nimo23, line 48: Why not using "TrueRangeIndicator.class" (within helpers-package) instead of "AverageTrueRange"? According to Wilder, DM+/DM- is calculated by "TrueRange" and not by "AverageTrueRange" (which gives it even more lag). 

```
+DI = ( +DM / TR ) * 100
-DI = ( -DM / TR ) * 100
```


Look formula at https://www.tradingtechnologies.com/help/x-study/technical-indicator-definitions/average-directional-movement-adx/. However, I am in doubt which is better..

team172011: Because the values are correct compared to other providers and because the calculations of your side does exact the same as i do in the "Average-" classes:

https://www.tradingtechnologies.com/help/x-study/technical-indicator-definitions/average-directional-movement-adx/:
The +DM, -DM and TR are each accumulated and smoothed using a custom smoothing method proposed by Wilder. For an n period smoothing, 1/n of each period’s value is added to the total each period, similar to an exponential smoothing:

team172011: You have to accumulate and smoothe the true range and the DMIs. Also mentiond in http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:average_directional_index_adx

PR #38, opened 2017-09-27
Introduce "IsEqualRule"

nimo23:
- New IsEqualRule<Decimal> that is satisfied if value of indicator1 equals value of indicator2

### DIFF
--- a/ta4j/src/test/java/org/ta4j/core/trading/rules/IsEqualRuleTest.java
+++ b/ta4j/src/test/java/org/ta4j/core/trading/rules/IsEqualRuleTest.java
@@ -1,0 +1,55 @@
+/**
+ * The MIT License (MIT)
+ *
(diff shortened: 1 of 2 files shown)


ta4j/src/test/java/org/ta4j/core/trading/rules/IsEqualRuleTest.java
```diff
@@ -0,0 +1,48 @@
...
+    @Before
+    public void setUp() {
+        indicator = new FixedDecimalIndicator(20);
+        rule = new IsEqualRule(indicator, Decimal.valueOf(20));
+    }
+    
+    @Test
+    public void isSatisfied() {
+        assertTrue(rule.isSatisfied(0));
+    }
```
mdeverdelhan: Please add an assertFalse(...) statement.

PR #41, opened 2017-10-01
Introduce "InSlopeRule"

nimo23:
Fixes #.

- Introduce "InSlopeRule"

### DIFF
--- a/ta4j/src/main/java/org/ta4j/core/indicators/RAVIIndicator.java
+++ b/ta4j/src/main/java/org/ta4j/core/indicators/RAVIIndicator.java
@@ -34,7 +34,6 @@ public class RAVIIndicator extends CachedIndicator<Decimal> {
 
     private final SMAIndicator shortSma;
     private final SMAIndicator longSma;
-    private final boolean toAbs;
    
     /**
      * Constructor.
@@ -46,7 +45,6 @@ public class RAVIIndicator extends CachedIndicator<Decimal> {
         super(price);
         shortSma = new SMAIndicator(price, shortSmaTimeFrame);
         longSma = new SMAIndicator(price, longSmaTimeFrame);
-        this.toAbs = false;
     }
     
     /**
@@ -54,23 +52,19 @@ public class RAVIIndicator extends CachedIndicator<Decimal> {
      * @param price the price
      * @param shortSmaTimeFrame the time frame for the short SMA (usually 7)
      * @param longSmaTimeFrame the time frame for the long SMA (usually 65)
-     * @param toAbs if true, absolute value of indicator is returned
      */
-    public RAVIIndicator(Indicator<Decimal> price, int shortSmaTimeFrame, int longSmaTimeFrame, boolean toAbs) {
+    public RAVIIndicator(Indicator<Decimal> price, int shortSmaTimeFrame, int longSmaTimeFrame) {
         super(price);
         shortSma = new SMAIndicator(price, shortSmaTimeFrame);
         longSma = new SMAIndicator(price, longSmaTimeFrame);
-        this.toAbs = toAbs;
     }
 
     @Override
     protected Decimal calculate(int index) {
         Decimal shortMA = shortSma.getValue(index);
         Decimal longMA = longSma.getValue(index);
-        Decimal val = shortMA.minus(longMA)
+        return shortMA.minus(longMA)
                 .dividedBy(longMA)
                 .multipliedBy(Decimal.HUNDRED);
-        
-        return toAbs ? val.abs() : val;
     }
 }
--- a/ta4j/src/main/java/org/ta4j/core/indicators/RAVIIndicator.java
+++ b/ta4j/src/main/java/org/ta4j/core/indicators/RAVIIndicator.java
@@ -27,13 +27,15 @@ import org.ta4j.core.Indicator;
 
 /**
  * Chande's Range Action Verification Index (RAVI) indicator.
+ * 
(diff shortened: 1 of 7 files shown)


nimo23:
If I have something like "AverageDirectionalMovementIndicator" and want to have a rule, which is satisfied, when the **slope** between the ADX(1) and ADX(n) has at least a minimum slope.

`return new InSlopeRule(value, Decimal.valueOf(0.2));`

I have **not** used the absolute value of the difference of the two indicator values, so we can handle negative slopes. For example, a rule which is satisfied, when the slope between the ADX(1) and ADX(nthPrevious=3) is negative (downtrend) and has at least a minimum (negative) slope of "-0.25" and a maximum (negative) slope of "-0.5":

`return new InSlopeRule(value, 3, Decimal.valueOf(-0.25), Decimal.valueOf(-0.5));`

nimo23:
Accidentally made a PR with a new version of "RAVIIndicator.java". Feel free to delete this. I dont know how to do that by myself?

team172011:
Will not be merged:

- Please only make PR to `develop `branch, you want to merge the current `develop `branch with your additional commits to the `master `of this rep
- Please reduce your commits. (as little as possible, as much as necessary)
- Please added an entry to the unreleased section of the `changelog.md`

nimo23:
How is the naming convention? "InSlopeRule" or "SlopeRule" or "SlopeIndicatorRule" ? 

Shorter is better..

ta4j/src/main/java/org/ta4j/core/trading/rules/InSlopeRule.java
```diff
@@ -0,0 +1,103 @@
...
+        this.minSlope = minSlope;
+        this.maxSlope = maxSlope;
+    }
+
+   @Override
+   public boolean isSatisfied(int index, TradingRecord tradingRecord) {
+	DifferenceIndicator diff = new DifferenceIndicator(ref, prev);
+	Decimal val = diff.getValue(index);
+	boolean minSlopeSatisfied = minSlope.isNaN() ? true : val.isGreaterThanOrEqual(minSlope);
+	boolean maxSlopeSatisfied = maxSlope.isNaN() ? true : val.isLessThanOrEqual(maxSlope);
```
team172011, line 97: if `minSlope.isNan() && maxSlope.isNan()` -> `true` the rule will be satisfied. I think it should be not satisfied if NaN?

ta4j/src/main/java/org/ta4j/core/indicators/RAVIIndicator.java
```diff
@@ -44,15 +46,31 @@ public RAVIIndicator(Indicator<Decimal> price, int shortSmaTimeFrame, int longSm
...
+    }
+    
+    /**
+     * Constructor.
+     * @param price the price
+     * @param shortSmaTimeFrame the time frame for the short SMA (usually 7)
+     * @param longSmaTimeFrame the time frame for the long SMA (usually 65)
+     * @param toAbs if true, absolute value of indicator is returned
+     */
+    public RAVIIndicator(Indicator<Decimal> price, int shortSmaTimeFrame, int longSmaTimeFrame, boolean toAbs) {
```
team172011: Not in the scope of this PR. If the use wants the absolute value he can use the `AbsoluteIndicator`

ta4j/src/main/java/org/ta4j/core/trading/rules/InSlopeRule.java
```diff
@@ -95,8 +95,9 @@ public boolean isSatisfied(int index, TradingRecord tradingRecord) {
 	Decimal val = diff.getValue(index);
 	boolean minSlopeSatisfied = minSlope.isNaN() ? true : val.isGreaterThanOrEqual(minSlope);
 	boolean maxSlopeSatisfied = maxSlope.isNaN() ? true : val.isLessThanOrEqual(maxSlope);
+	boolean isNaN = minSlope.isNaN() && maxSlope.isNaN();
```
team172011: Dont know if this makes sense. As far as i understand the logic should be as follows:

- If `minSlope.isNaN()` ->  `cannot use this rule` >> `minSlopeSatisfied=false`
- If `maxSlope.isNaN()`  -> `do not consider upper bound` -> `maxSlopeSatisfied=true`

so it would be enough to change line 96 to:
`boolean minSlopeSatisfied = minSlope.isNaN() ? false: val.isGreaterThanOrEqual(minSlope);`

nimo23: No, it is correct. 

As shown in the constructor, both "minSlope" and "maxSlope" are optional values. However, at least one of those (or both, if both are given) must be a number. Imagine 3 cases: 

1) I want to test if maxSlope between a and b is 0.7, I do not care about minSlope. 
2) I want to test if minSlope between a and b is 0.2, I do not care about maxSlope.
3) I want to test if minSlope is -0.2 and maxSlope is 0.7, so both params must be a number.

team172011: okay i see

nimo23:
Wanted to rebase with https://github.com/blog/2243-rebase-and-merge-pull-requests. Cannot find it. How can I do that afterwards?

team172011:
If you remove modifications on RAVIIndicator i will try to merge

nimo23:
Okay, I only have updated the javadoc for RAVI and revert other changes to the original.

team172011:
Alright, thank you!

PR #44, opened 2017-10-04
Removed unnecessary last tick index calculation

katsadim:
Removed unnecessary last tick index calculation if the ticks list is empty:

https://github.com/ta4j/ta4j/blob/d699fe77a21a1bdcd99a320a444f79a3c297d131/ta4j/src/main/java/eu/verdelhan/ta4j/BaseTimeSeries.java#L213-L219

The statement in line 213 should be included in the body of the if clause, so that in case of an empty tick list we won't have a redundant last tick index calculation.


team172011:
You are right.  The last index calculation is only needed within the if clause and should happem there

PR #48, opened 2017-10-08
New Rules: Falling/Rising/Highest/Lowest-IndicatorRules

nimo23:

### DIFF
--- a/ta4j/src/main/java/org/ta4j/core/indicators/SmoothedRSIIndicator.java
+++ b/ta4j/src/main/java/org/ta4j/core/indicators/SmoothedRSIIndicator.java
@@ -61,8 +61,4 @@ public class SmoothedRSIIndicator extends RSIIndicator {
         return super.calculate(index);
     }
     
-    @Override
-    public String toString() {
-        return getClass().getSimpleName() + " timeFrame: " + timeFrame;
-    }
 }
--- a/ta4j/src/main/java/org/ta4j/core/indicators/SmoothedRSIIndicator.java
+++ b/ta4j/src/main/java/org/ta4j/core/indicators/SmoothedRSIIndicator.java
@@ -61,4 +61,8 @@ public class SmoothedRSIIndicator extends RSIIndicator {
         return super.calculate(index);
     }
     
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + " timeFrame: " + timeFrame;
+    }
 }
--- a/ta4j/src/main/java/org/ta4j/core/indicators/SmoothedRSIIndicator.java
+++ b/ta4j/src/main/java/org/ta4j/core/indicators/SmoothedRSIIndicator.java
@@ -60,9 +60,4 @@ public class SmoothedRSIIndicator extends RSIIndicator {
         }
         return super.calculate(index);
     }
-    
-    @Override
-    public String toString() {
-        return getClass().getSimpleName() + " timeFrame: " + timeFrame;
-    }
 }
--- a/ta4j/src/main/java/org/ta4j/core/indicators/SmoothedRSIIndicator.java
+++ b/ta4j/src/main/java/org/ta4j/core/indicators/SmoothedRSIIndicator.java
@@ -60,4 +60,9 @@ public class SmoothedRSIIndicator extends RSIIndicator {
         }
         return super.calculate(index);
     }
+    
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + " timeFrame: " + timeFrame;
+    }
 }
--- a/ta4j/src/main/java/org/ta4j/core/trading/rules/IsFallingRule.java
+++ b/ta4j/src/main/java/org/ta4j/core/trading/rules/IsFallingRule.java
@@ -33,24 +33,26 @@ import org.ta4j.core.trading.rules.UnderIndicatorRule;
 /**
  * Indicator-falling-indicator rule.
  * <p>
- * Satisfied when the values of the {@link Indicator indicator} strict decrease within the previous (n-th) values.
+ * Satisfied when the values of the {@link Indicator indicator} strict decrease
+ * within the previous (n-th) values.
  */
 public class IsFallingRule extends AbstractRule {
 
 	/** The actual indicator */
-    private Indicator<Decimal> ref;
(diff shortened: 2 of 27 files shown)


ta4j/src/main/java/org/ta4j/core/trading/rules/IsFallingRule.java
```diff
@@ -0,0 +1,70 @@
...
+	/** The actual indicator */
+	private Indicator<Decimal> ref;
+	/** The previous n-th value of ref */
+	private int nthPrevious;
+
+	/**
+	 * Constructor.
+	 * 
+	 * @param ref
+	 * @param nthPrevious
```
team172011: Please maintain javadoc. Especially for this parameter it is important that the user knows that the `nthPrevious ` parameter means the `nthPrevious` values. Because the `timeFrame` parameter of other indicators means `index-(timeFrame-1)`

ta4j/src/main/java/org/ta4j/core/trading/rules/IsHighestRule.java
```diff
@@ -0,0 +1,70 @@
...
+	public IsHighestRule(Indicator<Decimal> ref, int nthPrevious) {
+		this.ref = ref;
+		this.nthPrevious = nthPrevious;
+	}
+
+	@Override
+	public boolean isSatisfied(int index, TradingRecord tradingRecord) {
+		Rule gt = new OverIndicatorRule(ref, new PreviousValueIndicator(ref));
+		if (nthPrevious > 1) {
+			for (int i = 1; i < nthPrevious; i++) { // i must start again at 1.
```
team172011: Why not:
```
HighestValueIndicator highestValue = new HighestValueIndicator(ref, nthPrevious);
final boolean satisfied = this.ref.getValue(index).equals(highestValue.getValue(index));
```
?

ta4j/src/main/java/org/ta4j/core/trading/rules/IsLowestRule.java
```diff
@@ -0,0 +1,70 @@
...
+	public IsLowestRule(Indicator<Decimal> ref, int nthPrevious) {
+		this.ref = ref;
+		this.nthPrevious = nthPrevious;
+	}
+
+	@Override
+	public boolean isSatisfied(int index, TradingRecord tradingRecord) {
+		Rule lt = new UnderIndicatorRule(ref, new PreviousValueIndicator(ref));
+		if (nthPrevious > 1) {
+			for (int i = 1; i < nthPrevious; i++) { // i must start again at 1.
```
team172011: Why not:
LowestValueIndicator highestValue = new LowestValueIndicator(ref, nthPrevious);
final boolean satisfied = this.ref.getValue(index).equals(highestValue.getValue(index));
```
?

File: ta4j/src/main/java/org/ta4j/core/trading/rules/IsFallingRule.java (hunk 1 above)
Comment (team172011) [reply ta4j/src/main/java/org/ta4j/core/trading/rules/IsFallingRule.java]: I would suggest to rename this parameter to `timeFrame` and handle it like the `timeFrame` parameter in indicators. It makes it easier to compare and implement.

File: ta4j/src/test/java/org/ta4j/core/trading/rules/IsFallingRuleTest.java [hunk 4]
```diff
@@ -0,0 +1,61 @@
...
+	public void setUp() {
+		indicator = new FixedDecimalIndicator(6, 5, 4, 3, 2, 1, 0, -1, 2, 3);
+		rule = new IsFallingRule(indicator, 3);
+	}
+
+	@Test
+	public void isSatisfied() {
+		// The first 3 values evaluates to false,
+		// because indicator needs at least 
+		// 3 previous values to decide (unstable period).
```
team172011: If rename and use like a `timeFrame` parameter the Rule needs 2 previous values to decide (3 means current index and 2 previous indices)

nimo23:
I changed all to "timeFrame". 

For IsFalling/IsRisingRule, timeFrame=2 is the same as timeFrame=1. If you know a shorter, faster way, let me know.

team172011:
Yes, it can be a little bit confusing with the `timeFrame` parameter but as far as i could see this is the standard behavior.

nimo23:
I get NPE with IsHighestRule/IsLowestRule on StochasticRSIIndicator-values, because of this:

```
HighestValueIndicator highest = new HighestValueIndicator(ref, timeFrame);
Decimal highestVal = highest.getValue(index);
Decimal refVal = ref.getValue(index);
final boolean satisfied = refVal != null && highestVal != null && highestVal.isNaN()
// this makes trouble				
&& refVal.equals(highestVal);
```

I do tests..

nimo23:
Coincidentally, I tested **IsHighestRule** against "**StochasticRSIIndicator**":
```
ClosePriceIndicator price = new ClosePriceIndicator(series);
StochasticRSIIndicator indicator = new StochasticRSIIndicator(closePrice, 14);
Rule isHighest = new IsHighestRule(indicator, 3);
```
and it returns a NPE:

```

Caused by: java.lang.NullPointerException
	at java.math.BigDecimal.compareTo(BigDecimal.java:2625)
	at eu.verdelhan.ta4j.Decimal.equals(Decimal.java:417)
	at io.plan.ext.rule.IsHighestRule.isSatisfied(IsHighestRule.java:66)
```

**I guess, there is something wrong with StochasticRSIIndicator as it can return "null"-values instead of "Decimal.NaN".** Please check out "StochasticRSIIndicator". 

As all other indicators (which I tested so far) works with "IsHighestRule", the problem seems to be in "StochasticRSIIndicator" or in the Decimal.class?

nimo23:
okay, now all rules works.

ta4j/src/main/java/org/ta4j/core/indicators/SmoothedRSIIndicator.java
```diff
@@ -61,8 +61,4 @@ protected Decimal calculate(int index) {
         return super.calculate(index);
     }
     
-    @Override
-    public String toString() {
-        return getClass().getSimpleName() + " timeFrame: " + timeFrame;
-    }
```
team172011: Please override `toString()` function here

PR #65, opened 2017-10-29
added code and unit test

### LINKED ISSUE #64: enhance "JustOnceRule"
**JustOnceRule** should assert that a rule is fired only one time and then never again. It should behave like the rule for "CrossedUp/CrossedDownRule" (which only fires one time as soon as the rule is satisfied and then never again).

However, I cannot put a rule within JustOnceRule, which I want to fire only one time. So for example, this is actually not possible:

```
// Fires only one time as soon as the actual value is higher than its two previous values.
// Even if the "next value" is also higher than its 2 previous values, 
// the rule is no more fired, because of JustOnceRule.
new JustOnceRule(isRisingRule(2));
```

Would be nice to enhance the "JustOnceRule" to made the above example possible.

TheCookieLab:
Fixes #64 

- Add alternate constructor that takes in an "inner" rule

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/trading/rules/JustOnceRule.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/trading/rules/JustOnceRule.java
@@ -22,6 +22,7 @@
  */
 package org.ta4j.core.trading.rules;
 
+import org.ta4j.core.Rule;
 import org.ta4j.core.TradingRecord;
 
 /**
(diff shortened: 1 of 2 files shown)


team172011:
Cann you add an entry to the CHANGELOG.md and make a PR to `develop` branch, please?

TheCookieLab:
Ah my mistake. I just now got my new ta4j core repo set up. I should've taken some time to read over the contribution rules.

team172011:
No worries, i am pleased about every contribution!

TheCookieLab:
How should I go about updating the CHANGELOG.md - do I create a new 0.11 section?

PR #67, opened 2017-10-29
code complete with updated unit tests

### LINKED ISSUE #64: enhance "JustOnceRule"

TheCookieLab:
Fixes #64

- Added new constructor in JustOnceRule which takes in an "inner" Rule. If specified, JustOnceRule will return true the first time the inner rule returns true

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/trading/rules/JustOnceRule.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/trading/rules/JustOnceRule.java
@@ -1,24 +1,24 @@
-/*
-  The MIT License (MIT)
-
-  Copyright (c) 2014-2017 Marc de Verdelhan & respective authors (see AUTHORS)
-
-  Permission is hereby granted, free of charge, to any person obtaining a copy of
-  this software and associated documentation files (the "Software"), to deal in
-  the Software without restriction, including without limitation the rights to
-  use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
-  the Software, and to permit persons to whom the Software is furnished to do so,
-  subject to the following conditions:
-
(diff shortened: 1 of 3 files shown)


ta4j-core/src/main/java/org/ta4j/core/trading/rules/JustOnceRule.java
```diff
@@ -1,32 +1,32 @@
...
 
 import org.ta4j.core.TradingRecord;
 
 /**
  * A one-shot rule.
- * <p></p>
+ * <p>
  * Satisfied the first time it's checked then never again.
  */
 public class JustOnceRule extends AbstractRule {
```
team172011, line 32: Where is the new constructor?

team172011: I think something went wrong, rule and test should look like your PR from #65

team172011:
Done with PR  #72

team172011:
[Reopened] Accidental closed this PR

team172011:
[Closed] Already pushed to develop branch

PR #69, opened 2017-10-30
IsPositiveDivergentRule/IsNegativeDivergentRule

nimo23:
- New Rule: IsPositiveDivergentRule
- New Rule: IsNegativeDivergentRule

### DIFF
--- a/IsPositiveDivergentRule.java
+++ b/IsPositiveDivergentRule.java
@@ -36,19 +36,13 @@ import org.ta4j.core.trading.rules.AbstractRule;
  * Satisfied when the values of the {@link Indicator indicator} strict increase
  * or is the highest or is not the lowest within the timeFrame and the values of
  * the other {@link Indicator indicator} strict decrease or is the lowest or is
- * not the highest within a timeFrame. It can test both, strict and non-strict
- * divergence.
+ * not the highest within a timeFrame. In short: "other" makes lower lows while
+ * "ref" makes higher lows. It can test both, strict and non-strict divergence.
  */
 public class IsPositiveDivergentRule extends AbstractRule {
 
-	/** The actual indicator */
-	private Indicator<Decimal> ref;
-	/** The other indicator */
-	private Indicator<Decimal> other;
-	/** The timeFrame */
-	private int timeFrame;
-	/** Test for strict divergence. */
-	private boolean useStrictDivergence;
+	/** The rule for positive divergence */
+	private final Rule isPositiveDivergent;
 
 	/**
 	 * Constructor. <br/>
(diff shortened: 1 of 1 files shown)


nimo23:
Please review my code and check if all things are logically okay.

I do not follow the common interpretation of pos/neg divergence but use a more rigid (strict divergence) and more weak (with "refIsNotHighest, otherIsNotLowest on negative divergence", refIsNotLowest, otherIsNotHighest on positive divergence) interpretation. 

Maybe the logical structur can be improved (xor instead of or?)

nimo23:
I guess, it is better to rename the 2 files to:

- "IsDivergencePositiveRule.java"
- "IsDivergenceNegativeRule.java"

IsNegativeDivergentRule.java
```diff
@@ -0,0 +1,111 @@
...
+	 * Constructor. <br/>
+	 * 
+	 * If "useStrictDivergence" is enabled, divergence is satisfied if values of
+	 * "ref" strict decrease while the values of the "other" indicator strict
+	 * increase.
+	 * 
+	 * @param ref
+	 * @param other
+	 * @param timeFrame
+	 * @param useStrictDivergence
```
team172011: Please, add at least a short description for the javadoc params. E.g.
```
@param ref the indicator
@param other the other indicator
..
```

IsNegativeDivergentRule.java
```diff
@@ -0,0 +1,111 @@
...
+		} else {
+
+			Rule refIsLowest = new IsLowestRule(ref, timeFrame);
+			Rule otherIsHighest = new IsHighestRule(other, timeFrame);
+			
+			Rule refIsNotHighest = new IsHighestRule(ref, timeFrame).negation();
+			Rule otherIsNotLowest = new IsLowestRule(other, timeFrame).negation();
+
+			isNegativeDivergent = (refIsFalling.or(refIsLowest).or(refIsNotHighest))
+					.and(otherIsRising.or(otherIsHighest).or(otherIsNotLowest));
```
team172011, line 75: - Case: Both indicators are rising until previous index, with current index both indicators are falling so that they are lower than the value of previous index but higher than the other values of the rising indices. Would be no negative divergence, but` refIsNotHighest = true ` and `otherIsNotLowest = true `

IsNegativeDivergentRule.java
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		final boolean satisfied = isNegativeDivergent.isSatisfied(index, tradingRecord);
```
team172011: `index `paramter is just needed here. Maybe you could make case differentiation and initialization of `isNegativeDivergent ` rule in the constructor?

IsNegativeDivergentRule.java
nimo23: **Is this the right way?:**

`isNegativeDivergent = (refIsFalling.or(refIsLowest)).and(otherIsRising.or(otherIsHighest));`

So no need for "refIsNotHighest " and "otherIsNotLowest"? 

**These cases should be handled for negative divergence:**
- "other" makes higher highs while "ref" makes lower highs
- ... other cases needed ?

Please explain by improving the code.

IsNegativeDivergentRule.java
nimo23: I dont understand what you mean. "index" is already provided within the method "isSatisfied()" as is. What is missing? 

Please explain by improving the code.

team172011: I think it could be stuctured as followed:

´´´
public class IsNegativeDivergentRule extends AbstractRule {

	private final Rule isNegativeDivergent;

	/**
	 * Constructor. <br/>
	 * 
	 * If "useStrictDivergence" is enabled, divergence is satisfied if values of
	 * "ref" strict decrease while the values of the "other" indicator strict
	 * increase.
	 * 
	 * @param ref the first indicator
	 * @param other the other indicator
	 * @param timeFrame the time frame
	 * @param useStrictDivergence true if only strict dicrease 
	 */
	public IsNegativeDivergentRule(Indicator<Decimal> ref, Indicator<Decimal> other, int timeFrame,
			boolean useStrictDivergence) {
		Indicator<Decimal> ref = ref; // can be local
		Indicator<Decimal> other = other; // can be local
		int timeFrame = timeFrame; // can be local
		this.useStrictDivergence = useStrictDivergence;
		
		// init rule isNegativeDivergent here...
		Rule refIsFalling = new IsFallingRule(ref, timeFrame);
		Rule otherIsRising = new IsRisingRule(other, timeFrame);
		
		if (useStrictDivergence) {
			isNegativeDivergent = refIsFalling.and(otherIsRising);
		} else {
			Rule refIsLowest = new IsLowestRule(ref, timeFrame);
			Rule otherIsHighest = new IsHighestRule(other, timeFrame);
			
			Rule refIsNotHighest = new IsHighestRule(ref, timeFrame).negation();
			Rule otherIsNotLowest = new IsLowestRule(other, timeFrame).negation();

			isNegativeDivergent = (refIsFalling.or(refIsLowest).or(refIsNotHighest))
					.and(otherIsRising.or(otherIsHighest).or(otherIsNotLowest));
		}
	}
	
	/**
	 * Constructor.
	 * 
	 * Rule for negative divergence between ref and MaxPriceIndicator.
	 * 
	 * @param series
	 * @param ref
	 * @param timeFrame
	 */
	public IsNegativeDivergentRule(TimeSeries series, Indicator<Decimal> ref, int timeFrame) {
		this(ref, new MaxPriceIndicator(series), timeFrame, false);
	}

	@Override
		
		// ...just check isSatisfied here
		final boolean satisfied = isNegativeDivergent.isSatisfied(index, tradingRecord);
		return satisfied;
	}
}
´´´

nimo23: What is the benefit of moving the logic within the constructor?

Please look at "IsFallingRule.java", "IsRisingRule.java", I had also put the logic within the isSatisfied()-method. Should I change these files and move the logic within the constructor, too?

IsNegativeDivergentRule.java
nimo23: I moved the logic within the constructor.

nimo23:
I moved the logic within the constructor and used local variables where possible.

nimo23:
Please delete this PR and I have replaced it by https://github.com/ta4j/ta4j/pull/72

PR #72, opened 2017-10-31
New: convergence and divergence rules

nimo23:
- ConvergenceDivergenceIndicator.java
- ChaikinOscillator.java
- add parameter "name" in (Base)Strategy.java


nimo23:
Please review code and check for **logical failures especially for the "otherIsNotHighest"** in "IsDivergent/IsConvergentPositiveRule" and **"refIsNotHighest"** in "IsDivergent/IsConvergentNegativeRule". 

I guess, with "otherIsNotHighest" and "refIsNotHighest", the "delay" between the two compared indicators are recognized: For example, even one makes higher highs (e.g. oscillator) and the other (e.g. price) does not follow this with the same speed, IsDivergentRule will recognize it as divergence (which is fine). Maybe you find improvements.

nimo23:
Maybe I can use the "CorrelationCoefficientIndicator.java" from ta4j/statistics within the convergent/divergentRules:

- if Correlation Coefficient>0, the two indicators move together (->convergent). 
- if Correlation Coefficient<0, the two indicators move in opposite directions (->convergent). 

 What do you think?

team172011:
I am unsure about the current `useStrictConvergence=true` case. Using statistics like `Correlation Coefficient` for that case sounds good

nimo23:
If useStrictConvergence=true, then CorrelationCoefficientIndicator is not used, otherwise it is used.

Look above at the example.

nimo23:
Which one should I use:

append with AND:
` .and(ccIsGt0);`

append with OR:
` .or(ccIsGt0);`

append with XOR:
` .xor(ccIsGt0);`

team172011:
I am unsure about the logical link in gerneral, because i can imagine several cases wherer the rule would be wrong satisfied:
For example `IsConvergentNegativeRule `
- `useStrictConvergence = true` -> "if both are falling" `satisfied = true` -> i am ok with that
- `useStrictConvergence = false`

| ref | other | case | possible situation |
| ----------|---------- |-------|-----------|
| ----------|---------- |-------|-----------|

nimo23:
Yes, you are right. I will use solely "CorrelationCoefficientIndicator". 

However, how should I determine the direction (**positive/negative**)? Should I use another Trend-Indicator/Oscillator to determine the direction of "ref" and the direction of "other". For example:

- 1.step: CorrelationCoefficientIndicator determines that ref and other is divergent
- 2.step: MovingAvgCrossover determines that "ref" is going up
- 3.step: MovingAvgCrossover determines that "other" is going down
--> this means: **positive** divergence


 Or does any other statistical function exist to determine the direction?

ta4j-core/src/main/java/org/ta4j/core/Strategy.java
```diff
@@ -83,8 +88,7 @@ default boolean shouldEnter(int index, TradingRecord tradingRecord) {
         if (isUnstableAt(index)) {
             return false;
         }
-        final boolean enter = getEntryRule().isSatisfied(index, tradingRecord);
```
nimo23: No need for "final boolean enter".

ta4j-core/src/main/java/org/ta4j/core/Strategy.java
```diff
@@ -104,7 +108,6 @@ default boolean shouldExit(int index, TradingRecord tradingRecord) {
         if (isUnstableAt(index)) {
             return false;
         }
-        final boolean exit = getExitRule().isSatisfied(index, tradingRecord);
```
nimo23: No need for "final boolean exit".

nimo23:
I accidently changed the Strategy/BaseStrategy within this PR instead of making a new PR. I only added a optional parameter "name" to be able to give a name for the strategy.

nimo23:
"IsConvergentNegative" will have the following logic:

```
@Override
		
		CorrelationCoefficientIndicator cc = new CorrelationCoefficientIndicator(ref, other, timeFrame);
		boolean isConvergent = cc.getValue(index).isGreaterThanOrEqual(Decimal.valueOf(minStrenght).abs());
		
		HMAIndicator hma = new HMAIndicator(ref, timeFrame);
		boolean isNegative = ref.getValue(index).isLessThan(hma.getValue(index));
		
		final boolean satisfied = isConvergent && isNegative;
		return satisfied;
	}
```

Maybe we can use something better than HMAIndicator to find the direction?

nimo23:
**The only problem with the logic is:**

Can we use something better than HMAIndicator to find the direction? Or is SimpleMovingAverage enough? Other TrendDirection-Indicators? Maybe with the help of ta4j-"SimpleLinearRegression.java" ?

team172011:
To detect a strict convergence/divergence is easy, but to detect some kind of "not so strict" convergence/divergence could be very vague. One user would use an EMA indicator another user would like to use HAMA and the third one would use Arron.... 
In my opining it is to specific to to identify the trend with the help of a specific indicator in a rule. A rule should be **as general as possible**. So one of the commonest detection could be like you say through a **simple linear regression** over the given time frame. To specifiy how strict this "not so strict" case can be we could use a parameter to compare the slope of the regression for each indicator. 

For example a value between 0 (not strict) - 100 (very strict)

- `parameter == 100` -> slope mus be exact the same

- `parameter == 50` -> slope of ref can be half of slop of other or more...

nimo23:
When I have this:

`SimpleLinearRegressionIndicator reg = new SimpleLinearRegressionIndicator(close, 7);`

and it outputs this for index n:

```
closeprice(n): "27.82"
simpleLinearReg(n): "27.7761"
```

**does this mean the following ?:**

For index n the n+1 value is probably around "27.7761" (= prediction value).

**Does "SimpleLinearRegressionIndicator" outputs the "predictionValue" for the next comming index?**

nimo23:
I should only look at the slope of each regression value?

**For example:**

```
(1) closeprice=27.68, regression=null, slope=null
(2) closeprice=27.66, regression=27.6643, slope= 27.6643
(3) closeprice=27.66, regression=27.6589, slope=-0.0054
(4) closeprice=27.67, regression=27.6567, slope=-0.0022
(5) closeprice=27.64, regression=27.656, slope=-0.0007 = (regression(5)) - (regression(4))
```
**With a timeFrame=3 on index (5), I can say:**
(regression(5)) - (regression(2)) = -0.0083 **(downtrend?)**

Should I assume that "-0.0083" is saying: negative->downtrend ?

team172011:
Spontaneous i would suggest to compare the slope of the regression line over the time frame:

grafik

nimo23:
When I have this:
I updated "SimpleLinearRegressionIndicator" to be able to set the returned type of the value (slope, intercept, y) within an optional constructor.

Now I can do this:

```
// timeFrame=7
SimpleLinearRegressionIndicator reg = new SimpleLinearRegressionIndicator(close, 7, SimpleLinearRegressionType.slope);
```

and it returns for index(n) the slope=-0.0058.

Is the value (-0.0058) **the slope between "timeFrame-7" till "timeFrame"** or is this **the slope between index 7 and its previous index 6** ?

team172011:
I have to take a look at your implementation of `SimpleLinearRegressionIndicator`, but you could check with

- slope between (8) and (1) = (27.6567-27.735)/(8-1) = -0,0112

- slope between (8) and (7) =  (27.6567-27.6589)/(8-7) = -022

nimo23:
I did not change the calculation of "SimpleLinearRegressionIndicator"..

nimo23:
I did not change the calculation of SimpleLinearRegressionIndicator. I only output the "slope" instead of "y" by setting the type within constructor. 

Why did you divide to "/(8-2)"?

The slope is nothing more than the difference of x and y and can also be calculated by DifferenceIndicator. Am I right?

So why using the slope of SimpleLinearRegression? Is there someting different?

nimo23:
By the way, actually the InSlopeRule only cares about the difference of two points and treats this as a slope. Should I change InSlopeRule to calculate the slope with the help of SimpleLinearRegression or the formula above you provided?

nimo23:
I need the "relative slope" instead of the "absolute slope" (which is calculated above) to be able to compare to "minSlope"-parameter.

**For example, a value between 0 (not strict) - 100 (very strict):**
- parameter >= +minSlope -> positive slope for positiveDivergence/positiveConvergence.
- parameter <= -minSlope -> negative slope for negativeDivergence/negativeConvergence.

nimo23:
I have (almost?) finished the "ConvergenceDivergenceIndicator.java". 

**Please review the source.** 

After that I am actually writing tests for "ConvergenceDivergenceIndicatorTest.java" for all cases.

team172011:
> By the way, actually the InSlopeRule.java only cares about the difference of two points and treats this as a slope. Should I change InSlopeRule.java to calculate the slope with the help of SimpleLinearRegression or the above formula you provided?

Good question, but for the moment i think the rule is regarding its javadoc description ok.

ta4j-core/src/main/java/org/ta4j/core/BaseStrategy.java
```diff
@@ -53,25 +56,43 @@ this software and associated documentation files (the "Software"), to deal in
...
     
     /**
      * Constructor.
+     * @param name the name of the strategy
      * @param entryRule the entry rule
      * @param exitRule the exit rule
+     * @param unstablePeriod
      */
-    public BaseStrategy(Rule entryRule, Rule exitRule, int unstablePeriod) {
+    public BaseStrategy(String name, Rule entryRule, Rule exitRule, int unstablePeriod) {
```
team172011, line 89: Please add old constructor `public BaseStrategy(Rule entryRule, Rule exitRule, int unstablePeriod)` that calls `this(null, entryRule,exitRule, unstablePerios)`

ta4j-core/src/main/java/org/ta4j/core/indicators/AwesomeOscillatorIndicator.java
```diff
@@ -24,27 +24,50 @@ this software and associated documentation files (the "Software"), to deal in
...
+import org.ta4j.core.TimeSeries;
+import org.ta4j.core.indicators.helpers.MedianPriceIndicator;
 
 /**
  * Awesome oscillator. (AO)
- * <p></p>
- * @see <a href="http://www.forexgurus.co.uk/indicators/awesome-oscillator">http://www.forexgurus.co.uk/indicators/awesome-oscillator</a>
+ * <p>
+ * see https://www.tradingview.com/wiki/Awesome_Oscillator_(AO)
  */
```
team172011: Okay with replacing `<p></p>` but we should take `<p/>` than

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicator.java
```diff
@@ -0,0 +1,352 @@
...
+		 * other-{@link Indicator indicator} decrease within the timeFrame. In
+		 * short: "other" and "ref" makes lower lows.
+		 */
+		negativeConvergent, 
+		
+		/**
+		 * Returns true for <b>"positiveDivergent"</b> when the values of the
+		 * ref-{@link Indicator indicator} increase and the values of the
+		 * other-{@link Indicator indicator} decrease within a timeFrame. In
+		 * short: "other" makes lower lows while "ref" makes higher lows.
```
team172011: `short: "other" makes lower lows while "ref" makes higher lows.`

Would it be better to say "...while "ref makes higher highs" ?

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicator.java
```diff
@@ -0,0 +1,352 @@
...
+	 * The <b>"minSlope"</b> is the minimum required slope for convergence or divergence
+	 * and must be a number between "0.1" and "1.0": <br/>
+	 * <br/>
+	 * 0.1: very unstrict<br/>
+	 * 0.3: strict (recommended) <br/>
+	 * 1.0: very strict <br/>
+	 * 
+	 * @param ref the indicator
+	 * @param other the other indicator
+	 * @param timeFrame
```
team172011: Please add at least simple description for parmeter e.g `@param timeFrame the time frame`

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicator.java
```diff
@@ -0,0 +1,352 @@
...
+		 */
+		negativeDivergent,
+		
+		/**
+		 * Returns true for <b>"positiveConvergentStrict"</b> when the values of
+		 * the ref-{@link Indicator indicator} and the values of the
+		 * other-{@link Indicator indicator} increase consecutively within a
+		 * timeFrame. In short: "other" and "ref" makes strict higher highs.
+		 */
+		positiveConvergentStrict,
```
team172011, line 89: The user must be informed that if he uses one of the "**strict**" enum members, the minStrenght and minSlope will have no effect

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicator.java
team172011: Also for second constructor and the `@param index`

nimo23:
Okay, I am finished! 

Thanks for your help!

team172011:
thanks, too. I have added you to the AUTHORS file

PR #76, opened 2017-11-05
First implementation of RandomWalkIndex[High|Low]Indicator

### LINKED ISSUE #57: formula of "randomWalkIndex" correct?
I compared the computed values of ta4j and it differs from that of tradingView. Please check out.

According to tradingView or other sources (http://www.fmlabs.com/reference/default.htm?url=RWI.htm, https://www.technicalindicators.net/indicators-technical-analysis/168-rwi-random-walk-index), the randomWalkIndexHigh/Low is calculated by **returning the maximum computed value of _all_ the values from the formula within a timeFrame**:

However, I dont see any recursive call or loop within the actual ta4j-formula, it only computes the value of the actual index without comparing this to all the other computed rwi-values within the timeFrame. 

According to tradingView, the formula must have something like a loop to compare all other calculated rwi-values and return the highest of that value for the actual index: 

```
// loop through the ticks within the timeFrame
for i = minp to maxp 
    // calculate each rwi and return the highest one
    rwihi:=i==minp?(high-low[i])/(atr(i)*sqrt(i)):max((high-low[i])/(atr(i)*sqrt(i)),rwihi)
    rwilo:=i==minp?(high[i]-low)/(atr(i)*sqrt(i)):max((high[i]-low)/(atr(i)*sqrt(i)),rwilo)
rwi=rwihi-rwilo 
```

team172011:
Fixes #57.

- Fix RandomWalkHighIndicator
- FixRandomWalkLowIndicator
- Test files with output of first calculations

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/RWIHighIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/RWIHighIndicator.java
@@ -33,7 +33,7 @@ import org.ta4j.core.indicators.helpers.MinPriceIndicator;
  *
  * See https://www.technicalindicators.net/indicators-technical-analysis/168-rwi-random-walk-index
  */
-public class RandomWalkIndexHighIndicator extends CachedIndicator<Decimal> {
+public class RWIHighIndicator extends CachedIndicator<Decimal> {
 
     private final MaxPriceIndicator maxPrice;
     
@@ -47,7 +47,7 @@ public class RandomWalkIndexHighIndicator extends CachedIndicator<Decimal> {
      * @param series the series
      * @param timeFrame the time frame
      */
-    public RandomWalkIndexHighIndicator(TimeSeries series, int timeFrame) {
+    public RWIHighIndicator(TimeSeries series, int timeFrame) {
         super(series);
         this.timeFrame = timeFrame;
         maxPrice = new MaxPriceIndicator(series);
@@ -57,23 +57,26 @@ public class RandomWalkIndexHighIndicator extends CachedIndicator<Decimal> {
 
     @Override
     protected Decimal calculate(int index) {
-        int lastIndex = Math.max(0, index - timeFrame+1);
-        int n = 2;
-        AverageTrueRangeIndicator averageTrueRange = new AverageTrueRangeIndicator(getTimeSeries(), n);
-        Decimal highestRWI = maxPrice.getValue(index).minus(minPrice.getValue(Math.max(0, index-1)))
-                .dividedBy(averageTrueRange.getValue(index).multipliedBy(Decimal.valueOf(Math.sqrt(n))));
-
-        for(int i = index-2; i >= lastIndex; i--) {
-            n++;
-            averageTrueRange = new AverageTrueRangeIndicator(getTimeSeries(), n);
-            Decimal currentRWI = maxPrice.getValue(index).minus(minPrice.getValue(i))
-                    .dividedBy(averageTrueRange.getValue(index).multipliedBy(Decimal.valueOf(Math.sqrt(n))));
-            if(currentRWI.isGreaterThan(highestRWI)){
-                highestRWI = currentRWI;
-            }
-
+        Decimal highestRWI = Decimal.NaN;
+        
+        for(int n = 2; n <= this.timeFrame; n++) {
+           Decimal currentRWI = calcRWIHighValue(index, index-n, n);
+           if(currentRWI.isGreaterThan(highestRWI)){
+               highestRWI = currentRWI;
+           }   
         }
         return highestRWI;
+    }
+    
+    /**
+     * @param t = current index
+     * @param t_n = current index - n
+     * @param n = n starting at 2 increments until n = time frame
+     */
+    private Decimal calcRWIHighValue(int t, int t_n, int n){
+        AverageTrueRangeIndicator averageTrueRange = new AverageTrueRangeIndicator(getTimeSeries(), n);
+        return maxPrice.getValue(t).minus(minPrice.getValue(t_n))
+                    .dividedBy(averageTrueRange.getValue(t).multipliedBy(Decimal.valueOf(Math.sqrt(n))));
     }
     
     @Override
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/RandomWalkIndexHighIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/RandomWalkIndexHighIndicator.java
@@ -29,17 +29,15 @@ import org.ta4j.core.indicators.helpers.MaxPriceIndicator;
 import org.ta4j.core.indicators.helpers.MinPriceIndicator;
 
 /**
- * The Class RandomWalkIndexHighIndicator.
+ * The Random Walk Index High Indicator. <p />
+ *
+ * See https://www.technicalindicators.net/indicators-technical-analysis/168-rwi-random-walk-index
  */
 public class RandomWalkIndexHighIndicator extends CachedIndicator<Decimal> {
 
     private final MaxPriceIndicator maxPrice;
     
     private final MinPriceIndicator minPrice;
-    
-    private final AverageTrueRangeIndicator averageTrueRange;
-    
-    private final Decimal sqrtTimeFrame;
     
     private final int timeFrame;
     
@@ -54,14 +52,28 @@ public class RandomWalkIndexHighIndicator extends CachedIndicator<Decimal> {
         this.timeFrame = timeFrame;
         maxPrice = new MaxPriceIndicator(series);
         minPrice = new MinPriceIndicator(series);
-        averageTrueRange = new AverageTrueRangeIndicator(series, timeFrame);
-        sqrtTimeFrame = Decimal.valueOf(timeFrame).sqrt();
+
     }
 
     @Override
     protected Decimal calculate(int index) {
-        return maxPrice.getValue(index).minus(minPrice.getValue(Math.max(0, index - timeFrame)))
-                .dividedBy(averageTrueRange.getValue(index).multipliedBy(sqrtTimeFrame));
+        int lastIndex = Math.max(0, index - timeFrame+1);
+        int n = 2;
+        AverageTrueRangeIndicator averageTrueRange = new AverageTrueRangeIndicator(getTimeSeries(), n);
+        Decimal highestRWI = maxPrice.getValue(index).minus(minPrice.getValue(Math.max(0, index-1)))
+                .dividedBy(averageTrueRange.getValue(index).multipliedBy(Decimal.valueOf(Math.sqrt(n))));
+
+        for(int i = index-2; i >= lastIndex; i--) {
+            n++;
+            averageTrueRange = new AverageTrueRangeIndicator(getTimeSeries(), n);
+            Decimal currentRWI = maxPrice.getValue(index).minus(minPrice.getValue(i))
+                    .dividedBy(averageTrueRange.getValue(index).multipliedBy(Decimal.valueOf(Math.sqrt(n))));
+            if(currentRWI.isGreaterThan(highestRWI)){
+                highestRWI = currentRWI;
+            }
+
+        }
+        return highestRWI;
     }
     
     @Override
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/RandomWalkIndexLowIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/RandomWalkIndexLowIndicator.java
@@ -60,8 +60,23 @@ private final MaxPriceIndicator maxPrice;
 
     @Override
     protected Decimal calculate(int index) {
-        return maxPrice.getValue(Math.max(0, index - timeFrame)).minus(minPrice.getValue(index))
-                .dividedBy(averageTrueRange.getValue(index).multipliedBy(sqrtTimeFrame));
+        int lastIndex = Math.max(0, index - timeFrame+1);
+        int n = 2;
+        AverageTrueRangeIndicator averageTrueRange = new AverageTrueRangeIndicator(getTimeSeries(), n);
+        Decimal lowestRWI = maxPrice.getValue(Math.max(0, index-1)).minus(minPrice.getValue(index))
+                .dividedBy(averageTrueRange.getValue(index).multipliedBy(Decimal.valueOf(Math.sqrt(n))));
+
+        for(int i = index-2; i >= lastIndex; i--) {
+            n++;
+            averageTrueRange = new AverageTrueRangeIndicator(getTimeSeries(), n);
+            Decimal currentRWI = maxPrice.getValue(index).minus(minPrice.getValue(i))
+                    .dividedBy(averageTrueRange.getValue(index).multipliedBy(Decimal.valueOf(Math.sqrt(n))));
+            if(currentRWI.isLessThan(lowestRWI)){
+                lowestRWI = currentRWI;
+            }
+
+        }
+        return lowestRWI;
     }
     
     @Override
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,4 +12,5 @@ Contributors
     gcauchis (Gabriel Cauchis)
     bukefalos
     TheCookieLab
-    team172011
+    team172011 (Simon-Justus Wimmer)
+    nimo23
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,15 +7,21 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 
 ### Changed
 - **JustOnceRule**: Now it is possible to add another rule so that this rule is satisfied if the inner rule is satisfied for the first time
+- **MeanDeviationIndicator**: moved to statistics package
+- **Decimal**: Use `BigDecimal::valueof` instead of instantiating a new BigDecimal for double, int and long
+- **Strategy:** can now have a optional parameter "name".
 
 ## Added
+- **ConvergenceDivergenceIndicator**: New Indicator for positive/negative convergence and divergence.
+- **BooleanTransformIndicator**: New indicator to transform any decimal indicator to a boolean indicator by using logical operations.
+- **DecimalTransformIndicator**: New indicator to transforms any indicator by using common math operations.
 
 
 ## 0.10 (2017-10-23)
 
 ### VERY Important note!!!!
 
-with the release 0.10 we have changed the previous java package definition to org.ta4j or to be more specific to org.ta4j.core (the new organisation). YOu have to reorganize all your refernces to the new packages!
+with the release 0.10 we have changed the previous java package definition to org.ta4j or to be more specific to org.ta4j.core (the new organisation). You have to reorganize all your refernces to the new packages!
 In eclipse you can do this easily by selecting your sources and run "Organize imports"
 _Changed ownership of the ta4j repository_: from mdeverdelhan/ta4j (stopped the maintenance) to ta4j/ta4j (new organization)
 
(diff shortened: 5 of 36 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/RandomWalkIndexLowIndicator.java
```diff
@@ -60,8 +60,23 @@ public RandomWalkIndexLowIndicator(TimeSeries series, int timeFrame) {
```
nimo23, line 59: Please delete these global props from the class:

```
averageTrueRange = new AverageTrueRangeIndicator(series, timeFrame);
sqrtTimeFrame = Decimal.valueOf(timeFrame).sqrt();
```

ta4j-core/src/main/java/org/ta4j/core/indicators/RandomWalkIndexHighIndicator.java
```diff
@@ -54,14 +52,28 @@ public RandomWalkIndexHighIndicator(TimeSeries series, int timeFrame) {
...
+
     }
 
     @Override
     protected Decimal calculate(int index) {
-        return maxPrice.getValue(index).minus(minPrice.getValue(Math.max(0, index - timeFrame)))
-                .dividedBy(averageTrueRange.getValue(index).multipliedBy(sqrtTimeFrame));
+        int lastIndex = Math.max(0, index - timeFrame+1);
+        int n = 2;
+        AverageTrueRangeIndicator averageTrueRange = new AverageTrueRangeIndicator(getTimeSeries(), n);
```
nimo23: Maybe you can extract this to a method instead of duplicating here and in the loop:

```
private Decimal calculateRwiHigh(int index, int pos){
		AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), pos);
		return maxPrice.getValue(index).minus(minPrice.getValue(Math.max(0, index - 1)))
				.dividedBy(atr.getValue(index).multipliedBy(Decimal.valueOf(Math.sqrt(pos))));
	}
```

The same goes for RWILowInciator.java

ta4j-core/src/main/java/org/ta4j/core/indicators/RandomWalkIndexHighIndicator.java
```diff
@@ -54,14 +52,28 @@ public RandomWalkIndexHighIndicator(TimeSeries series, int timeFrame) {
         this.timeFrame = timeFrame;
         maxPrice = new MaxPriceIndicator(series);
         minPrice = new MinPriceIndicator(series);
-        averageTrueRange = new AverageTrueRangeIndicator(series, timeFrame);
-        sqrtTimeFrame = Decimal.valueOf(timeFrame).sqrt();
+
     }
 
     @Override
```
nimo23, line 61: **Please replace your rwiHigh-calculate()-method to something like this:**
(The same goes for rwiLow-calculate())
```

@Override
		
	Decimal highestRWI = Decimal.NaN;
	int startIndex = Math.max(0, index - timeFrame + 1);
		
        // TODO: should we go back or forth? 
	for (int i = startIndex; i <= index - 1; i++) {
		Decimal currentRWI = calculateRwiHigh(index, i);
		if (currentRWI.isGreaterThan(highestRWI) || highestRWI.isNaN()) {
			highestRWI = currentRWI;
		}
	}
	return highestRWI;
}
	
private Decimal calculateRwiHigh(int index, int pos){
	AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), pos);
        // TODO: which to use?: "atr.getValue(index)" or "atr.getValue(pos)"
	return maxPrice.getValue(index).minus(minPrice.getValue(pos))
			.dividedBy(atr.getValue(index).multipliedBy(Decimal.valueOf(Math.sqrt(pos))));
}
```

However, both yours and my version does not return the same results each other and both does not even return the results compared to the results in tradingview. To find the reason, we should at first reduce the code to something like my version.

ta4j-core/src/main/java/org/ta4j/core/indicators/RandomWalkIndexHighIndicator.java
```diff
@@ -29,18 +29,16 @@ this software and associated documentation files (the "Software"), to deal in
 import org.ta4j.core.indicators.helpers.MinPriceIndicator;
 
 /**
- * The Class RandomWalkIndexHighIndicator.
+ * The Random Walk Index High Indicator. <p />
+ *
+ * See https://www.technicalindicators.net/indicators-technical-analysis/168-rwi-random-walk-index
  */
 public class RandomWalkIndexHighIndicator extends CachedIndicator<Decimal> {
```
nimo23: We should use the shorter filename **"RWIHighIndicator.java"** and **"RWILowIndicator.java"** instead of "RandomWalkIndexHighIndicator.java" and "RandomWalkIndexLowIndicator.java". Better to read and _"RWI"_ is the common abbreviation for **"RandomWalkIndex"**.

ta4j-core/src/main/java/org/ta4j/core/indicators/RandomWalkIndexHighIndicator.java
team172011: `maxPrice.getValue(index).minus(minPrice.getValue(pos)` -> `pos` must **starting with index - one** (today-yesterday) and gets incremented in each iteration by one.

The time frame of the ATR (**starting with two**) must be incremented by one in each iteration. In your version it starts at startIndex and ends with currentIndex ??

nimo23: for example, timeframe=4:

it goes back 4 times and starts there (=startIndex), calculates each rwi till actual index-1 (because the actual index needs at least one previous bar/tick -> today-yesterday) and then returns **the highest of all the 3 calculated rwi for the actual index.**

However, take your loop with start=2 and decrement (go back till index-timeframe) instead of increment and eliminate the redundancy to the form above.

If right, change the for-loop to the loop from your version but eliminate the redundant code of your version like I did.

team172011: `new AverageTrueRangeIndicator(getTimeSeries(), pos);` pos has to be the timeFame not index

nimo23: https://rtmath.net/helpFinAnalysis/html/934563a8-9171-42d2-8444-486691234b1d.htm

>where Atr is the average true range over N periods preceding the current period. 

ok, then:

```
private Decimal calculateRwiHigh(int index, int pos){
	AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), timeFrame);
        // TODO: which to use?: "atr.getValue(index)" or "atr.getValue(timeFrame)" or "atr.getValue(pos)"
	return maxPrice.getValue(index).minus(minPrice.getValue(pos))
			.dividedBy(atr.getValue(index).multipliedBy(Decimal.valueOf(Math.sqrt(pos))));
}
```

which to use?: ".dividedBy(atr.getValue(index))" or ".dividedBy(atr.getValue(pos))"

according to http://www.fmlabs.com/reference/default.htm?url=RWI.htm, it must be: ".dividedBy(atr.getValue(index))"

However, according to https://books.google.de/books?id=kFKLCwAAQBAJ&pg=PA11&lpg=PA11&dq=random+walk+index+formula&source=bl&ots=oroGdgtRsn&sig=2BTuK0HLagx1nbXMyEa_ANcebEo&hl=de&sa=X&ved=0ahUKEwiozKTVlK3XAhUhJJoKHf7SCFc4FBDoAQhSMAU#v=onepage&q=random%20walk%20index%20formula&f=false:

it must be 

".dividedBy(atr.getValue(pos))"

Or?

team172011: I think `.dividedBy(atr.getValue(index))` because in `AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), timeFrame);` we have to take the "**current**" timeFrame, _**form `n=2 ` to `this.timeFrame`.**_ So we need in each iteration the index (always same) and the currentTimeFrame (n that increments in each iteration)

nimo23: http://www.neuroshell.com/manuals/ais2/index.html?rndwalkhigh.htm

nimo23:
Look at this solution and compare it with yours:

```
@Override

	Decimal highestRWI = Decimal.NaN;
	int lastIndex = Math.max(0, index - timeFrame + 1);
		
	for (int i = index - 1; i >= lastIndex; i--) {
		Decimal currentRWI = calculateRwiHigh(index, i);
		if (currentRWI.isGreaterThan(highestRWI) || highestRWI.isNaN()) {
			highestRWI = currentRWI;
		}
	}
		
	return highestRWI;
}
	
private Decimal calculateRwiHigh(int index, int pos) {
	// high - low(index-timeFrame+1) / atr(n) * sqrt(n)
	// high - low(-(n-1)) / atr(n) * sqrt(n)
	AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), pos);
	return maxPrice.getValue(index).minus(minPrice.getValue(pos))
			.dividedBy(atr.getValue(pos).multipliedBy(Decimal.valueOf(Math.sqrt(pos))));
}
```

- According to https://www.technicalindicators.net/indicators-technical-analysis/168-rwi-random-walk-index and the other sources I mentioned above, `atr.getValue(pos)` must be used and not `atr.getValue(index)` (atr must be calculated for the actual position which goes backwards till lastIndex)
- you start your calculation with `n=2` which is the right way, however, within my loop I also start the calculation with the current(high) and the previous(low), which is the same as n=2.

team172011:
For sure we could encapsulates the RWI calculation, but i still do not understand your function call in the loop:

- for example ` index = 1000` and `this.timeFrame = 5`
 -> first iteration will call `calculateRwiHigh(1000, 999)`
 -> `AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), 999);` -> timeFrame is 999 here ???
 -> `pos` should be `n` = 2 and increment each iteration and not the `index-i `value

Am i right with this:

```
@Override

	Decimal highestRWI = Decimal.NaN;
	int lastIndex = Math.max(0, index - timeFrame + 1);
	int diff = index - lastIndex; // max n
	for (int i = 2; i <= diff; i++) {
		Decimal currentRWI = calculateRwiHigh(index, i);
		if (currentRWI.isGreaterThan(highestRWI) || highestRWI.isNaN()) {
			highestRWI = currentRWI;
		}
	}
		
	return highestRWI;
}
	
private Decimal calculateRwiHigh(int index, int pos) {
	// high - low(index-timeFrame+1) / atr(n) * sqrt(n)
	// high - low(-(n-1)) / atr(n) * sqrt(n)
	AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), pos);
	return maxPrice.getValue(index).minus(minPrice.getValue(pos))
			.dividedBy(atr.getValue(pos).multipliedBy(Decimal.valueOf(Math.sqrt(pos))));
}
```

team172011:
I suggest we should implement the formula of http://www.fmlabs.com/reference/default.htm?url=RWI.htm that is written of this paper (last page) www.books.mec.biz/downloads/Detecting_Trend_Direction_And_Strength/NjQxNDI4NzM2




In your approach you are using **indices for the** **time frame** of the ATR. I think this does not make any sense:
```
for n=2 (first iteration):
AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), 1000);
and then:
AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), 999);
AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), 998);
AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), 997);
.. till lastIndex
```
-> independent from `this.timeFrame` the ATR gets a timeFrame of `1000` just because the `index` is `1000`. If the `index` would be `10000000` you would start with a `timeFrame` of `10000000` for the ATR in the first iteration?!

We have to use `n` for the time frame of the ATR. Starting at 2 
```
for n=2 (first iteration, i = index - this.timeFrame+1 ):
AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), n);
(...) atr.get(i);
and then:

for n=3 (second iteration, i = index - this.timeFrame+2 ):
AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), n+1);
(...) atr.get(i);
for n=4 (second iteration, i = index - this.timeFrame+3 ):
AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), n+2);
(...) atr.get(i);
for n=5 (second iteration, i = index - this.timeFrame+4 ):
AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), n+3);
(...) atr.get(i);
.. till n+timeFrame
```

nimo23:
> In your approach you are using indices for the time frame of the ATR. I think this does not make any sense

yes, makes no sense.

According to "atr: is the average daily range over the n days **preceding** today." (means to me n-1, where 1=today). So yes, we have to start with 2 and the atr is the atr of the actual index.

http://www.neuroshell.com/manuals/ais2/index.html?rndwalkhigh.htm:

```
RWH[i,p] = (High[i]–Low[i-p+1]) / ((TR[i-1]+TR[i-2]+..+TR[i-p])/p)*Sqrt(p)),
TR[i] – true range,
p=2..Period,
i=current bar.
```

Could you make a renew version of "RwiHighIndicator.java" and "RwiLowIndicator.java". I will test it afterwards with real data.

nimo23:
You can german: http://www.tradesignalonline.com/lexicon/view.aspx?id=Random+Walk+Index+(RWI)

Please read the overall article, there is also a **accurate implementation below**.

team172011:
I don't know what he means with a 'open calculation period' that is determined by 5 and 65, also the equlila sripts are hard to read for me because i don't know the build in functions

ta4j-core/src/main/java/org/ta4j/core/indicators/RWIHighIndicator.java
```diff
@@ -57,25 +57,28 @@ public RandomWalkIndexHighIndicator(TimeSeries series, int timeFrame) {
...
-                    .dividedBy(averageTrueRange.getValue(index).multipliedBy(Decimal.valueOf(Math.sqrt(n))));
-            if(currentRWI.isGreaterThan(highestRWI)){
-                highestRWI = currentRWI;
-            }
-
+        Decimal highestRWI = Decimal.NaN;
+        
+        for(int n = 2; n <= this.timeFrame; n++) {
+           Decimal currentRWI = calcRWIHighValue(index, index-n, n);
+           if(currentRWI.isGreaterThan(highestRWI)){
```
nimo23: Will be always "false" because of "highestRwi" is Decimal.NaN at beginning.
Hence, you need: 

```
// so highestRwi can be changed from "NaN" to the value of currentRwi
if (currentRWI.isGreaterThan(highestRWI) || highestRWI.isNaN()) {
	highestRWI = currentRWI;
}
```

The same for RwiLowIndicator.

ta4j-core/src/main/java/org/ta4j/core/indicators/RWIHighIndicator.java
```diff
@@ -57,25 +57,28 @@ public RandomWalkIndexHighIndicator(TimeSeries series, int timeFrame) {
...
         }
         return highestRWI;
     }
     
+    /**
+     * @param t = current index
+     * @param t_n = current index - n
+     * @param n = n starting at 2 increments until n = time frame
+     */
+    private Decimal calcRWIHighValue(int t, int t_n, int n){
```
nimo23: no need for method-parameter "t_n".

Say "t-n" directly within the method and rename param "t" to "index" and use shorter name "atr".

```
private Decimal calcRWIHighValue(int index, int n){
      AverageTrueRangeIndicator atr = new AverageTrueRangeIndicator(getTimeSeries(), n);
      return maxPrice.getValue(index).minus(minPrice.getValue(index - n))
                  .dividedBy(atr.getValue(index).multipliedBy(Decimal.valueOf(Math.sqrt(n))));
   }
```

nimo23:
Look at the scripts Equilla "_rrwi" (for rwiHigh) and "_lrwi" (for rwiLow) from http://www.tradesignalonline.com/lexicon/view.aspx?id=Random+Walk+Index+(RWI):
```
**I commmented the source.**

// calculation: rwiHigh
if Period > beg then
begin
// normally, from "2"..to "timeFrame" (-> this is your open calculation period)
For Counter = beg to Period
begin
// atr(counter+1) and sqrt(counter)
value2 = average( truerange, counter+1 ) * squareroot(counter);
if value2 <> 0 then
begin
// high(counter)
value3=( High[Counter]-Low ) / value2;
end
else
begin
value3=0;
end;
```

For rwiLow (from Equilla "_lrwi"):

```
// calculation: rwiLow
if Period > beg then
begin
// normally, from "2"..to "timeFrame" (-> this is your open calculation period)
For Counter = beg to Period
begin
// atr(counter+1) and sqrt(counter)
value2 = average( truerange,counter + 1 ) * squareroot(counter);
if value2 <> 0 then
begin
// low(counter)
value3 = (High - Low[counter]) / value2;
end
else
begin
value3 = 0;
end;
```

**The functions are called by:**

```
LowRWI = _lrwi(Period,Beg);
HighRWI = _rrwi(Period,Beg);
where Period = 50 and Beg= 10.
```

The **"open calculation period"** is defined by start ("Beg") and end ("Period"), so you need 2 parameters: **start** and **end** and not only timeFrame. You defined your start to a fixed number "2", however, you need to extract this to method parameter, so user can decide another start-value.
And look at the atr: the atr-value is always the actual tick/bar and not the fixed value of index (`average( truerange,counter + 1 )`).

team172011:
Okay, thanks. 
That means `beg` is a fixed index and the calculation period gets bigger with increasing `index`. And `Period` is the current index of the `calculate(index)` call?

Than once have to initialize the indicator with
`RWIHighIndicator rwih = new RWIHighIndicator(series, beg) // could also add constructor for other indicator`

And every `calculate(index)` call with `index < beg ` would be NaN?

nimo23:
As I understand, "beg" is the starting point (you start e.g. at "2") end "timeFrame" is the ending point (which also must be given by constructor).

The code above says it.

nimo23:
https://www.metastock.com/customer/resources/formulas/?id=48

team172011:
But than if

- index < beg -> calculation is not possible because starting point in future
- index < timeFrame -> calculation is not possible because endpoint in future
- beg < index > timeFrame -> ??


http://www.tradesignalonline.com/lexicon/view.aspx?id=Random+Walk+Index+(RWI)#

nimo23:
Look at how the calculation of this

```
// constructor with two properties: "period" and "beg"
LowRWI = _lrwi(Period,Beg);
HighRWI = _rrwi(Period,Beg);
where Period = 50 and Beg= 10.
```

is possible?

team172011:
I don't agree with the implementation from http://www.tradesignalonline.com/lexicon/view.aspx?id=Random+Walk+Index+(RWI)#

All other sources use just a timeFrame and calculate the min/max of (RWI(n-timeFrame), RWI( n))

- https://www.technicalindicators.net/indicators-technical-analysis/168-rwi-random-walk-index#

> How to understand the formulas? Well, first of all you should choose
> some days for RWI calculation. So, if you decide to calculate it for
> e.g. 10 days then you have to calculate it for each day separately. The
> first RWI of High is today’s High minus Low "t" days ago (in this case 1
> day ago). The next RWI of High would be today’s High minus Low 2 days
> ago etc. You would have to calculate all the differences between the
> first 1 and the previous 9 days ago. Each difference must be divided by
> (Average daily range multiplied by the Square root of the days). The
> actual RWI value is the highest one of all the 9 values within the last
> 10 days.

- https://www.investopedia.com/terms/r/random-walk-index.asp#ixzz4ymDRXHQe (The plot shows also just time frame as parameter)

> The Random Walk Index is typically used over two to seven periods for short-term trading and eight to 64 periods for long-term trading, but traders may want to experiment with these numbers to see what works best for them.

- https://www.linnsoft.com/techind/random-walk-index
> For each bar, the RWI High is computed by calculating the maximum of the following values:
> 
> (HI - LO.1) / (ATR.1(1) * SQRT(1)) (HI - LO.2) / (ATR.1(2) * SQRT(2)) (HI - LO.3) / (ATR.1(3) * SQRT(3)) (HI - LO.4) / (ATR.1(4) * SQRT(4)) (HI - LO.5) / (ATR.1(5) * SQRT(5)) ... (HI - LO.n) / (ATR.1(n) * SQRT(n))
> 
> Where n is the RWI Period.
> 
> Similarly,
> 
> the RWI Low is computed for each bar by calculating the maximum of the following values:
> 
> (HI.1 - LO) / (ATR.1(1) * SQRT(1)) (HI.2 - LO) / (ATR.1(2) * SQRT(2)) (HI.3 - LO) / (ATR.1(3) * SQRT(3)) (HI.4 - LO) / (ATR.1(4) * SQRT(4)) (HI.5 - LO) / (ATR.1(5) * SQRT(5)) ... (HI.n - LO) / (ATR.1(n) * SQRT(n))
> 
> ATR.1(n) is the n-period Average True Range, one bar back. HI.n is the high n bars back.

nimo23:
Ok, u are right. So one point is not clear for me:

https://www.technicalindicators.net/indicators-technical-analysis/168-rwi-random-walk-index#

> ...Each difference must be divided by Average daily range multiplied by the Square root **of the days**. ..

**Which days?** The **overall days** or **the days until the actual iterations**?

**For example:**

> The next RWI of High would be today’s High minus Low 2 days ago 

And which ATR does this RWI(2) corresponds to? Is this the "Average daily range multiplied by the Square root **of the **2** days or the **overall 10 days**? This is the question.

PR #84, opened 2017-11-11
Use BigDecimal::valueof instead of instantiating a new BigDecimal for int and long

katsadim:
This is a minor change done in Decimal class.

Depending on JRE implementation, this will give a slight performance boost when creating BigDecimals out of integer values 0 to 10. This is due to the caching mechanism that BigDecimal implements.

Using `MATH_CONTEXT` during long or int creation is not particularly useful. Defining the `MATH_CONTEXT` during math calculations suffices. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 ### Changed
 - **JustOnceRule**: Now it is possible to add another rule so that this rule is satisfied if the inner rule is satisfied for the first time
 - **MeanDeviationIndicator**: moved to statistics package
+- **Decimal**: Use `BigDecimal::valueof` instead of instantiating a new BigDecimal for int and long
 
 ## Added
 - **ConvergenceDivergenceIndicator**: New Indicator for positive/negative convergence and divergence.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 ### Changed
 - **JustOnceRule**: Now it is possible to add another rule so that this rule is satisfied if the inner rule is satisfied for the first time
 - **MeanDeviationIndicator**: moved to statistics package
-- **Decimal**: Use `BigDecimal::valueof` instead of instantiating a new BigDecimal for int and long
+- **Decimal**: Use `BigDecimal::valueof` instead of instantiating a new BigDecimal for double, int and long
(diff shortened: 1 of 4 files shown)


team172011:
Would also be appropriate for double:

valueOf(double val)
> Note: This is generally the preferred way to convert a double (or float) into a BigDecimal, as the value returned is equal to that resulting from constructing a BigDecimal from the result of using Double.toString(double).

But here accuracy has greater priority than performance. And we need MathContext (?)
> Notes:
The results of this constructor can be somewhat unpredictable. One might assume that writing new BigDecimal(0.1) in Java creates a BigDecimal which is exactly equal to 0.1 (an unscaled value of 1, with a scale of 1), but it is actually equal to 0.1000000000000000055511151231257827021181583404541015625. This is because 0.1 cannot be represented exactly as a double (or, for that matter, as a binary fraction of any finite length). Thus, the value that is being passed in to the constructor is not exactly equal to 0.1, appearances notwithstanding.
    The String constructor, on the other hand, is perfectly predictable: writing new BigDecimal("0.1") creates a BigDecimal which is exactly equal to 0.1, as one would expect. Therefore, it is generally recommended that the String constructor be used in preference to this one.
    When a double must be used as a source for a BigDecimal, note that this constructor provides an exact conversion; it does not give the same result as converting the double to a String using the Double.toString(double) method and then using the BigDecimal(String) constructor. To get that result, use the static valueOf(double) method.

katsadim:
Actually i think you are right in in using `BigDecimal.valueOf(double val)` as well. And we don't loose in accuracy. 
Consider the following example:

```java
BigDecimal bigDecimal = new BigDecimal(Double.MAX_VALUE, new MathContext(32, RoundingMode.HALF_UP)); // this is MATH_CONTEXT
BigDecimal bigDecimal2 = new BigDecimal(Double.MAX_VALUE, new MathContext(33, RoundingMode.HALF_UP));
BigDecimal bigDecimal3 = BigDecimal.valueOf(Double.MAX_VALUE);
System.out.println(bigDecimal);
System.out.println(bigDecimal2);
System.out.println(bigDecimal3);
System.out.println(Double.MAX_VALUE);
```

This is the output:
```bash
1.7976931348623157081452742373170E+308
1.79769313486231570814527423731704E+308
1.7976931348623157E+308
1.7976931348623157E308
```
By using `MATH_CONTEXT` with such a big precision setting is like trying to increase the precision of the already limited precision of double. But all it does is to add garbage digits to the right.

I will go ahead and make the change.

nimo23:
https://github.com/ta4j/ta4j/issues/86

Issue #89, opened 2017-11-15
fix RSI indicator computation


abzif:
Computation of RSI indicator is different than original Welles Wilder formula

described here:
https://en.wikipedia.org/wiki/Relative_strength_index
or here:
http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:relative_strength_index_rsi

I can fix this computation and contribute the code ta4j project

I was developing a few indicators some time ago and "invented" method of indicator testing using xls

Indicator formula is written in xls. Unit test reads and evaluate values from xls (using poi library) and compare it to indicator values.
In my opinion such independent testing raises quality of indicator implementation. I can also contribute this method of testing to project.

nimo23:
There are 2 version of RSI within taj4:

- SmoothedRSIIndicator (should normally based on wilders forumula)
- RSIIndicator

Please look at the computation of "**SmoothedRSIIndicator**" if it is equal to yours.

_By the way, normally Wilder uses his own MovingAverage in different kind of his formulas (called "WilderMovingAverage" which is nearly the same of an EMA, but it differs). However, we should be compliant to the original wilder formula. If this differs, we should fix it._

abzif:
SmoothedRSIIndicator is more like original formula but the computation is overly complicated.

I would rather introduce new helper inicator: MMAIndicator (modified moving average or welles wilder average) and use it to smooth GainIndicator and LossIndicator. Then AverageGainIndicator, AverageLossIndicator would be unnecessary. Also we could stay with one vesion of RSIIndicator.

Btw: ADX computations could also be simplified this way

nimo23:
> SmoothedRSIIndicator is more like original formula but the computation is overly complicated.
> Also we could stay with one vesion of RSIIndicator.
>ADX computations could also be simplified this way

Yes, good idea!! I think, also. Please make a PR with your solution.

team172011:
Changes requested

team172011:
#96 merged

PR #92, opened 2017-11-16
fix RSI indicator computation #89

abzif:
fix RSI indicator computation #89

added new helper indicator MMAIndicator (Welles Wilder moving average) + tests
added new helper indicators GainIndicator, LossIndicator + tests
simplify computation of RSIIndicator (AverageGain, AverageLoss etc are no longer necessary if MMAIndicator is used)
add comparative xls testing - indicator values are compared with values computed in excel, about 350 rows of data. Such tests should raise robustness of indicators.


ta4j-core/src/test/java/org/ta4j/core/XlsTestsUtils.java
```diff
@@ -0,0 +1,107 @@
...
+        // second param is in B4 cell
+        Row row = sheet.getRow(2 + paramIndex);
+        Cell cell = row.getCell(1);
+        cell.setCellValue(value);
+    }
+
+    public static TimeSeries readTimeSeries(Sheet sheet) {
+        // prices starts from 7th row until the end of sheet
+        // price data exists in first 5 columns (week date, open, high, low, close)
+        TimeSeries series = new BaseTimeSeries();
```
team172011: Please add columns for volume and amount

abzif: I can add column in xls for volume - just in case, some indicators use it
But what is amount? I don't know any indicator which uses it. I don't think it is necessary.

Unfortunately I didn't know about these branching rules in this project. I will change this PR.

team172011:
Merged in  #89

PR #94, opened 2017-11-18
Add: Z-Score-Indicator

nimo23:
- add statistic z-score-indicator

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **ConvergenceDivergenceIndicator**: New Indicator for positive/negative convergence and divergence.
 - **BooleanTransformIndicator**: New indicator to transform any decimal indicator to a boolean indicator by using logical operations.
 - **DecimalTransformIndicator**: New indicator to transforms any indicator by using common math operations.
+- **ZScoreIndicator**: New indicator to measure the statistic z-score.
 
 
 ## 0.10 (2017-10-23)
@@ -42,7 +43,6 @@ _Changed ownership of the ta4j repository_: from mdeverdelhan/ta4j (stopped the 
 - **AroonOscillatorIndicator**: new indicator based on AroonUp/DownIndicator
 - **AroonUpIndicator** and **AroonDownIndicator**: New constructor with parameter for custom indicator for min price and max price calculation
 - **ROCVIndicator**: rate of Change of Volume
-- **ZScoreIndicator**: the statistic z-score
 - **DirectionalMovementPlusIndicator**: new indicator for Directional Movement System (DMI+)
 - **DirectionalMovementDownIndicator**: new indicator for Directional Movement System (DMI-)
 - **ChaikinOscillatorIndicator**: new indicator.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ _Changed ownership of the ta4j repository_: from mdeverdelhan/ta4j (stopped the 
 - **AroonOscillatorIndicator**: new indicator based on AroonUp/DownIndicator
 - **AroonUpIndicator** and **AroonDownIndicator**: New constructor with parameter for custom indicator for min price and max price calculation
 - **ROCVIndicator**: rate of Change of Volume
+- **ZScoreIndicator**: the statistic z-score
 - **DirectionalMovementPlusIndicator**: new indicator for Directional Movement System (DMI+)
 - **DirectionalMovementDownIndicator**: new indicator for Directional Movement System (DMI-)
 - **ChaikinOscillatorIndicator**: new indicator.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **ConvergenceDivergenceIndicator**: New Indicator for positive/negative convergence and divergence.
 - **BooleanTransformIndicator**: New indicator to transform any decimal indicator to a boolean indicator by using logical operations.
 - **DecimalTransformIndicator**: New indicator to transforms any indicator by using common math operations.
-- **ZScoreIndicator**: New indicator to measure the statistic z-score.
+- **SigmaIndicator**: New indicator to measure the statistic z-score.
 
 
 ## 0.10 (2017-10-23)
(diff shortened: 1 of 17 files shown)


nimo23:
I am unsure about the z-score-values calculated with the help of "StandardDeviationIndicator.java" and "MeanDeviationIndicator.java".

When I have 6 close prices:

```
 @Before
    public void setUp() {
        data = new MockTimeSeries(1, 2, 3, 4, 5, 6);
    }
```
then, normally, the z-score must be between -3 and +3 standard deviations. However, it calculates a sd higher than 3. Is this correct?

nimo23:
Maybe we should rename "ZScoreIndicator" to "SigmaIndicator" because there is also a ZScoreIndicator which is not the typical statistical z-score (other names for z-score are "sigma" or "Standard Score").

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ _Changed ownership of the ta4j repository_: from mdeverdelhan/ta4j (stopped the
 - **AroonOscillatorIndicator**: new indicator based on AroonUp/DownIndicator
 - **AroonUpIndicator** and **AroonDownIndicator**: New constructor with parameter for custom indicator for min price and max price calculation
 - **ROCVIndicator**: rate of Change of Volume
+- **ZScoreIndicator**: the statistic z-score
```
team172011: Please add your entry to the unreleased section. This was for the last release 0.10

ta4j-core/src/main/java/org/ta4j/core/BaseStrategy.java
```diff
@@ -121,6 +129,11 @@ public Rule getExitRule() {
     public int getUnstablePeriod() {
     	return unstablePeriod;
     }
+	
+    @Override
+    public void setHint(String hint) {
+    	this.hint = hint;
+    }
```
team172011: Field is not necessary (at least not in the scope of this PR "Add: Z-Score-Indicator")

ta4j-core/src/main/java/org/ta4j/core/Decimal.java
```diff
@@ -175,6 +175,19 @@ public Decimal pow(int n) {
...
+     * Returns a {@code Decimal} whose value is <tt>(this<sup>n</sup>)</tt>.
+     * @param n power to raise this {@code Decimal} to.
+     * @return <tt>this<sup>n</sup></tt>
+     * @see BigDecimal#pow(int, java.math.MathContext)
+     */
+    public Decimal exp(int n) {
+        if (this == NaN) {
+            return NaN;
+        }
+        return Decimal.valueOf(Math.E).pow(n);
```
team172011: `Returns a {@code Decimal} whose value is <tt>(this<sup>n</sup>)</tt>. ` exactly what the javadoc of already existing `Decimal.pow(int n)` does

`Decimal.valueOf(Math.E).pow(n); ` why?

ta4j-core/src/main/java/org/ta4j/core/Strategy.java
```diff
@@ -35,6 +35,11 @@ this software and associated documentation files (the "Software"), to deal in
      */
     String getName();
 	
+    /**
+     * @return the description of the strategy
+     */
+    String getHint();
```
team172011: `hint` field in `Strategy` is not necessary

ta4j-core/src/main/java/org/ta4j/core/Decimal.java
nimo23: Please feel free to delete, I played with my repo and forgot to be within z-score-branch..The strange thing because of the https://github.com/ta4j/ta4j/issues/103 with the RecursiveCachedIndicator which gives me every time a stackoverflow error so slow calculations. I made something like this Double.valueOf(Math.exp()).. the need for conversion. So how can I prevent this "stackoverflow error" with my Indicators??

nimo23:
I delete "hint" from strategy and deleted exp from Decimal.

nimo23:
I played with my repo and did not leave z-score-branch..The strange thing because of the #103 with the RecursiveCachedIndicator which gives me every time a stackoverflow error so slow calculations. I made something like this Double.valueOf(Math.exp()).. the need for conversion. So how can I prevent this "stackoverflow error" with my Indicators??

ta4j-core/src/main/java/org/ta4j/core/Decimal.java
```diff
@@ -202,6 +202,84 @@ public Decimal sqrt() {
...
+     * @return the hyperbolic sine of {@code this}
+     * @see StrictMath#sinh(double)
+     */
+    public Decimal sinh() {
+        if (this == NaN) {
+            return NaN;
+        }
+        return new Decimal(StrictMath.sinh(delegate.doubleValue()));
+    }
+
```
team172011, line 282: So far `public Decimal sqrt()` is the only function that uses a _dirty_ double transformation... (used by 6 indicators at the moment). I would prefer not to implement this kind of math in the API of Decimal, but, if needed, in the corresponding indicator. Otherwise we support, like you said, nether accuracy nor performance

nimo23: You know my opinion. Sqrt/Exp and the like are common math and exist very long time in the world of mathematics. There is a reason for such formulas. It would be pitty not to have common math within this library. **As u see, it takes the whole story of "accuracy" ad absurdum.** You cannot guarantee this for indicators, and you really dont need to! If a indicator needs that tool, high accuracy is - at least there - no more satisifed. And I dont talk about all the other indicators which inherits this "unaccuracy" indicator within its formula. As u said - 6 indicators make direct use of them - and indirect maybe a lot more. This is only the beginning..there wil come more who wants to have common math within its formula and all their indicators have a loos of precision with the loos of performance. Please understand, you dont need such accuracy within indicators - **no need to measure "noise"**. I cannot find any use case where it matters. Think practically!

nimo23:
Values of "SigmaIndicator" are all correct.

nimo23:
ok done. I also opened a issue to clarify the answer about crypto-currencies and the need for accuracy.

nimo23:
> maybe with a note

Users accept high accuracy when using BigDecimal-Interface.  **A note is at least needed to tell the user that for this indicator high accuracy cannot be satisfied.**

nimo23:
Ok, can now be merged! In future, I will do separate PR for each major change. 

Please check the IsFalling/IsRisingRule, I have rewritten the whole logic.

PR #96, opened 2017-11-18
Develop

abzif:
fix RSI indicator computation #89

repeated pull request - this time from develop branch


ta4j-core/src/main/java/org/ta4j/core/indicators/RSIIndicator.java
```diff
@@ -24,53 +24,40 @@ this software and associated documentation files (the "Software"), to deal in
 
 import org.ta4j.core.Decimal;
 import org.ta4j.core.Indicator;
-import org.ta4j.core.indicators.helpers.AverageGainIndicator;
-import org.ta4j.core.indicators.helpers.AverageLossIndicator;
+import org.ta4j.core.indicators.helpers.GainIndicator;
+import org.ta4j.core.indicators.helpers.LossIndicator;
+import org.ta4j.core.indicators.helpers.MMAIndicator;
```
nimo23, line 29: Please move this to "org.ta4j.core.indicators", because it s a moving average indicator like sma or ema and not only a helper.

PR #106, opened 2017-11-21
Feature/decimal upgrades

briandilley:
Here it is again.  I fixed the `PivotPointIndicatorTest`.  I couldn't get the other test to fail that is failing on travis though.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/Decimal.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/Decimal.java
@@ -547,6 +547,9 @@ public final class Decimal
      * @return the {@code Decimal}
      */
     public static Decimal valueOf(float val) {
+        if (val == Float.NaN) {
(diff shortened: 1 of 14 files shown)


ta4j-core/src/main/java/org/ta4j/core/Decimal.java
```diff
@@ -72,24 +74,28 @@ private Decimal(String val) {
...
     }
 
     private Decimal(int val) {
-        delegate = BigDecimal.valueOf(val);
+        delegate = new BigDecimal(val, MATH_CONTEXT);
     }
 
     private Decimal(long val) {
         delegate = BigDecimal.valueOf(val);
     }
```
team172011, line 88: In  #84 the creation of Decimal based on `double`, `int` and `long` was changed to `valueOf(int | long | double )` to increase performance and accuracy. Why do you change back to `Decimal` constructor for int?

PR #120, opened 2017-11-30
allow limiting range of analysis calculations

### LINKED ISSUE #119: AnalysisCriterion start/finish indices
```
```
```
        fakeRecord.enter(startIndex);
        fakeRecord.exit(finishIndex);
    }
```
and changed the code to
```
```

edlins:
Fixes #119 .

- Add two methods which limit the calculation of buy-and-hold to a subset of the series
- Edit existing methods to call new methods with default getBeginIndex() and getEndIndex() values
- 

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
@@ -80,6 +80,8 @@ public class BaseTradingRecord implements TradingRecord {
         }
         this.startingType = entryOrderType;
         currentTrade = new Trade(entryOrderType);
+        this.startIndex = 0;
+        this.finishIndex = Integer.MAX_VALUE;
     }
 
     /**
@@ -141,7 +143,7 @@ public class BaseTradingRecord implements TradingRecord {
     }
     
     @Override
-    public final boolean exit(int index, Decimal price, Decimal amount) {
+    public boolean exit(int index, Decimal price, Decimal amount) {
         if (currentTrade.isOpened()) {
             operate(index, price, amount);
             return true;
@@ -151,12 +153,7 @@ public class BaseTradingRecord implements TradingRecord {
     
     @Override
     public List<Trade> getTrades() {
-    	if (trades.size() == 0) {
-    		return trades;
-    	}
-        return getTrades(
-        		trades.get(0).getEntry().getIndex(),
-        		trades.get(trades.size()-1).getExit().getIndex());
+        return trades;
     }
     
     @Override
--- a/ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
@@ -53,6 +53,12 @@ public class BaseTradingRecord implements TradingRecord {
 
     /** The entry type (BUY or SELL) in the trading session */
     private Order.OrderType startingType;
+
+    /** The starting index of the trading session */
+    private int startIndex;
+
+	/** The finishing index of the trading session */
+    private int finishIndex;
     
     /** The current non-closed trade (there's always one) */
     private Trade currentTrade;
@@ -74,6 +80,18 @@ public class BaseTradingRecord implements TradingRecord {
         }
         this.startingType = entryOrderType;
         currentTrade = new Trade(entryOrderType);
+    }
+
+    /**
+     * Constructor.
+     * @param entryOrderType the {@link Order.OrderType order type} of entries in the trading session
+     * @param startIndex the starting index of the trading session
+     * @param finishIndex the finish index of the trading session
+     */
+    public BaseTradingRecord(Order.OrderType entryOrderType, int startIndex, int finishIndex) {
+    	this(entryOrderType);
+        this.startIndex = startIndex;
+        this.finishIndex = finishIndex;
     }
 
     /**
@@ -133,7 +151,24 @@ public class BaseTradingRecord implements TradingRecord {
     
     @Override
     public List<Trade> getTrades() {
-        return trades;
+    	if (trades.size() == 0) {
+    		return trades;
+    	}
+        return getTrades(
+        		trades.get(0).getEntry().getIndex(),
+        		trades.get(trades.size()-1).getExit().getIndex());
+    }
+    
+    @Override
+    public List<Trade> getTrades(int beginIndex, int endIndex) {
+        List<Trade> trades = new ArrayList<Trade>();
+    	for (Trade trade : this.trades) {
+    		if (trade.getEntry().getIndex() < beginIndex || trade.getExit().getIndex() > endIndex) {
+    			continue;
+    		}
+    		trades.add(trade);
+    	}
+    	return trades;
     }
     
     @Override
@@ -203,4 +238,18 @@ public class BaseTradingRecord implements TradingRecord {
             currentTrade = new Trade(startingType);
         }
     }
+
+    /**
+     * @return the starting index of the trading session
+     */
+    public int getStartIndex() {
+		return startIndex;
+	}
+
+    /**
+     * @return the finishing index of the trading session
+     */
+	public int getFinishIndex() {
+		return finishIndex;
+	}
 }
--- a/ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
@@ -53,12 +53,6 @@ public class BaseTradingRecord implements TradingRecord {
 
     /** The entry type (BUY or SELL) in the trading session */
     private Order.OrderType startingType;
-
-    /** The starting index of the trading session */
-    private int startIndex;
-
-	/** The finishing index of the trading session */
-    private int finishIndex;
     
     /** The current non-closed trade (there's always one) */
     private Trade currentTrade;
@@ -80,20 +74,6 @@ public class BaseTradingRecord implements TradingRecord {
         }
         this.startingType = entryOrderType;
         currentTrade = new Trade(entryOrderType);
-        this.startIndex = 0;
-        this.finishIndex = Integer.MAX_VALUE;
-    }
-
-    /**
-     * Constructor.
-     * @param entryOrderType the {@link Order.OrderType order type} of entries in the trading session
-     * @param startIndex the starting index of the trading session
-     * @param finishIndex the finish index of the trading session
-     */
-    public BaseTradingRecord(Order.OrderType entryOrderType, int startIndex, int finishIndex) {
-    	this(entryOrderType);
-        this.startIndex = startIndex;
-        this.finishIndex = finishIndex;
     }
 
     /**
@@ -159,15 +139,15 @@ public class BaseTradingRecord implements TradingRecord {
     @Override
     public List<Trade> getTrades(int beginIndex, int endIndex) {
         List<Trade> trades = new ArrayList<Trade>();
-    	for (Trade trade : this.trades) {
-    		if (trade.getEntry().getIndex() < beginIndex || trade.getExit().getIndex() > endIndex) {
-    			continue;
-    		}
-    		trades.add(trade);
-    	}
-    	return trades;
+        for (Trade trade : this.trades) {
+            if (trade.getEntry().getIndex() < beginIndex || trade.getExit().getIndex() > endIndex) {
+                continue;
+            }
+            trades.add(trade);
+        }
+        return trades;
     }
-    
+
     @Override
     public Order getLastOrder() {
         if (!orders.isEmpty()) {
@@ -240,13 +220,19 @@ public class BaseTradingRecord implements TradingRecord {
      * @return the starting index of the trading session
      */
     public int getStartIndex() {
-		return startIndex;
+        if (trades.size() == 0 || trades.get(0).getEntry() == null) {
+            return 0;
+        }
+        return trades.get(0).getEntry().getIndex();
 	}
 
     /**
      * @return the finishing index of the trading session
      */
 	public int getFinishIndex() {
-		return finishIndex;
+	    if (trades.size() == 0 || trades.get(trades.size()-1).getExit() == null) {
+	        return 0;
+	    }
+	    return trades.get(trades.size()-1).getExit().getIndex();
 	}
 }
--- a/ta4j-core/src/main/java/org/ta4j/core/analysis/CashFlow.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/analysis/CashFlow.java
@@ -62,9 +62,7 @@ public class CashFlow implements Indicator<Decimal> {
      * @param tradingRecord the trading record
      */
     public CashFlow(TimeSeries timeSeries, TradingRecord tradingRecord) {
-        this.timeSeries = timeSeries;
-        calculate(tradingRecord, timeSeries.getBeginIndex(), timeSeries.getEndIndex());
-        fillToTheEnd();
+    	this(timeSeries, tradingRecord, timeSeries.getBeginIndex(), timeSeries.getEndIndex());
     }
 
     /**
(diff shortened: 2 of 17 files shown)


edlins:
> Please add a test case to TradingRecordTest.java

Sure, but there are different ways to do this.  

OPTION 1
The other tests call calculate() via AnalysisCriterion but I did not add this method to the interface so I used
```
((VersusBuyAndHoldCriterion)buyAndHold).calculate(Series, TradingRecord, int, int)
```

OPTION 2
Add this to AnalysisCriterion (misleading but easy)
```
    default double calculate(TimeSeries series, TradingRecord tradingRecord, int startIndex, int FinishIndex) {
    	return calculate(series, tradingRecord);
    }
```

OPTION 3
Add this to AnalysisCriterion AND implement the indexed version in every class that extends AbstractAnalysisCriterion (more correct but more changes)
```
    default double calculate(TimeSeries series, TradingRecord tradingRecord) {
    	return calculate(series, tradingRecord, series.getBeginIndex(), series.getEndIndex());
    }
    double calculate(TimeSeries series, TradingRecord tradingRecord, int startIndex, int FinishIndex);
```

OPTION 4
Fix TradingRecord by adding getStartIndex() and getFinishIndex()
AND fix BaseTradingRecord to store the indices in the constructor
AND fix TimeSeriesManager to pass the indices in the constructor
AND add the code below to AnalysisCriterion
AND implement calculate(TimeSeries, TradingRecord, int, int) in every class that extends AbstractAnalysisCriterion (most correct but most changes)
```
    default double calculate(TimeSeries series, TradingRecord tradingRecord) {
    	return calculate(series, tradingRecord, tradingRecord.getStartIndex(), tradingRecord.getFinishIndex());
    }
    double calculate(TimeSeries series, TradingRecord tradingRecord, int startIndex, int FinishIndex);
```

I'm inclined to go with option 1 for now.  But I think IF it's possible to make a TradingRecord with TimeSeriesManager#run(Strategy, int, int) THEN those int's should be captured to indicate that part of the Series may have been ignored in the TradingRecord's creation.  I think the best fix for that is option 4.  Otherwise your analysis may be comparing criteria based on different series data.

team172011:
Yes, I also think option 4 is coherent

edlins:
> Yes, I also think option 4 is coherent

I've fixed this as option 1 for now.  I'll open a new issue for implementing option 4 as it is much more involved.

team172011:
I would rather skip option 1 because, as you say, the comparison of different criteria could be inconsistent. You do not have to close this PR, but you can commit further on to this branch. If merging, your commits will be squashed to one single commit. So we can support you during the implementation

edlins:
Okay, 17 files changed.  Being rather new to the project I would appreciate a close review.  That was a lot of changes just to call AnalysisCriterion#calculate() with indices.  With this, all analyses can be range limited but default to the entire series or the entire trading record (depending on the analysis.)

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
```diff
@@ -133,7 +151,24 @@ public final boolean exit(int index, Decimal price, Decimal amount) {
     
     @Override
     public List<Trade> getTrades() {
-        return trades;
+    	if (trades.size() == 0) {
+    		return trades;
+    	}
+        return getTrades(
+        		trades.get(0).getEntry().getIndex(),
+        		trades.get(trades.size()-1).getExit().getIndex());
```
team172011: Whats the difference between returning this.trades and returning all of this.trades that are between this.trades.first entry and this.trades.last entry?

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
```diff
@@ -53,6 +53,12 @@ this software and associated documentation files (the "Software"), to deal in
 
     /** The entry type (BUY or SELL) in the trading session */
     private Order.OrderType startingType;
+
+    /** The starting index of the trading session */
+    private int startIndex;
+
+	/** The finishing index of the trading session */
+    private int finishIndex;
```
team172011: - These fields are never used locally and just needed to clarify what start and end of this trading record is, or? There are two constructors that don't initialize the fields. -> initialize them or handle it in the getters (example in getter comment)

- The `TimeSeriesManager` creates `new BaseTradingRecord` with corresponding **startIndex** and **finishIndex**  and **just** records orders that are **between** **startIndex** and **finishIndex** we deliver in `TimeSeriesManager#run(order, amount, startIndex, endIndex) ` method -> that looks valid.
But are there other situation in which an order can be insert into `BaseTradingRecord` that is out of scope (not between **startIndex** and **finishIndex**)? I could imagine in a custom extend of `TimeSeriesManager` one could try to do so. Should this be possible? I think this would lead to an inconsistence state (a `BaseTradingRecord` holds orders that are out of its range(startIndex, finishIndex)) that we should not allow. -> Maybe we could use those fields to check in `enter(index, price, amount)` and `exit(index, price, amount)` if the order is in scope?

- Please remove final keyword from `exit(..)` method, we don't need this anymore

team172011: Other possibility would be to remove `startIndex` and `finishIndex` fields and do not let the `BaseTradingRecord` decide which orders are valid or not. But I would prefer your solution with `startIndex` and `finishIndex` fields, because it is more transparent which frame this records covers.

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
```diff
@@ -203,4 +238,18 @@ private void recordOrder(Order order, boolean isEntry) {
...
+    public int getStartIndex() {
+		return startIndex;
+	}
+
+    /**
+     * @return the finishing index of the trading session
+     */
+	public int getFinishIndex() {
+		return finishIndex;
+	}
```
team172011: `getStartIndex()` and ` getFinishIndex()` return a null pointer if this `BaseTradingRecord` was not initialized by your new constructor, but with one of the old once that do not initialize the fields.
- You could initialize them with `Integer.MAX_VALUE`
- or check in getter` if startIndex == null return Integer.MAX_VALUE`

team172011: I would prefer to initialize them with `finishIndex = Integer.MAX_VALUE` and `startIndex = 0`

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
edlins: > These fields are never used locally and just needed to clarify what start and end of this trading record is

Yes, and the new indices are only meaningful if the new constructor is used.  I will add inits of 0 and Integer.MAX_VALUE to the old constructor(s).  The new constructor indices are simply stored and do not internally impact logic in any way.  My only current use of the indices is for calls like this first one
```
buyAndHold.calculate(series, tradingRecord, tradingRecord.getStartIndex(), tradingRecord.getFinishIndex()); // new, limit range of series
buyAndHold.calculate(series, tradingRecord, currentIndex-199, currentIndex); // new, limit range to previous 200 indices
buyAndHold.calculate(series, tradingRecord); // old behavior, use entire range of series
```
Also, one could use the old constructors but evaluate VersusBuyAndHold for the previous 200-tick windows like the second above.  CAVEAT: that probably requires better handling of trades spanning across either index (like fake openOrder at startIndex and closeOrder at finishIndex.)  But that is all new functionality that won't break anything existing as it gets worked out.

So, the constructor indices are not coupled to the calculate() indices unless you couple them with the getters as above. [EDIT: see my latest comment re always using {start|finish}Index.]

> an order can be insert into BaseTradingRecord that is out of scope

I think that's fine.  To me, scope is defined by the call to calculate(), not by the record constructor.  Add orders outside of startIndex and finishIndex.  Call the third example above and they are included.  Call the first example and they are excluded (but see the CAVEAT.)  In fact, I'll probably add these cases to the unit test to demonstrate. [EDIT: cases added as comments]

Finally, if your series manager calls the new constructor with indices inside run(), then it's your responsibility to only operate() between those indices.  You don't have to keep them locally in run() though, as you can then use the getters.  If you operate() outside of your calculate() indices (or calculate() outside your operate() indices), you are probably just wasting cycles.

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
edlins: I wanted to make getTrades() use getTrades(int, int) for re-use but that's messy.  Reverted to return this.trades.

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
edlins: Now initialized in BaseTradingRecord#BaseTradingRecord(Order.OrderType), which is in the constructor chain for all other constructors.  When constructed with indices, those fields are set after being initialized as above.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/VersusBuyAndHoldCriterion.java
```diff
@@ -43,11 +43,20 @@ public VersusBuyAndHoldCriterion(AnalysisCriterion criterion) {
 
     @Override
     public double calculate(TimeSeries series, TradingRecord tradingRecord) {
+        if (tradingRecord.getStartIndex() == 0 && tradingRecord.getFinishIndex() == 0) {
+            // this is only here to satisfy the old test case calculateWithNoTrades()
+            return calculate(series, tradingRecord, series.getBeginIndex(), series.getEndIndex());
```
edlins, line 48: This mimics the old behavior to pass the calculateWithNoTrades() unit test

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/VersusBuyAndHoldCriterionTest.java
```diff
@@ -66,6 +89,9 @@ public void calculateWithNoTrades() {
         MockTimeSeries series = new MockTimeSeries(100, 95, 100, 80, 85, 70);
 
         AnalysisCriterion buyAndHold = new VersusBuyAndHoldCriterion(new TotalProfitCriterion());
+        // per the logic that strategies are only evaluated for closed trades, this should be 1d like so:
+        //assertEquals(1d, buyAndHold.calculate(series, new BaseTradingRecord()), TATestsUtils.TA_OFFSET);
```
edlins, line 93: I strongly feel this should be 1d

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
```diff
@@ -136,6 +136,18 @@ public final boolean exit(int index, Decimal price, Decimal amount) {
         return trades;
     }
     
+    @Override
+    public List<Trade> getTrades(int beginIndex, int endIndex) {
+        List<Trade> trades = new ArrayList<Trade>();
+        for (Trade trade : this.trades) {
+            if (trade.getEntry().getIndex() < beginIndex || trade.getExit().getIndex() > endIndex) {
```
edlins, line 143: null check?

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
```diff
@@ -203,4 +215,24 @@ private void recordOrder(Order order, boolean isEntry) {
             currentTrade = new Trade(startingType);
         }
     }
+
+    /**
+     * @return the starting index of the trading session
+     */
+    public int getStartIndex() {
+        if (trades.size() == 0 || trades.get(0).getEntry() == null) {
```
edlins: null check?

ta4j-core/src/main/java/org/ta4j/core/analysis/CashFlow.java
```diff
@@ -115,8 +124,11 @@ private void calculate(Trade trade) {
      * Calculates the cash flow for a trading record.
      * @param tradingRecord the trading record
      */
-    private void calculate(TradingRecord tradingRecord) {
+    private void calculate(TradingRecord tradingRecord, int beginIndex, int endIndex) {
         for (Trade trade : tradingRecord.getTrades()) {
+            if (trade.getEntry().getIndex() < beginIndex || trade.getExit().getIndex() > endIndex) {
```
edlins, line 129: null check?

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterion.java
```diff
@@ -72,11 +72,14 @@ public double calculate(TimeSeries series, Trade trade) {
     }
 
     @Override
-    public double calculate(TimeSeries series, TradingRecord tradingRecord) {
+    public double calculate(TimeSeries series, TradingRecord tradingRecord, int beginIndex, int endIndex) {
         double totalCosts = 0d;
         double tradedAmount = initialAmount;
         
         for (Trade trade : tradingRecord.getTrades()) {
+            if (trade.getEntry().getIndex() < beginIndex || trade.getExit().getIndex() > endIndex) {
```
edlins, line 80: null check?

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfBarsCriterion.java
```diff
@@ -34,9 +34,12 @@ this software and associated documentation files (the "Software"), to deal in
 public class NumberOfBarsCriterion extends AbstractAnalysisCriterion {
 
     @Override
-    public double calculate(TimeSeries series, TradingRecord tradingRecord) {
+    public double calculate(TimeSeries series, TradingRecord tradingRecord, int beginIndex, int endIndex) {
         int nBars = 0;
         for (Trade trade : tradingRecord.getTrades()) {
+            if (trade.getEntry().getIndex() < beginIndex || trade.getExit().getIndex() > endIndex) {
```
edlins, line 40: null check?

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/TotalProfitCriterion.java
```diff
@@ -35,9 +35,12 @@ this software and associated documentation files (the "Software"), to deal in
 public class TotalProfitCriterion extends AbstractAnalysisCriterion {
 
     @Override
-    public double calculate(TimeSeries series, TradingRecord tradingRecord) {
+    public double calculate(TimeSeries series, TradingRecord tradingRecord, int beginIndex, int endIndex) {
         double value = 1d;
         for (Trade trade : tradingRecord.getTrades()) {
+            if (trade.getEntry().getIndex() < beginIndex || trade.getEntry().getIndex() > endIndex) {
```
edlins, line 41: null check?

edlins:
Okay, so maybe ignore most of my previous comments..  This works without storing the indices (with the exception of one degenerate test case) by using this in the AnalysisCriterion interface
```
default double calculate(TimeSeries series, TradingRecord tradingRecord) {
    return calculate(series, tradingRecord, tradingRecord.getStartIndex(), tradingRecord.getFinishIndex());
}
```
where **get{Start|Finish}Index()** just looks at the first trade entry and last trade exit.  Pretty simple and automatic.  Have a look at **VersusBuyAndHoldCriterionTest.java** to see the changes in action, and my notes in **calculateWithNoTrades()** and **VersusBuyAndHoldCriterion.java**.  I also have minor concerns about null checks on things like **trades.get().getEntry().getIndex()**.  But I think this is much closer to the solution.

Still not ready for a merge because I want to play out one more idea..

edlins:
Please cancel in favor of #130 which is a better way.

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
team172011: no null check needed. If user wants to get trades between null and/or null he will get his npe

edlins:
Not ready for merging, closing pull request.

PR #128, opened 2017-12-06
Pearson indicator

nimo23:
- PearsonCorrelationIndicator

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/PearsonCorrelationIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/PearsonCorrelationIndicator.java
@@ -1,0 +1,70 @@
+package org.ta4j.core.indicators.statistics;
+
+import org.ta4j.core.Decimal;
+import org.ta4j.core.Indicator;
+import org.ta4j.core.indicators.RecursiveCachedIndicator;
+
+/**
+ * Indicator-Pearson-Correlation
+ * <p/>
+ * see
+ * http://www.statisticshowto.com/probability-and-statistics/correlation-coefficient-formula/
+ */
+public class PearsonCorrelationIndicator extends RecursiveCachedIndicator<Decimal> {
+
+	private static final long serialVersionUID = 6317147143504055664L;
+	
+	private final Indicator<Decimal> indicator1;
+	private final Indicator<Decimal> indicator2;
+	private final int timeFrame;
+
+	/**
+	 * Constructor.
+	 * 
+	 * @param indicator1 the first indicator
+	 * @param indicator2 the second indicator
+	 * @param timeFrame the time frame
+	 */
+	public PearsonCorrelationIndicator(Indicator<Decimal> indicator1, Indicator<Decimal> indicator2, int timeFrame) {
+		super(indicator1);
+		this.indicator1 = indicator1;
+		this.indicator2 = indicator2;
+		this.timeFrame = timeFrame;
+	}
+	
+
+	@Override
+	protected Decimal calculate(int index) {
+
+		Decimal n = Decimal.valueOf(timeFrame);
+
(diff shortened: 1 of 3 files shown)


nimo23:
Here you see, I must make use of "Math.sqrt" to implement PearsonCorrelationIndicator. The question is, does it matter to calculate the pearson correlation with double and only make the return type Decimal by return Decimal.valueOf(doubleValueOfPearson).

Would be a way faster when calculation is made in double and return type is Decimal. So it can be compared against other without risk. Or?
```
	@Override
		double Sx = 0;
		double Sy = 0;
		double Sxx = 0;
		double Syy = 0;
		double Sxy = 0;

		for (int i = Math.max(0, index - timeFrame + 1); i <= index; i++) {
     		        double x = indicator1.getValue(i).toDouble();
			double x = indicator1.getValue(i).toDouble();
			Sx = Sx + x;
			Sy = Sy + y;
			Sxx = Sxx + x * x;
			Sxy = Sxy + x * y;
			Syy = Syy + y * y;
		}
		if ((m * Sxx - Sx * Sx) * (m * Syy - Sy * Sy) > 0) {
			double term = (m * Sxy - Sx * Sy) / Math.sqrt((m * Sxx - Sx * Sx) * (m * Syy - Sy * Sy));
			return Decimal.valueOf(term);
		}

		return Decimal.NaN;
	}
```

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/PearsonCorrelationIndicator.java
```diff
@@ -0,0 +1,70 @@
...
+
+		Decimal n = Decimal.valueOf(timeFrame);
+
+		Decimal Sx = Decimal.ZERO;
+		Decimal Sy = Decimal.ZERO;
+		Decimal Sxx = Decimal.ZERO;
+		Decimal Syy = Decimal.ZERO;
+		Decimal Sxy = Decimal.ZERO;
+		
+		for (int i = Math.max(0, index - timeFrame + 1); i <= index; i++) {
```
team172011: Use getTimeSeries().getBeginIndex() instead of zero :)

nimo23:
Ok, changed, if "getTimeSeries()" is always available, we can use this:

```
for(int i = Math.max(getTimeSeries().getBeginIndex(), index - timeFrame + 1); i <= index; i++) {
..}
```

I think there is no need to check if "timeSeries" is not null. Or should we use this to be on the safe side?:

```
for(int i = Math.max(getTimeSeries()!= null : getTimeSeries().getBeginIndex() : 0, index - timeFrame + 1); i <= index; i++) {
..}
```

team172011:
No need to check if TimeSeries is null. If TimeSeries is null, also index 0 is null

nimo23:
>No need to check if TimeSeries is null. If TimeSeries is null, also index 0 is null

**This is not true! **

if BaseTimeSeries is null, then object is null. It does not matter if beginIndex is primitive type. NPE is thrown, you can easily check by yourself:
```

BaseTimeSeries s = null;
log.info("s.getBeginIndex()={}", s.getBeginIndex());
```

team172011:
If you iterate over TimeSeries (or another indicator that is based on a TimeSeris) it doesn't matter if you are using` timeSeries.getBeginIndex()` or if you say index = 0 regarding checking for null pointer. Both will throw npe if `TimeSeries` is null,` getBeginIndex()` immediately and index=0 if you call the first time` getBar(index)`. So far we do not check every time `if getBar(index) != null` and so we also don't need to check `if timeSeries != null` if using ` timeSeries.getBeginIndex()`. If the user tries to use the calculate method of indicator on a TimeSeries object that is a null pointer he will get the npe. Nothing to do here,  let it be thrown

nimo23:
> Nothing to do here, let it be thrown
yes, right.

Issue #125, opened 2017-12-04
"constrained time" series vs "subseries"


nimo23:
I have a timeseries s and want to create a new subseries s1 from it. s has 20 ticks with index 0 to 19. s1 should be a subseries from s containing only the last 10 ticks from the actual index backwards.

`TimeSeries s1 = new BaseTimeSeries(s, index - 10, index);`

However, after printing "s1.getTickData()" and "series.getTickData()", both timeseries has the same ticks included. s1 normally should have only 10 ticks included and not 20. What is wrong?

team172011:
Because it calls internally:
`this(defaultSeries.getName(), defaultSeries.getBarData(), seriesBeginIndex, seriesEndIndex, true);`

All Ticks will be added to the subseries, just the begin and end index are modified 

**Javadoc of TimeSeries#getBarData() (getTickData() in master):**

```
    /**
     * Warning: should be used carefully!
     * <p>
     * Returns the raw bar data.
     * It means that it returns the current List object used internally to store the {@link Bar bars}.
     * It may be:
     *   - a shortened bar list if a maximum bar count has been set
     *   - a extended bar list if it is a constrained time series
     * @return the raw bar data
     */
    List<Bar> getBarData();
```

nimo23:
So s1 contains all ticks of s but only the ticks from "seriesBeginIndex" to  "seriesEndIndex" will be considered by any other indicators (for example, highestValueIndicator). Am I right?
"
```
// returns the highest value within index - 10 to index
new HighestValueIndicator(s1);
```

```
// returns the highest value within the whole timeSeries.
new HighestValueIndicator(s);
```

team172011:
Yes, I think so. As far the indicators use `getBeginIndex()`. I am unsure what happens if you call `s1.getBar(0)`.

team172011:
Many indicators use statements like 
`int index = Math.max(0. value) `
to prevent negative values for timeFrame or so ... If working on a subseries it must be
`int index = Math.max(series.getBeginIndex, value)`


Example in ArronUp indicator:
`int endIndex = Math.max(0,index - timeFrame);`

nimo23:
So normally such calls within indicators

`int endIndex = Math.max(0,index - timeFrame);`

should always be done this way:

`int endIndex = Math.max(series.getBeginIndex, value)`

Or?

It does not matter if subseries or not. If series is not a subseries, then "series.getBeginIndex" will also return "0". But for subseries, it returns the correct index as declared.

[...]

PR #129, opened 2017-12-06
Added the Donchian channel indicator

jnidzwetzki:
- add Donchian channel indicator

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/DonchianChannelLower.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/DonchianChannelLower.java
@@ -1,0 +1,78 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2014-2017 Marc de Verdelhan & respective authors (see AUTHORS)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/DonchianChannelLower.java
```diff
@@ -0,0 +1,78 @@
...
+	public DonchianChannelLower(Indicator<Decimal> indicator, int timeFrame) {
+        super(indicator.getTimeSeries());
+		this.indicator = indicator;
+		this.timeFrame = timeFrame;
+    }
+
+	
+	@Override
+	protected Decimal calculate(int index) {
+		int startIndex = Math.max(0, index - timeFrame + 1);
```
team172011, line 58: Please use `this.getTimeSeries().getBeginIndex()` instead of 0. The underlying `TimeSeries` can be in a state where the first bar is not equal to the index 0.  (This has to be fixed in other indicators, too. see #125)

team172011: What to do if 
- `index-this.getTimeSeries().getBeginIndex() < timeFrame` (not enough data for correct calculation) -> return NaN 

(needs also to be fixed for other indicators, see #125)

ta4j-core/src/main/java/org/ta4j/core/indicators/DonchianChannelLower.java
```diff
@@ -0,0 +1,78 @@
...
+		
+		for(int pos = startIndex; pos <= index; pos++) {
+			final Decimal value = indicator.getValue(pos);
+			
+			if(value.isLessThanOrEqual(result)) {
+				result = value;
+			}
+		}
+		
+		return result;
```
team172011, line 70: You could also instantiate a `new LowestValueIndicator(indicator, timeFrame)` in the constructor as class variable and write
 
```
@Override
    this.lowestValueIndiator.getValue(index);
}
```

Same in DonchianChannelUpper indicator

team172011:
And I see you wants to push from your `master` into ta4j `master`. Regarding [how to contribute](https://github.com/ta4j/ta4j/wiki/How-to-contribute) PRs to master will not be merged. Please make a new PR from your updated `develop` branch to ta4js `develop` branch

jnidzwetzki:
Hi,

I had not seen that a similar behavior could be achieved by the Lowest-/Highest-ValueIndicator. So, I would suggest to close my PR as it does not contain a really new indicator.

Best Regards
   Jan

PR #130, opened 2017-12-08
Fix/limit ranges

### LINKED ISSUE #119: AnalysisCriterion start/finish indices

edlins:
Fixes #119 .

- Limit [Versus]BuyAndHold calculations to range of trading record
- Fix test cases with no trades to return 1d
- 

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/BuyAndHoldCriterion.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/BuyAndHoldCriterion.java
@@ -35,13 +35,12 @@ public class BuyAndHoldCriterion extends AbstractAnalysisCriterion {
 
     @Override
     public double calculate(TimeSeries series, TradingRecord tradingRecord) {
-        //return series.getBar(series.getEndIndex()).getClosePrice().dividedBy(series.getBar(series.getBeginIndex()).getClosePrice()).doubleValue();
         if (tradingRecord.getTradeCount() == 0) {
             return 1.0d;
         }
         int firstEntryIndex = tradingRecord.getTrades().get(0).getEntry().getIndex();
         int lastExitIndex = tradingRecord.getTrades().get(tradingRecord.getTradeCount()-1).getExit().getIndex();
-        
+
         return series.getBar(lastExitIndex).getClosePrice().dividedBy(series.getBar(firstEntryIndex).getClosePrice()).doubleValue();
     }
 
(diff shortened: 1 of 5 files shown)


edlins:
This solves #119  in a very simple manner.  The only contentious changes are to the test cases where the trading record has no closed trades.  Per Marc in #125 , trades that aren't closed should not affect the score of a strategy so this should be a correct fix.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/BuyAndHoldCriterion.java
```diff
@@ -35,7 +35,13 @@ this software and associated documentation files (the "Software"), to deal in
...
     @Override
     public double calculate(TimeSeries series, TradingRecord tradingRecord) {
-        return series.getBar(series.getEndIndex()).getClosePrice().dividedBy(series.getBar(series.getBeginIndex()).getClosePrice()).doubleValue();
+        if (tradingRecord.getTradeCount() == 0) {
+            return 1.0d;
+        }
+        int firstEntryIndex = tradingRecord.getTrades().get(0).getEntry().getIndex();
+        int lastExitIndex = tradingRecord.getTrades().get(tradingRecord.getTradeCount()-1).getExit().getIndex();
+
+        return series.getBar(lastExitIndex).getClosePrice().dividedBy(series.getBar(firstEntryIndex).getClosePrice()).doubleValue();
```
team172011, line 44: Is it correct to say that the range of the `TradingRecord` is just from first entry to last exit index of its trades? Maybe the run was over a much bigger range, but the first trade started after some bars or the last exit was earlier. In my opinion in most cases this assumption is not valid, the `TradingRecord` needs start and end indices that are set in TimeSeries#run()

edlins:
You are correct (again.)  I think I've got the correct solution and I'm spending extra time developing robust test cases.  I'll push another commit as soon as I can.

edlins:
This is more complicated than I thought.  When you introduce indices, you are introducing a context against which everything is evaluated.  But TradingRecord has a fixed context based on the chronological sequence of adding trades.  The whole concept of a current trade depends on the context, but TradingRecord#getCurrentTrade() does not account for an arbitrary context.  So, when looking at different subranges of a TradingRecord, getCurrentTrade() will always return the same Trade even if it's outside of the subrange.  For example, you have a TradingRecord with a Trade that buys at index 0 and sells at index 10.  Now you calculate from index 0 to index 5.  The calculation views currentTrade.isNew() as true, isOpened() as false, and isClosed() as false.  It should see !isNew(context), isOpened(context), and !isClosed(context).  All of the context-dependent methods will need the context indices.  This would take a pretty serious re-write of TradingRecord and BaseTradingRecord.

edlins:
This is not ready for merging so I will close it until it's ready.  I'll keep the branch updated in my fork if you want to follow along.

PR #131, opened 2017-12-10
Fix/linear transaction cost

edlins:
Fixes #.

- Fix calculation of linear transaction cost
- Fix unit test cases
- 

### DIFF
--- a/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
+++ b/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
@@ -40,6 +40,8 @@ public class LinearTransactionCostCriterionTest {
         AnalysisCriterion transactionCost = new LinearTransactionCostCriterion(amount, a, b);
 
         TradingRecord tradingRecord = new BaseTradingRecord(Order.buyAt(0, series), Order.sellAt(1, series));
+        // bad form
+        assertEquals(12.861, transactionCost.calculate(series, tradingRecord), TATestsUtils.TA_OFFSET);
 
         double entryCost = amount * a + b;
         double profit = series.getBar(1).getClosePrice().dividedBy(series.getBar(0).getClosePrice()).doubleValue();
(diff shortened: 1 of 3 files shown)


edlins:
I really don't like assertEquals() with some hardcopied mystery number.  If you don't show your work it's really hard to validate.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
```diff
@@ -34,17 +34,48 @@ this software and associated documentation files (the "Software"), to deal in
...
+        double exitCost = amount * a + b;
+        amount = amount - exitCost;
+        double totalCost = entryCost + exitCost;
+        System.out.println(entryCost + " " + profit + " " + amount + " " + exitCost + " " + totalCost);
+        // good form
+        assertEquals(totalCost, transactionCost.calculate(series, tradingRecord), TATestsUtils.TA_OFFSET);
+
         tradingRecord.operate(2);
         tradingRecord.operate(3);
-        assertEquals(24.3473, transactionCost.calculate(series, tradingRecord), TATestsUtils.TA_OFFSET);
```
edlins, line 44: not very useful, and wrong

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
```diff
@@ -34,17 +34,48 @@ this software and associated documentation files (the "Software"), to deal in
...
+        assertEquals(24.3759, transactionCost.calculate(series, tradingRecord), TATestsUtils.TA_OFFSET);
```
edlins, line 58: not very useful, but correct

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
```diff
@@ -34,17 +34,48 @@ this software and associated documentation files (the "Software"), to deal in
...
+
+        entryCost = amount * a + b;
+        profit = series.getBar(3).getClosePrice().dividedBy(series.getBar(2).getClosePrice()).doubleValue();
+        amount = (amount - entryCost) * profit;
+        exitCost = amount * a + b;
+        amount = amount - exitCost;
+        totalCost += entryCost + exitCost;
+        assertEquals(totalCost, transactionCost.calculate(series, tradingRecord), TATestsUtils.TA_OFFSET);
```
edlins, line 66: much better

team172011:
> I really don't like assertEquals() with some hardcopied mystery number. If you don't show your work it's really hard to validate.

Not the best way to validate correctness, but still a good solution to satisfy that everything else works as before after one pushed...

Since #111 we also have the possibility of excel sheet validation

edlins:
I agree only that it's a check to make sure the results didn't change, but says nothing about correctness. 
 If the original developer has the algorithm wrong, they might do it wrong in the method, the unit test, and Excel.  It's best if a different person writes the unit tests, and they should do so in a way that demonstrates the logic in a transparent manner.  And the unit tests should be written before the methods..  Regardless, there should be a very high standard for thorough and transparent unit tests in a technical analysis library.  In fact, I care much more about clear unit tests than I do about clear method code.  I happily offer this unit test for consideration as a template for analysis criteria unit tests.

edlins:
Will rewrite with xls validation.

Issue #103, opened 2017-11-21
Improve caching system


nimo23:
Maybe we should merge these abstract classes. I dont know exactly, when to use "RecursiveCachedIndicator" and when to use "CachedIndicator". 

As described in recursiveCachedIndicator:

>This class is only here to avoid (OK, to **postpone** ???!) the StackOverflowError that may be thrown on the first getValue(int) call of a recursive indicator.

Looks for me like a workaround because of the caching layer..hence we should improve "CachedIndicator" to make the need for this "workaround" obsolete.

As the caching layer should also be improved, we could remove the "RecursiveCachedIndicator ". Maybe we can merge it all to one central point, so all indicators need to extend only one, call it "AbstractIndicator" or **"BaseIndicator"**.

nimo23:
There is another problem with the RecursiveCacheIndicator:

when calling this within my calculate():
```

Decimal calcPrev = getValue(index - 1);  
Decimal calcPrev2 = getValue(index - 2);

log.info("calcPrev: {}", outputPrev1);
log.info("calcPrev2: {}", outputPrev2);
```

then I get sometimes for both of the values "0" which should be not possible!
Because there were a calculated value at these previous states.

Why is that so?

nimo23:
Look at this:
@Override
	protected Decimal calculate(int index) {
		if (index <= 0) {
			return Decimal.ZERO;
		}

                Decimal input = ref.getValue(index);
		Decimal inputPrev1 = ref.getValue(index - 1);
		Decimal period = Decimal.valueOf(timeFrame);
		
		Decimal outputPrev1 = getValue(index - 1);
		Decimal outputPrev2 = getValue(index - 2);
		
		log.info("input: {}", input);
		log.info("inputPrev1: {}", ref.getValue(index - 1));
		log.info("period: {}", Decimal.valueOf(timeFrame));
		log.info("outputPrev1: {}", outputPrev1);
		log.info("outputPrev2: {}", outputPrev2);
..}

I get this log, which is normally not possible:
- wrong previous values from input (`ref(index-x)`)
- wrong getPreviousValues from calcualtion of this indicator (`getValue(x)`)

```
21:43:42,208 INFO  input: 16.359999999999999431565811391920
21:43:42,208 INFO  inputPrev1: 16.315000000000001278976924368180
21:43:42,208 INFO  period: 10
21:43:42,209 INFO outputPrev1: 70245.393287345490534789860248566
[...]

Issue #90, opened 2017-11-15
Perf - remove calls to getSimpleName() from log.trace()


edlins:
AbstractRule and BaseStrategy both have calls to getClass().getSimpleName() inside log.trace() calls.  When working with lots of data this causes a significant performance hit due to being called many times.  Change this to use a final String class variable that's set once and subsequently used in the log.trace() calls.

See #75

edlins:
#91 created.

team172011:
thanks, merged!

PR #144, opened 2017-12-21
performance improvement

Mulder8472:
according to visualvm, these "traceXYZ" methods used up to 15% of CPU time during my backtesting runs.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/BaseStrategy.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BaseStrategy.java
@@ -182,7 +182,7 @@ public class BaseStrategy implements Strategy {
      */
     protected void traceShouldEnter(int index, boolean enter) {
         if ( log.isTraceEnabled() ) {
-            log.trace(">>> {}#shouldEnter({}): {}", getClass().getSimpleName(), index, enter);
+            log.trace(">>> {}#shouldEnter({}): {}", className, index, enter);
         }
     }
 
@@ -193,7 +193,7 @@ public class BaseStrategy implements Strategy {
      */
     protected void traceShouldExit(int index, boolean exit) {
         if ( log.isTraceEnabled() ) {
-            log.trace(">>> {}#shouldExit({}): {}", getClass().getSimpleName(), index, exit);
+            log.trace(">>> {}#shouldExit({}): {}", className, index, exit);
         }
     }
 }
--- a/ta4j-core/src/main/java/org/ta4j/core/BaseStrategy.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BaseStrategy.java
@@ -181,7 +181,9 @@ public class BaseStrategy implements Strategy {
      * @param enter true if the strategy should enter, false otherwise
      */
     protected void traceShouldEnter(int index, boolean enter) {
-        log.trace(">>> {}#shouldEnter({}): {}", className, index, enter);
+        if ( log.isTraceEnabled() ) {
+            log.trace(">>> {}#shouldEnter({}): {}", getClass().getSimpleName(), index, enter);
+        }
     }
 
     /**
@@ -190,6 +192,8 @@ public class BaseStrategy implements Strategy {
      * @param exit true if the strategy should exit, false otherwise
      */
     protected void traceShouldExit(int index, boolean exit) {
-        log.trace(">>> {}#shouldExit({}): {}", className, index, exit);
+        if ( log.isTraceEnabled() ) {
+            log.trace(">>> {}#shouldExit({}): {}", getClass().getSimpleName(), index, exit);
+        }
     }
 }
--- a/ta4j-core/src/main/java/org/ta4j/core/trading/rules/AbstractRule.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/trading/rules/AbstractRule.java
@@ -43,6 +43,8 @@ public abstract class AbstractRule implements Rule {
      * @param isSatisfied true if the rule is satisfied, false otherwise
      */
     protected void traceIsSatisfied(int index, boolean isSatisfied) {
-        log.trace("{}#isSatisfied({}): {}", className, index, isSatisfied);
+        if ( log.isTraceEnabled() ) {
+            log.trace("{}#isSatisfied({}): {}", getClass().getSimpleName(), index, isSatisfied);
+        }
     }
 }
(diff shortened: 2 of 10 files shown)


ta4j-core/src/main/java/org/ta4j/core/BaseStrategy.java
```diff
@@ -181,7 +181,9 @@ public Strategy or(String name, Strategy strategy, int unstablePeriod) {
      * @param enter true if the strategy should enter, false otherwise
      */
     protected void traceShouldEnter(int index, boolean enter) {
-        log.trace(">>> {}#shouldEnter({}): {}", className, index, enter);
+        if ( log.isTraceEnabled() ) {
+            log.trace(">>> {}#shouldEnter({}): {}", getClass().getSimpleName(), index, enter);
```
edlins: These were changed to use the string className instead of calling getSimpleName() on every single entry and exit.  IMO trace() should minimize or eliminate calls to other methods due to the potential frequency of execution.  See #90

codehustler: of course, i missed that change, updated the pr. log4j supports lambda logging, that eliminated the need for those ugly if blabla is enabled before the actual log statement, slf4j doent have that yet, so i think in this case it is still a valid performance improvement. it had a significant impact for me running parallel backtests, this methods used 15% CPU and now its arround 0.1% (trace disabled)

edlins: You only see this 15% consumption with trace logging enabled, correct?  If you have trace logging enabled you really shouldn't be overly concerned with performance.  Just turn down the logging to run fast.  Turn up the logging to debug.  Putting if-enabled wrappers on every log call kind of defeats the purpose of having log levels in the first place.  Please let me know if I'm mistaken but I don't recall the log calls taking any considerable CPU when the level is set correctly.  If I'm wrong, please open an issue and attach visualvm data (screenshot or export) that shows high consumption when NOT at trace level logging.  Or, if you like, have a look at #103 and see if that could use some attention.  It looks stagnant..?

codehustler: it was with trace disabled. its a very common pattern to surround those low level log statements because its faster to check than to build the strings every call. keep in mind that the message which is logged is build regardless if logging is enabled or no.

team172011:
@Mulder8472 

> it was with trace disabled. its a very common pattern to surround those low level log statements because its faster to check than to build the strings every call. keep in mind that the message which is logged is build regardless if logging is enabled or no

Could it be that the specific call `getClass().getSimpleName()` in `AbstractRule` is responsible for the performance loss and not the traceXYZ call in general? Because the [String will not be build](https://www.slf4j.org/faq.html#logging_performance) if logging is disabled if it is a parameterized message format like we use.

ta4j-core/src/main/java/org/ta4j/core/trading/rules/AbstractRule.java
```diff
@@ -43,6 +43,8 @@ this software and associated documentation files (the "Software"), to deal in
      * @param isSatisfied true if the rule is satisfied, false otherwise
      */
     protected void traceIsSatisfied(int index, boolean isSatisfied) {
-        log.trace("{}#isSatisfied({}): {}", className, index, isSatisfied);
+        if ( log.isTraceEnabled() ) {
+            log.trace("{}#isSatisfied({}): {}", getClass().getSimpleName(), index, isSatisfied);
```
team172011, line 47: should be `className` instead of `getClass().getSimpleName()`. Then we should not need an additional `isTraceEnabled()` check

edlins:
> We should also update the dependency to 1.7.5 or higher because of performance improvements in slf4j

I'm using slf4j 1.7.25 with logback-classic 1.2.3 with no problems.

Please read https://www.slf4j.org/faq.html#logging_performance to understand why in slf4j parametrized messages are better than if-wrappers.  In parametrized messages the string is not even built until after it is determined whether to log or not.  However, if the parametrized message contains a method call like getSimpleName() then that is evaluated prior to the determination of whether or not to log.  All logger calls should use parametrized messages as documented at that FAQ link (ie **no method calls, no string concatenations, no formatters**) and then logger performance will be maximized.  Do we have a volunteer to check all logger calls for proper parametrized messages?  A quick search found 16 "log." strings in core.

edlins:
> We should also update the dependency to 1.7.5 or higher because of performance improvements in slf4j

I'm using slf4j 1.7.25 with logback-classic 1.2.3 with no problems.  And I just confirmed that https://github.com/ta4j/ta4j/blob/master/ta4j-core/pom.xml uses 1.7.25.

Please read https://www.slf4j.org/faq.html#logging_performance to understand why in slf4j parametrized messages are "better" than if-wrappers.  In parametrized messages the string is not even built until after it is determined whether to log or not.  However, if the parametrized message contains a method call like getSimpleName() then that is evaluated prior to the determination of whether or not to log.  If all logger calls use parametrized messages as documented at that FAQ link (ie **no method calls, no string concatenations, no formatters**) then logger performance will be maximized when both logging and not logging.

This case is an exception because a method call, string concatenation, formatter, etc *must* be used to generate the parametrized message.  By putting that call in the field initializer we reduced the frequency of calls to getSimpleName() *and* corrected the logger call to use a proper parametrized message.  Once using the className field to make a *proper parametrized message* **there is no advantage to adding an if-wrapper** because the string is not built if trace is disabled.  And when trace is enabled, it's still a proper parametrized message so it follows the best practice at the FAQ.

Previously, when we did *not* have a proper parametrized message, wrapping the call with an if-test would have yielded an improvement.  But now it would be a slight decrease in performance due to an unnecessary if-test to prevent a string build that would not happen anyway when trace is disabled.  And when trace is enabled the if-test still takes overhead and doesn't prevent anything.

I confirmed this by adding the if-wrapper and with trace disabled **runtime increased by 5.3%** compared to exactly the same run without the if-wrapper.  With trace enabled **runtime increased by 3.7%**.  Try it all four ways for yourself (if-wrapper + trace, no if-wrapper + trace, if-wrapper + no trace, no if-wrapper + no trace.)

I think the if-wrappers might have value if the logging requires some processing that must happen at logger call time, and you don't want it to happen when logging is disabled.  But this is not the case here since that processing was (re)moved.

team172011:
This improvement will not be merged, because of the reasons clarified in my and @edlins comments. I would make a new PR that uses parameterized message format in all cases

PR #158, opened 2018-01-22
Leave last bar uncached

### LINKED ISSUE #149: Question: Handle partial ticks on TimeSeries
Hi!

Is there a way to handle partial ticks on TimeSeries? Such as keep changing over and over the one at the end index, and switch to the next one just when it is the final for the timeframe.

I know that there are other possible ways to do it, but for example, I want to run indicators (e.g. stop-loss) before receiving the final tick for the 30 min.
One way that I think is to create a new TimeSeries every time, but I'm sure this wouldn't have a good performance and probably can't leverage the cached indicators.

Also, if you are okay with it and don't see any problems, I can submit a PR adding a new `addTick` method with a `partial` flag.

Thank you!

mdesmet:
Fixes #149 

- Leaves last bar uncached in order to have real time indicators, allowing real time data sources to add trades to the last bar and retrieve updated indicators.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/CachedIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/CachedIndicator.java
@@ -97,7 +97,10 @@ public abstract class CachedIndicator<T> extends AbstractIndicator<T> {
                 // Result not calculated yet
                 highestResultIndex = index;
                 result = calculate(index);
-                results.set(results.size()-1, result);
+                if(index != series.getEndIndex()) {
+                    // Only cache results before last bar
+                    results.set(results.size()-1, result);
+                }
             } else {
                 // Result covered by current cache
                 int resultInnerIndex = results.size() - 1 - (highestResultIndex - index);
--- a/ta4j-core/src/test/java/org/ta4j/core/indicators/CachedIndicatorTest.java
+++ b/ta4j-core/src/test/java/org/ta4j/core/indicators/CachedIndicatorTest.java
@@ -158,4 +158,22 @@ public class CachedIndicatorTest {
             fail(t.getMessage());
         }
     }
+
+    @Test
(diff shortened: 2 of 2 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/CachedIndicator.java
```diff
@@ -97,7 +97,10 @@ public T getValue(int index) {
                 // Result not calculated yet
                 highestResultIndex = index;
                 result = calculate(index);
-                results.set(results.size()-1, result);
+                if(index != series.getEndIndex()) {
+                    // Only cache results before last bar
+                    results.set(results.size()-1, result);
+                }
```
team172011, line 103: I think this does not work because the following is possible:
**A call of the `getValue(int index)` function with**
- cache= {1,2,3}
- `highestResultIndex `= 2
- `index `= 3 (last bar of series)

1. `increaseLengthTo(index, maximumResultCount)`;
    - cache = {1,2,3,null}
2. `if (index > highestResultIndex) `-> true
    - `highestResultIndex `= 3
    - cache =  {1,2,3,null} (stays the same because index == series.getEndIndex())

**Next call of the `getValue(int index)` function with:**
- `index `= 3 (again)
- cache = {1,2,3,null} 
- `highestResultIndex `= 3
1. `increaseLengthTo(index, maximumResultCount)`;
    - cache = {1, 2, 3, null} (nothing to do)
2. `if (index > highestResultIndex) `-> false!
3. `else...`
    - `resultInnderIndex` = 4-1-0 = 3
    - `result` = null
    - `if (result == null)` -> true
        - `results.set(3, result);` **_## -> last index of indicator is in cache!!!_**


#### Maybe it works if we change `if (index > highestResultIndex)` to `if (index >= highestResultIndex)` ?

PR #170, opened 2018-01-29
Indicator window rules

edlins:
Fixes #.

- Add *IndicatorWindowRules


edlins:
Just another one of my personal items, transparent addition.

team172011:
I think this Rule will be mostly redundant with `InSlope ` and `InPipe ` rule

edlins:
Hmm, I don't see how those rules allow you to say "satisfied if 75% of the previous 20 indicator values (any 15 out of the 20) are greater than 65", but I'll just keep these weird rules to myself.  I may submit a few more PRs as I'm syncing to 0.11 but don't feel obligated to take them unless you want.  Just figured I'd offer them as I work through the diffs.

team172011:
You could consider to modify the existing rules if you think this is a necessary feature

Issue #175, opened 2018-02-09
Divide by Zero in Indicators and XLS


edlins:
In order to do an XLS test of StochasticRSIIndicator I applied RSI-type logic to the indicator and to the spreadsheet so that it never returns NaN.  Is this correct, or is it better to check isNaN() in ta4j against "#DIV/0!" in Excel?  If we do this then TaTestsUtils#assertIndicatorEquals() will have to be verified to handle NaNs, and XlsTestsUtils#getValues() will have to be fixed to handle "#DIV/0!" as NaN.  (Since I'm still using my primitive double implementation of Decimal I actually used nulls for NaNs so I had to fix NPE in assertIndicatorEquals.)

edlins:
I think I'll go with isNan() and "#DIV/0!" instead of nulls.  I've never seen specific guidance on divide by zero on Stochastic RSI.  It seems that only Wilder's Relative Strength provides the "100 or zero" special cases.

team172011:
merged!

Issue #147, opened 2018-01-01
Construct StochasticRSI indicator properly


FatihErdem:
Hello,

I am trying to build Stochastic RSI indicator. In StochasticRSI there are K, D, Rsi and Stochastic parameters.

I want build StochRSI(14, 9, 3, 3):

- RSI = 14
- Stoch = 9
- K = 3
- D = 3  

```
StochasticOscillatorKIndicator k = new StochasticOscillatorKIndicator(series, 3);
StochasticOscillatorDIndicator d = new StochasticOscillatorDIndicator(k);
RSIIndicator rsi = new RSIIndicator(d, 14); // I think it's wrong
StochasticRSIIndicator stochRsi = new StochasticRSIIndicator(rsi, 9);
```

Result of the stochRsi calculation is totally wrong when compare to result with exchange's graphs. 
How can I construct StochasticRSIIndicator properly?

edlins:
There are many different formulas for RSI in common use.  ta4j uses a modified moving average to calculate average gain and average loss.  Then a simple ratio of the mma's is used for RS (which is then converted to RSI.)  Take a look at https://github.com/edlins/ta4j/blob/b511e18ead627470f64b13b0b5e71cbd37063e57/ta4j-core/src/test/java/org/ta4j/core/indicators/RSIIndicatorTest.java (unreleased code).  The last test case refers to a website that uses a different formula.  That site uses a simple moving average to calculate average gain and average loss.  Then a modified moving average of the RS ratios to calculate RSI.  Do you know exactly what formula your exchange is using?

FatihErdem:
Hello @edlins , 

Is there any multiple Stochastic RSI variation? I thought StochasticRSI is the one of the RSI variation but is standard indicator. I am trying to build this version. 
https://www.tradingview.com/wiki/Stochastic_RSI_(STOCH_RSI)#CALCULATION

edlins:
Stoch RSI is just stochastics on the RSI.  AFAIK the stochastics formula does not have multiple variations like RSI does.  But Stoch on Cutler's RSI will be different than Stoch on Wilder's RSI, so the RSI variation matters.  If you're trying to match other people's results, you'll still need to match their RSI variation.  https://www.tradingview.com/wiki/Relative_Strength_Index_(RSI) shows:
```
RS = Average Gain / Average Loss
```
Which I take to be Cutler's RSI.  https://en.wikipedia.org/wiki/Relative_strength_index#Cutler's_RSI
[...]

PR #178, opened 2018-02-14
Enhance/xlsdailydata

### LINKED ISSUE #176: Daily Data in XLS
I've modified XlsTestsUtils#getSeries() to handle daily (and monthly, yearly, etc) data automatically.  It had been coded to only handle weekly data.  Should this go on the develop branch?  Are you rebasing num to develop whenever there's a merge to develop?

edlins:
Fixes #176 #175 .

- Handle daily, weekly, monthly, etc data in XLS
- Handle #DIV/0! in XLS
- 


edlins:
I have Organize Imports on automatic and it changes around the imports on Save.  Let me know if I should turn that off.
This also helps address #147 but does not fully resolve it.
I think I got all of the Num changes applied so this should be ready to merge.

ta4j-core/src/test/java/org/ta4j/core/XlsTestsUtils.java
```diff
@@ -46,11 +57,16 @@ this software and associated documentation files (the "Software"), to deal in
...
-     * @throws IOException if the workbook constructor or close throws
-     *             IOException
+     * @throws IOException if inputStream returned by
+     *             getResourceAsStream is null or if HSSFWorkBook constructor
+     *             throws IOException or if close throws IOException
      */
     private static Sheet getSheet(Class<?> clazz, String fileName) throws IOException {
-        HSSFWorkbook workbook = new HSSFWorkbook(clazz.getResourceAsStream(fileName));
+        InputStream inputStream = clazz.getResourceAsStream(fileName);
+        if (inputStream == null) {
```
team172011, line 66: `null` check is not needed, because exactly the same will be done next line in the constructor of [HSSFWorkbook](https://poi.apache.org/apidocs/org/apache/poi/hssf/usermodel/HSSFWorkbook.html#HSSFWorkbook-java.io.InputStream-)

team172011:
Could we use `assertNumEquals ` instead of `assertEquals `in the `public static void assertIndicatorEquals(Indicator<Num> expected, Indicator<Num> actual)`?

ta4j-core/src/test/java/org/ta4j/core/XlsTestsUtils.java
edlins: Actually, the workbook constructor does not perform a null check.  If inputStream is null, it throws an NPE which isn't very helpful to the developer.  It only throws an IOException if the stream is non-null but cannot be read.  Try renaming one of the XLS files without this check and you'll see an NPE.  Then add the check and you get a nice message stating that the file is not found.

edlins:
```
    public HSSFWorkbook(InputStream s, boolean preserveNodes)
            throws IOException
    {
        this(new NPOIFSFileSystem(s).getRoot(), preserveNodes);
    }
```

edlins:
> Could we use assertNumEquals

Sure.  We should also
- change `assertValuesEquals()` to use `assertNumEquals()`
- rename `assertDecimalNotEquals()` to `assertNumNotEquals()` and fix calls
- fix permuted parameters and remove swapping TODOs for permuted parameters

I'll fix `assertIndicatorEquals()` and `assertDecimalNotEquals()` here but the others will probably require their own PR because they will likely touch many other files.

ta4j-core/src/test/java/org/ta4j/core/XlsTestsUtils.java
team172011: Okay, just thought that this check was enough
```java
public static ReadableByteChannel newChannel(final InputStream in) {
        checkNotNull(in, "in");
       (...)
}
```
but the check is not reached

PR #181, opened 2018-02-14
Leave last bar uncached

### LINKED ISSUE #149: Question: Handle partial ticks on TimeSeries

mdesmet:
Fixes #149

Leaves last bar uncached in order to have real time indicators, allowing real time data sources to add trades to the last bar and retrieve updated indicators.

### DIFF
--- a/ta4j-core/src/test/java/org/ta4j/core/indicators/CachedIndicatorTest.java
+++ b/ta4j-core/src/test/java/org/ta4j/core/indicators/CachedIndicatorTest.java
@@ -163,4 +163,26 @@ public class CachedIndicatorTest extends AbstractIndicatorTest<Indicator<Num>,Nu
             fail(t.getMessage());
         }
     }
+
+    @Test
+    public void leaveLastBarUncached() {
+        TimeSeries timeSeries = new MockTimeSeries(numFunction);
+        ClosePriceIndicator closePrice = new ClosePriceIndicator(timeSeries);
+        assertNumEquals(5000, closePrice.getValue(timeSeries.getEndIndex()));
+        timeSeries.getLastBar().addTrade(numOf(10), numOf(5));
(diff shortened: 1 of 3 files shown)


ta4j-core/src/test/java/org/ta4j/core/indicators/CachedIndicatorTest.java
```diff
@@ -163,4 +163,26 @@ public void recursiveCachedIndicatorOnMovingTimeSeriesShouldNotCauseStackOverflo
             fail(t.getMessage());
         }
     }
+
+    @Test
+    public void leaveLastBarUncached() {
+        TimeSeries timeSeries = new MockTimeSeries(numFunction);
+        ClosePriceIndicator closePrice = new ClosePriceIndicator(timeSeries);
+        assertEquals(5000, closePrice.getValue(timeSeries.getEndIndex()).doubleValue(),0.001);
```
team172011: Please use` TestUtils.assertNumEquals(expected, actual)` to allow comparison of current `Num` implementations and to avoid double transformation

team172011:
thanks. Corresponding functions to `TimeSeries ` interface will be added soon. (To support adding price to the last bar and setting the last bar)

PR #183, opened 2018-02-15
TestUtils changes

edlins:
Fixes #none.

- TestUtils changes to assertNumEquals methods
- ~~Remove assertNumEquals(Num, Num), use assertEquals(Object, Object)~~
- Remove assertNumEquals(int, Num), calls implicitly revert to assertNumEquals(double, Num)
- Add assertNumNotEquals(), assertIndicatorNotEquals() methods
- Add TestUtilsTest for testing assertNum(Not)Equals(), assertIndicator(Not)Equals() methods
- Change Decimal to Num
- Remove assertValuesEquals(), not used


edlins:
I'm sure there will be some discussion regarding this PR.  Look it over and let me know what you think.  I wanted to clean up TestUtils and at the same time decided that all assertNumEquals() methods should call assertEquals(double, double, double).  Also decided assertNumEquals(Num, Num) is not needed because JUnit has assertEquals(Object, Object) **unless** you want to compare BigDecimalNum against DoubleNum.  If you do want that, can easily add it back in, convert the Nums to doubles and call assertEquals(double, double, double).

ta4j-core/src/test/java/org/ta4j/core/TestUtils.java
```diff
@@ -22,96 +22,93 @@ this software and associated documentation files (the "Software"), to deal in
...
-     * @param expected the given {@code String} representation to compare the actual value to
-     * @throws AssertionError if the actual value is not equal to the given {@code String} representation
+     * @param actual the actual {@code Num} value
+     * @param expected the given {@code String} representation to compare the
+     *            actual value to
+     * @throws AssertionError if the actual value is not equal to the given
+     *             {@code String} representation
      */
     public static void assertNumEquals(String expected, Num actual) {
-        assertEquals(expected, actual.numOf(new BigDecimal(expected)).toString());
```
edlins: Hrm, expected = expected?

ta4j-core/src/test/java/org/ta4j/core/TestUtils.java
```diff
@@ -22,96 +22,93 @@ this software and associated documentation files (the "Software"), to deal in
...
-     * @param expected the given {@code Integer} representation to compare the actual value to
-     * @throws AssertionError if the actual value is not equal to the given {@code Integer} representation
+     * @param actual the actual {@code Num} value
+     * @param expected the given {@code int} representation to compare the
+     *            actual value to
+     * @throws AssertionError if the actual value is not equal to the given
+     *             {@code Integer} representation
      */
     public static void assertNumEquals(int expected, Num actual) {
-        assertEquals(expected, actual.intValue());
```
edlins: `assertEquals(3, numOf(3.4))` passes?
Better `assertEquals(3d, numOf(3.4).doubleValue(), GENERAL_OFFSET)` should fail?

ta4j-core/src/test/java/org/ta4j/core/TestUtils.java
```diff
@@ -22,96 +22,93 @@ this software and associated documentation files (the "Software"), to deal in
...
-     * @throws AssertionError if the actual value is not equal to the given {@code String} representation
+     * @param actual the actual {@code Num} value
+     * @param expected the given {@code String} representation to compare the
+     *            actual value to
+     * @throws AssertionError if the actual value is not equal to the given
+     *             {@code String} representation
      */
     public static void assertNumEquals(String expected, Num actual) {
-        assertEquals(expected, actual.numOf(new BigDecimal(expected)).toString());
+        assertEquals(Double.valueOf(expected), actual.doubleValue(), GENERAL_OFFSET);
```
team172011: The expected value could have higher precision as `Double ` type provides. Should be
`assertEquals(actual.numOf(new BigDecimal(expected)), actual);`

ta4j-core/src/test/java/org/ta4j/core/TestUtils.java
team172011: Maybe we should not provide this kind of assert?

edlins: Okay.  Will remove `assertNumEquals(int, Num)` and `assertNumNotEquals(int, Num)`.  Any existing calls will implicitly cast int to double and call `assertNumEquals(double, Num)` which will call `assertEquals(double, double, double)`.  `BarTest` has these types of calls as an example.

ta4j-core/src/test/java/org/ta4j/core/TestUtils.java
edlins: Okay, will correct.

edlins:
> provide assertNumEquals(Num, Num)

Okay, adding back in.  Of course this usage that calls `assertEquals(Object, Object)` which does not allow for any precision offset.  These used to pass:
```
assertNumEquals("3.1415", numOf(3.14159265358979));
assertNumEquals("3.14159265358979", numOf(3.1415));
```
But these now will fail and should be changed to `assertNumNotEquals()`.  If you want those to pass then `expected.doubleValue()` should be used to call `assertNumEquals(double, Num)` which calls `assertEquals(double, double, double)`.

edlins:
Take a close look at `TestUtilsTest` and make sure that works the way you expect.  Also consider carefully my note about `assertIndicatorEquals()` not using exact matches (test case forthcoming).

edlins:
I have no more TODOs here.  Pending any suggested changes this may be done.  Next step will be to fix all of the unit tests to only use TestUtils and not use JUnit directly.

edlins:
Hold on this, please.  I'm making more significant changes.

edlins:
Okay, lots of changes here.
- ExternalIndicatorTest: allow passing of explicit numFunction
- TestUtils: cast all Nums to BigDecimalNums for high precision comparisons, some methods take explicit deltas
- XLSTestUtils renamed, better cell reading
- BigDecimalDoubleCompareTest: added, compares precision of Indicator<BigDecimalNum> and Indicator<DoubleNum>
- RSIIndicatorTest: use 64-digit BigDecimalNum result Strings for expected, use TestUtils instead of JUnit Asserts
- SMAIndicatorTest: use 64-digit BigDecimalNum result Strings for expected, use TestUtils instead of JUnit Asserts
- XLSIndicatorTest: allow passing of explicit numFunction

edlins:
So, it is possible to dramatically increase the precision of the unit tests.  After updating `RSI` and `SMA` tests, they now pass `BigDecimalNum` and `DoubleNum` up to `0.000000000001`.  Beyond that, `DoubleNum` fails for those two `Indicators`.  `BigDecimalNum` works up to `0.00000000000000000000000000001`.  You can enable those different deltas by uncommenting them in `TestUtils`.  Right now it's set to the old `0.0001` so the non-enhanced unit tests will still pass.

To do this, in `TestUtils` I convert all incoming `Nums` to `BigDecimalNums` before comparing.  Of course, this explicit conversion causes the failure in `testWrongNumType` since there is no longer an `Exception` thrown when comparing different `Num` types.  I think that test should be removed.

I also added `BigDecimalDoubleCompareTest` which compares `Indicators` of different `Num` types using XLS series data.  You can see exactly how much precision is lost in `DoubleNum` for a given XLS `TimeSeries` for specific `Indicators`.

The method that was being used to read XLS cells was not ideal because it was getting values like `126.100000000003` when the cell really contained `126.1`.  The new method looses one digit of precision but is more accurate.  Uncomment the commented code to see the differences.

I eagerly await feedback.

team172011:
> BigDecimalNum works up to 0.00000000000000000000000000001
I am happy to see first approach for precision validation of `BigDecimal ` Nums

So far I have not found time to review all your changes, but does 
> TestUtils: cast all Nums to BigDecimalNums for high precision comparisons, some methods take explicit deltas
means that there are no calls of the `DoubleNum#equals(other)` function in DoubleNum tests?

edlins:
> means that there are no calls of the `DoubleNum#equals(other)` function in `DoubleNum` tests?

Correct.  In fact, there are no `DoubleNum` tests in `TestUtils` now.  Tests of that sort should live in `DoubleNumTest` for testing the `DoubleNum` class.  AFAIK `TestUtils` is only for testing `Indicators` and `Criteria`, not `core` classes.  After some changes I just made I now have:
```
public static void assertNumEquals(String expected, Num actual) {
    assertBigDecimalNumEquals(null, 
        BigDecimalNum.valueOf(expected, HIGH_PRECISION), 
        BigDecimalNum.valueOf(actual.toString()), 
        BIGDECIMALNUM_OFFSET);
}
```
and supporting the `HIGH_PRECISION` parameter in `BigDecimalNum`:
```
private BigDecimalNum(String val, int precision) {
    delegate = new BigDecimal(val, new MathContext(precision, RoundingMode.HALF_UP));
}
public static Num valueOf(String val, int precision) {
    return val.toUpperCase().equals("NAN") ? NaN : new BigDecimalNum(val, precision);
}
```
`String expected` is converted with `BigDecimalNum.valueOf(expected, HIGH_PRECISION)` (currently precision 64 in `TestUtils`).  `Num actual` is converted with `BigDecimalNum.valueOf(actual)` (default precision 32 in `BigDecimalNum`).  `BIGDECIMALNUM_OFFSET` is also `HIGH_PRECISION`.  They are then passed to the `private` method `assertBigDecimalNumEquals(String, Num, Num, Num)`.  

I implemented `assertBigDecimalNumEquals()` explicitly without using JUnit Assert:
```
private static void assertBigDecimalNumEquals(String message, Num expected, Num actual, Num delta) {
    if (expected.minus(actual).abs().isGreaterThan(delta)) {
        if (message == null) message = "value " + actual.toString() + " does not match expected " + expected.toString();
        throw new AssertionError(message);
    }
}
```

I used `BigDecimalNum` set to 64 digit precision to calculate the high precision expected values in the enhanced `Indicator` tests.  I then had to pass those expected values as `Strings` because they are much higher precision than `double`.  I just re-ran the `SMA` tests and for those unit tests these are where they fail:
```
0.000000000001 all tests pass
0.0000000000001 `externalData()` fails (XLS precision)
0.0000000000000001 `usingTimeFrame3UsingClosePrice()` fails in Test Case 1 (`DoubleNum`)
0.00000000000000000000000000000001 `usingTimeFrame3UsingClosePrice()` fails in Test Case 2 (`BigDecimalNum`)
```
By using precision 64 `BigDecimalNums` `Strings` for expected values we can accurately test `BigDecimalNum` default precisions up to 64.  If you want to go beyond that, we'll have to calculate higher precision expected value `Strings` and increase `HIGH_PRECISION` accordingly.  But max precision 64 should satisfy most users.

edlins:
Did some performance testing on 200-day `SMAIndicator.calculate()` for all indices of daily data for the S&P 500 from 1/1/1970 to 1/1/2018 for `BigDecimalNum` vs `DoubleNum`.

19.2% ratio of total runtime of `DoubleNum` vs `BigDecimalNum` (5.2x faster)
5.1% ratio of `Num` methods runtimes of `DoubleNum` vs `BigDecimalNum` (19.6x faster)

Also found that `Objects.requireNonNull()` in `BigDecimalNum.<init>(BigDecimal)` consumed 9.8% of the non-`Num` methods runtimes and 1.6% of all methods runtimes for `BigDecimalNum`.  I believe that call may be removed for a small performance boost.

edlins:
Out of curiosity, I implemented `IntNum` and the test took 94% of the time of `DoubleNum`, so hardly any performance improvement.  And the precision in my test case dropped to `1.5`.  `DoubleNum` primitive `double` operations are probably as fast as we can get with `Num`.

ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
```diff
@@ -187,8 +185,7 @@ public boolean equals(Object obj) {
         if(obj == NaN){
             return false;
         }
-        DoubleNum doubleNumObj = (DoubleNum) obj;
-        return Math.abs(delegate - doubleNumObj.delegate) < EPS;
+        return Double.compare(delegate, ((DoubleNum) obj).delegate) == 0 ? true : false;
```
edlins: I'm concerned about having that `EPS` precision hidden away in `DoubleNum`.  I changed it and it did not impact any core unit tests.

team172011: This is not the correct way to compare approximations (https://stackoverflow.com/questions/8081827/how-to-compare-two-double-values-in-java). For instance this will fail:
```
    @Test
    public void simpleEqualsTest(){
        Num a = numOf(1.000001);
        Num b = numOf(0.000001);
        assertNumEquals(numOf(1.0), a.minus(b));
    }

java.lang.AssertionError: 
Expected :1.0
Actual   :0.9999999999999999
 <Click to see difference>
```

By the way: you should skip `? true : false;` check

ta4j-core/src/test/java/org/ta4j/core/indicators/BigDecimalDoubleCompareTest.java
```diff
@@ -10,46 +9,60 @@
...
+import static org.ta4j.core.TestUtils.assertIndicatorNotEquals;
 
 public class BigDecimalDoubleCompareTest {
 
     ExternalIndicatorTest xls;
-    Indicator<Num> dnClose;
-    Indicator<Num> bdnClose;
+    Indicator<Num> doubleClose;
+    Indicator<Num> bigDecimalClose;
     protected Logger log;
```
team172011: no need for logging in tests

ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
```diff
@@ -24,6 +24,8 @@ this software and associated documentation files (the "Software"), to deal in
 
 import java.util.function.Function;
 
+import org.ta4j.core.TestUtils;
```
team172011: TestUtlis in main package? Should be in test

ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
```diff
@@ -36,8 +38,9 @@ this software and associated documentation files (the "Software"), to deal in
 
     private final double delegate;
 
-    private final static double EPS = 0.00001; // precision
-
+    //private static final double EPS = 0.000000000000001;
+    private static final double EPS = TestUtils.GENERAL_OFFSET;
```
team172011: I would prefer to swith dependency ( TestUtils.GENERAL_OFFSET = DoubleNum.EPS). The main package (core) should be able to run without test package

ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
team172011: Should not be possible to use imports of test classes in the core package.

edlins: Looks like everything in test is in package org.ta4j.core or a sub-package of that.  If that changes to org.ta4j.test (or org.ta4j.test.core) then that's probably a separate PR.

ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
edlins: Okay.  Update test cases to use DoubleNum.EPS instead of TestUtils.GENERAL_OFFSET also?

team172011:
> Also consider moving to an ULPs-based technique rather than absolute delta:

Can we discus in Release 0.13? Why using DoubleNum if one needs such precision

ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
team172011: I think there must be something wrong in your branch? 
grafik

Your IDE should not be able to find TestUtils from core package:
grafik

team172011: You will not be able to compile. Travis tests failed because of this import:

``` 
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:compile (default-compile) on project ta4j-core: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/ta4j/ta4j/ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java:[27,21] cannot find symbol
[ERROR]   symbol:   class TestUtils
[ERROR]   location: package org.ta4j.core
[ERROR] /home/travis/build/ta4j/ta4j/ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java:[42,39] cannot find symbol
[ERROR]   symbol:   variable TestUtils
[ERROR]   location: class org.ta4j.core.num.DoubleNum
```

edlins:
There seems to be an inconsistency between `DoubleNum` and `BigDecimalNum`.  `DoubleNum.equals()` is (once again) a delta comparison but `BigDecimalNum.equals()` is an exact comparison using `BigDecimal.compareTo()`.  Shouldn't `BigDecimalNum.equals()` be a delta comparison also, but with it's own higher-precision `EPS` (like you suggested at one point)?  `BigDecimal` has limited precision (from `MathContext`) just as `double` does.  It can just have a much higher precision.  I don't see why one should be exact and the other be delta.

ta4j-core/src/test/java/org/ta4j/core/TestUtils.java
```diff
@@ -22,96 +22,214 @@ this software and associated documentation files (the "Software"), to deal in
...
     /** Offset for double equality checking */
-    public static final double GENERAL_OFFSET = 0.0001;
+    // deprecated so that JUnit Assert calls direct from unit tests will show warnings
+    // TODO: modify unit tests to not call JUnit Assert but to call TestUtils Assert
+    @Deprecated
+    public static final double GENERAL_OFFSET = DoubleNum.valueOf(DoubleNum.EPS).doubleValue();
+    private static Num HIGH_PRECISION_OFFSET = highPrecisionNumFunc.apply("0.0001");
+
+    public static void setPrecision(String precision) {
+        HIGH_PRECISION_OFFSET = highPrecisionNumFunc.apply(precision);
```
edlins: This was lazy.  Should not be a static method to update a static field.  Should probably just add signatures for methods that accept an explicit ~precision~ delta parameter.

team172011:
> There seems to be an inconsistency between DoubleNum and BigDecimalNum. DoubleNum.equals() is (once again) a delta comparison but BigDecimalNum.equals() is an exact comparison using BigDecimal.compareTo()

The inconsistency comes with the delegates. `BigDecimal ` can represent decimal values  **exact** but the double may be an approximation. `equals ` and `compareTo ` functions had been overwritten because they are not related to the (this)class but the decimal value represented by delegate. So `DoubleNum `needs a delta, but `BigDecimalNum `not.  https://ta4j.github.io/ta4j-wiki/Num.html

edlins:
```
String string = "1.34805205093455024967890329029384708293458243759872430958720985750984325897234095782309482057";
BigDecimal A = new BigDecimal(string, new MathContext(16, RoundingMode.HALF_UP));
BigDecimal B = new BigDecimal(string, new MathContext(32, RoundingMode.HALF_UP));
assertEquals(A, B);  // fails
```
BigDecimalNum just defaults the precision to 32 so it rounds off at 32 digits.  It's not unlimited, just arbitrary.  Why would you have a RoundingMode if it were unlimited precision?

edlins:
> Why creating a new object??

So you can `BigDecimalNum.equals(DoubleNum)` without a `ClassCastException`.  Review `NumTest` for examples.

edlins:
I'm also considering switching entirely to `int precision` from `Num delta` to avoid [Catastrophic Cancellation](https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/) in `expected.minus(actual)`.  It's super easy to check `BigDecimalNum.matches(Num other, int precision)` with my custom `BigDecimalNum.valueOf(String val, int precision)`:
```
    public boolean matches(Num other, int precision) {
        Num otherNum = BigDecimalNum.valueOf(other.toString(), precision);
        Num thisNum = BigDecimalNum.valueOf(this.toString(), precision);
        if (thisNum.toString().equals(otherNum.toString())) {
            return true;
        }
        return false;
    }
```
If you want to take some of this PR and leave others for now, I can break it up but I'd rather just keep plugging away and get it right rather than come back to it again.

team172011:
> So you can BigDecimalNum.equals(DoubleNum) without a ClassCastException

The `ClassCastException `is great! This way we do not need an explicit check if `other` is type of `BigDecimalNum ` (`DoubleNum`) in every operation and if the cast fails the user will get a meaningfull exception like "Can not cast DoubleNum to BigDecimalNum"

edlins:
Will reopen in the future.

PR #190, opened 2018-02-24
Replace 'timeFrame' parameter with 'barCount'

abzif:
Replace 'timeFrame' parameter with 'barCount'
An issue for that change existed some time ago, however it was closed in the meantime.


abzif:
added entry in CHANGELOG.md

team172011:
Okay, will be merged after #183 to avoid conflicts

PR #195, opened 2018-02-26
Always package sources

shalupov:
This helps https://jitpack.io to have browsable sources


team172011:
@shalupov thanks, I have deployed 0.12-SNAPSHOT, can you confirm browsable sources in jitpack know?

PR #198, opened 2018-03-01
Enhance/precision num

edlins:
Fixes #.

- Introduce PrecisionNum with arbitrary precision
- Add methods for generating Nums from Strings with arbitrary precision
- Add matches() for inexact matching

- [x] added an entry to the unreleased section of ~~`CHANGES.md`~~ `CHANGELOG.md` (fix template)


edlins:
- The second commit only turns on logback-classic for `PrecisionNum` to see how `matches()` works.  Cherry pick [the first commit](https://github.com/ta4j/ta4j/pull/198/commits/3018e24a34d1e1fdc2a1c26502fe47b0bd669127) to leave `pom.xml` and `logback.xml` out.
- **matches() belongs in Num**
- PrecisionNum would have been easier (and smaller) by extending BigDecimalNum but that's final with private constructors

edlins:
Also, I would only use `PrecisionNum` for unit tests because of the increased memory consumption of storing a `MathContext` as an instance field.  `BigDecimalNum` cannot be precision-tested without much higher precision (>>32, like 64 or 128) expected values.

ta4j-core/pom.xml
```diff
@@ -18,6 +18,12 @@
             <artifactId>slf4j-api</artifactId>
             <version>1.7.25</version>
         </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>1.2.3</version>
+            <scope>test</scope>
+		</dependency>
```
team172011: library should not have a logger implementation as dependency. This should be up to the user

edlins: > library should not have a logger implementation as dependency. This should be up to the user

[first commit only](https://github.com/ta4j/ta4j/pull/198#issuecomment-369730379)



> Is it not possible to add this functionality to the existing BigDecimalNum instead of creating a new implementation with the same delegate?

[I'm concerned about the heap overhead](https://github.com/ta4j/ta4j/pull/198#issuecomment-370167986) but I have not made a quantitative assessment yet.  

EDIT: If the delegate had a getMathContext() getter then the MathContext would not have to be an instance field.  But there is no such getter.

team172011: > first commit only

I can only merge your PR and not pick a special commit...

> I'm concerned about the heap overhead

Would be nice to check how big the additional memory consumption is. If it is to big and we have to add this as new Num implementation, it is worth to change the existing `BigDecimalNum` using [Decimal4j](https://github.com/tools4j/decimal4j). Than we would have
 - **DoubleNum** -> fast floating point arithmethic (for those who want to have very high performance)
- **DecimalNum** -> fixed-point arithmetic using Decimal4j (for those who want to have fixed-point arithmetic but do not need high precision, like for stocks)
- **PrecisionNum** ->  fixed-point arithmetic using BigDecimal (for those who want to have fixed-point arithmetic and very high precision)

edlins:
Okay, figured out that I had to 'force' on my push to get rid of logging.  I thought you could cherry-pick the commits but I guess not.

Unfortunately, the comment thread disappeared when logging was removed. This was the latest:
> 
>
>    I'm concerned about the heap overhead
>
>Would be nice to check how big the additional memory consumption is. If it is to big and we have to add this as new Num implementation, it is worth to change the existing BigDecimalNum using Decimal4j. Than we would have
>
>    DoubleNum -> fast floating point arithmethic (for those who want to have very high performance)
>    DecimalNum -> fixed-point arithmetic using Decimal4j (for those who want to have fixed-point arithmetic but do not need high precision, like for stocks)
>    PrecisionNum -> fixed-point arithmetic using BigDecimal (for those who want to have fixed-point arithmetic and very high precision)

I can do some memory testing with my large S&P500 XLS data on the various types, probably tomorrow.

I don't see the need to change BigDecimalNum but I guess I don't really care.  If you want to, fine, but 32 digit precision BigDecimalNum seems to be working fine for our purposes.

I only need PrecisionNum for testing the accuracy of BigDecimalNum.  I'm fine if it goes in test and never gets released in core.

edlins:
These results do not actually compare high precision Nums because the source data was XLS and BigDecimal automatically down-samples to the precision of the source String so this just shows the difference by adding a MathContext to the instances.

I assume units are in bytes.  PrecisionNum's MathContext appears to take 16 bytes in this usage, and appears to take a total retained size of 64 bytes:

precisionnum

BigDecimalNum total retained size appears to be 48 bytes:

bigdecimalnum

So, a single PrecisionNum appears to be 33% larger than a single BigDecimalNum.

I built TimeSeries for all three types with 12,110 bars and the total consumptions were:
6,837,928 PrecisionNum (20% larger than BigDecimalNum)
5,675,368 BigDecimalNum
3,350,584 DoubleNum (60% of BigDecimalNum)

allnums

edlins:
Just ran a more realistic test with actual high precision Strings and got these sizes
```
92,079,752 PrecisionNum 128 (151% of BigDecimalNum)
68,880,384 PrecisionNum 32 (113% of BigDecimalNum)
60,880,464 BigDecimalNum 32
10,134,496 DoubleNum 16 (17% of BigDecimalNum)
```
Something seems a bit off in the Indicators I used so this may be off.

edlins:
I think you're going to like this.  Most of it's in `PrecisionNum` and `PrecisionNumTest`.  Had to touch `BaseBar` and `TestUtils` to achieve ludicrous precision.

The `BaseBar` change caused other test failures where Mocks are used that never pass in a `numFunc` so I hacked a null check into `BaseBar`.  Might be better to force a `numFunc` in the Mocks.

edlins:
rsis

Given that `PrecisionNum` 32-digit is only 13% larger than `BigDecimalNum`, I'm inclined now to think that `PrecisionNum` can replace `BigDecimalNum` but I'd like to do a CPU comparison next because of the extra String operations.

edlins:
cpus

`PrecisionNum` 32 is only taking about 12% more CPU time than `BigDecimalNum`, consistent with the memory difference as well.  `DoubleNum` only took 31% as much time as `BigDecimalNum`!

edlins:
Outstanding issues:
- `PrecisionNum.matches()` belongs in `Num`: I strongly support this.
- `BaseBar` `null` check vs change constructors to require `numFunc`: Seems cleaner to require `numFunc`.

ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
```diff
@@ -57,6 +57,8 @@ this software and associated documentation files (the "Software"), to deal in
     private Num volume;
     /** Trade count */
     private int trades = 0;
+    /** Num function **/
+    private Function<Number, Num> numFunction;
```
team172011, line 62: I have also thought about this `numFunction ` field in `BaseBar ` and I hesitated for a long time, but I think this solution is cleaner. It is also easier to make the `Num ` check in `TimeSeries ` if adding a new bar

team172011: Like in `TimeSeries` the `Bar `interface should have a `function() ` getter for the `numFunction`. Then we can simplify the `Bar ` check in the `TimeSeries#addBar(Bar b)` implementation of `BaseTimeSeries`

team172011:
@edlins is it possible to replace BigDecimalNum with PrecisionNum within this PR?

edlins:
> is it possible to replace BigDecimalNum with PrecisionNum within this PR?

Sure, I'll work on that today.

MarijnKoesen:
I've got a sqrt implementations ready for bigdecimalnum (not all overflow tests are ready though) which I'll implement on precisionnum when it's merged :+1:

edlins:
Oh boy, really blew up the world with the replacement of BigDecimalNum with PrecisionNum.  There are massive unit test failures that will have to be studied.  Replacement may have to be a separate PR..

edlins:
I see the problem.  Fixing it.

edlins:
Okay, fixed problems with auto-precision but now I have another major problem.

It seems that when high precision `PrecisionNum` is used I'm uncovering rounding errors in the unit tests that exceed the `0.0001` `GENERAL_OFFSET` in `TestUtils`.  Take a look at `EMAIndicatorTest.usingBarCount10UsingClosePrice()`:
```
java.lang.AssertionError: expected:<63.2648> but was:<63.26490307>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:553)
	at org.junit.Assert.assertEquals(Assert.java:683)
	at org.ta4j.core.TestUtils.assertNumEquals(TestUtils.java:90)
	at org.ta4j.core.indicators.EMAIndicatorTest.usingBarCount10UsingClosePrice(EMAIndicatorTest.java:71)
```

edlins:
I may need to roll back the switch from BDN to PN in order to see why default precision PN doesn't match BDN "magic" values in unit tests.  This was my fault, I should have added PN to the AbstractIndicatorTest array and tested before I opened the PR.  Brain's a little fried right now so I'll have to put this down until tomorrow.

edlins:
Please move forward and do not wait on me here.  I will investigate the discrepancies I've found and re-submit a fresh PR when I've got a drop-in replacement ready.  Furthermore, I'll probably split off the replacement from the addition because of all of the secondary modifications required.  Thanks for your patience, and go forward with the quite excellent BigDecimalNum for now.  Stay tuned..

MarijnKoesen:
Let me know if you need some help

PR #202, opened 2018-03-02
add ContextTradingRecord with getCurrentTrade(int index)

### LINKED ISSUE #119: AnalysisCriterion start/finish indices

edlins:
Fixes #119 . (partial)

- `getCurrentTrade(int index)` gives a context for defining the "current trade"
- Fix `TradingRecord.isClosed()`
- Rename TradingRecordTest to BaseTradingRecordTest


edlins:
Maybe `getCurrentTrade(int index)` belongs in the `TradingRecord` interface?

edlins:
`getCurrentTrade(int index)` belongs in the `TradingRecord` interface.  This is to support #119 .

Major simplification of the `ContextTradingRecord`.

I would be fine with this replacing `BaseTradingRecord` as it's a huge improvement.  There is a potential concern about performance of the linear search in `getCurrentTrade(int index)` but that can be further optimized if that's needed.

I also have concerns about `TradingRecord.enter()` and `TradingRecord.exit()` simply returning `false` when they fail.  I think they should probably throw exceptions.

edlins:
Rebase, license updates, cleanup.

edlins:
Still working on this.  The CPU penalty for the linear search is significant.

edlins:
Performance is vastly improved using a `TreeMap` of `Trades` but is still slower than `BaseTradingRecord`.  May be able to do a little more cleanup.

petoncle:
I'm not sure the word "context" is the right one to describe the concept. I have some difficulties understanding what it means. 

If I understand correctly, what we want is the ability to do some analysis on a sub-window of the time series, i.e. we want to do some analysis on a subset of contiguous bars instead of all the bars. So, why not simply create a new TimeSeries that contains the subset of bars you want, and do the analysis on that time series? Would that be possible?

A minor comment about the code: you currently have some unused imports, and you could make the code a bit more modern by using the diamond operator (e.g. replace new TreeMap<Integer, Trade>() by new TreeMap<>()).

edlins:
> I'm not sure the word "context" is the right one to describe the concept.

It means, when `getCurrentTrade()` is called, it is passed an index parameter representing the time context.  Right now, `getCurrentTrade()` has zero parameters and thus only an implicit time context.  So it's like saying `what was the current trade at time index 10?` vs saying `what is the current trade?`.


> why not simply create a new TimeSeries that contains the subset of bars you want, and do the analysis on that time series?

Just seems unnecessary.  Waste of memory, and extra hoops to jump through but maybe similar CPU requirements.  But if this is the chosen method and it works I'll go with it.

It's been several months now but IIRC this PR originated from my attempts to run `VersusBuyAndHoldCriterion` on a subset of a `TimeSeries`.  It just doesn't work due to the implicit determination of `CurrentTrade` and/or the usage of `series.getBeginIndex()` / `series.getEndIndex()`.

I simply wanted to say `how did this strategy perform vs buy-and-hold from index 10 to index 100?` and was unable to do so in a straightforward manner.  Of course I'd have to dust this off to make sure IRC and provide a use case but it hasn't been a top priority.

PR #206, opened 2018-03-04
Create new DifferencePercentageIndicator

### LINKED ISSUE #199: New indicator: PercentageChange
Hi,

I'd like to create a new PercentageChange Indicator (got POC working already) to indicate if the series went up- or down in a specific percentage (based on the last bar when starting out, or otherwise the last time the indicator fired).

This can be used to do some monitoring, e.g. keep track how some markets are doing with some notifications, but also for a trade strategy to do a buy/sell when the stocks go down/up by x percent.

I read how to contribute and was asked to first create an issue for it, so hereby :)

Thanks,
Marijn

PS: Some really rough pseudo to give the idea (the pseudo doesn't work btw, but it's just to illustrate the idea):

```java
/**
 * Get the difference in percentage from the last bar, or the last time
 * the thresHold was reached.
 */
public PercentageChangeIndicator extends AbstractIndicator<Decimal> {
        // if the thresholdPercentage is set to 0
        // you always get the diff from the last bar
	public  (TimeSeries series, Decimal thresholdPercentage) {
		lastNotifcation = series.getLastBar().getClosePrice()
	}

	Decimal calculate() {	
		changePercentage = series.getLastBar().getClose() / lastNotifcation;

		if (abs(changePercentage) > thresholdPercentage) {
			lastNotifcation = series.getLastBar().getClose();
		}

		return changePercentage;
	}
}
```

MarijnKoesen:
Fixes #199 .

- New indicator to get the difference in percentage from last value

- [x ] added an entry to the unreleased section of `CHANGES.md`

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DifferencePercentage.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DifferencePercentage.java
@@ -1,0 +1,81 @@
+/*
+  The MIT License (MIT)
+
+  Copyright (c) 2014-2017 Marc de Verdelhan, Ta4j Organization & respective authors (see AUTHORS)
+
+  Permission is hereby granted, free of charge, to any person obtaining a copy of
+  this software and associated documentation files (the "Software"), to deal in
+  the Software without restriction, including without limitation the rights to
+  use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+  the Software, and to permit persons to whom the Software is furnished to do so,
+  subject to the following conditions:
+
+  The above copyright notice and this permission notice shall be included in all
+  copies or substantial portions of the Software.
+
+  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+  FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+  COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+  IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+  CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package org.ta4j.core.indicators.helpers;
+
+import org.ta4j.core.Indicator;
+import org.ta4j.core.indicators.CachedIndicator;
+import org.ta4j.core.num.NaN;
+import org.ta4j.core.num.Num;
+
+/**
+ * Difference Change Indicator.
+ * <p>
+ * Get the difference in percentage from the last time the threshold was reached.
+ * <p>
+ * Or if you don't pass the threshold you will always just get the difference percentage from the precious value.
+ * </p>
+ */
+public class DifferencePercentage extends CachedIndicator<Num> {
+    private Indicator<Num> indicator;
+    private final Num percentageThreshold;
+    private Num lastNotification;
+    private Num hundred;
+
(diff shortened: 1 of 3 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DifferencePercentage.java
```diff
@@ -0,0 +1,74 @@
...
+public class DifferencePercentage extends CachedIndicator<Num> {
+    private Indicator<Num> indicator;
+    private final Num percentageThreshold;
+    private Num lastNotification;
+
+    public DifferencePercentage(Indicator<Num> indicator) {
+        this(indicator, indicator.numOf(0));
+    }
+
+    public DifferencePercentage(Indicator<Num> indicator, Num percentageThreshold) {
```
team172011: Please use/add constructor with Number/double type for percentage threshold

MarijnKoesen:
Done, test added as well 👍

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DifferencePercentage.java
```diff
@@ -0,0 +1,78 @@
...
+
+        if (changePercentage.abs().isGreaterThanOrEqual(percentageThreshold)) {
+            lastNotification = value;
+        }
+
+        return changePercentage;
+    }
+
+    private Num fractionToPercentage(Num changeFraction) {
+        return changeFraction.multipliedBy(numOf(100)).minus(numOf(100));
```
team172011: You can store the value 100 as a field `private Num hundred` that will be initialized in the constructor and used by this function. This will be a little bit faster and does not create two new `Num ` classes every time the function is called

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DifferencePercentage.java
```diff
@@ -0,0 +1,78 @@
...
+import org.ta4j.core.Indicator;
+import org.ta4j.core.indicators.CachedIndicator;
+import org.ta4j.core.num.Num;
+
+/**
+ * Difference Change Indicator.
+ * <p>
+ * Get the difference in percentage from the last time the threshold was reached.
+ * <p>
+ * Or if you don't pass the threshold you will always just get the difference from the precious value.
```
team172011: Maybe the indicator should return the difference **in percentage** (like it does with threshold)? I think it is a little bit confusing if it returns absolute values on the one hand and relative values on the other hand

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DifferencePercentage.java
```diff
@@ -0,0 +1,78 @@
...
+        this.indicator = indicator;
+        this.percentageThreshold = percentageThreshold;
+    }
+
+    @Override
+    protected Num calculate(int index) {
+        Num value = indicator.getValue((index));
+        if (lastNotification == null) {
+            lastNotification = value;
+            return numOf(0);
```
team172011: If there is no previous value, the Indicator cannot return a valid result. In such cases ta4js convention is to return `NaN` to avoid incorrect interpretation of the result

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DifferencePercentage.java
MarijnKoesen: It's actually doing exactly that (correct me if i'm misunderstanding you), i've updated that comment and the test to make it more clear. E.g. I've changed the test case to 1000 instead of 100, this makes it easier to see that from 1000 -> 1010 will result in 1(%) return value.

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DifferencePercentage.java
MarijnKoesen: Thanks, didn't know about that, much better :)

MarijnKoesen:
Thanks @team172011 :) I've processed the feedback (and rebased it, let me know if you prefer non-rebased fixes).

team172011:
Rebasing is great, thanks

PR #208, opened 2018-03-05
Update AnalysisCriterion to Num

### LINKED ISSUE #186: Update AnalysisCriterion to Num
Update AnalysisCriterion interface to Num from double (will break all subclasses).
For each subclass:
- fix calculate() to return Num
- use Nums in internal calculations
- criteria parameters stay as primitives
- use TimeSeries.numOf() to generate Nums

This could be a good starter task for a new contributor.  ta4j needs YOU!

MarijnKoesen:
Fixes #186 .

- Fixed requirements as mentioned in #186
- Use Num in AnalysisCriterion
- But keep Criterion parameters as primitives

Notes:
* There was 1 test, namely NumberOfTradesTest::calculateWithOneTrade that passed the TimeSeries as null, but I changed this to some series, otherwise series.numOf() would give a null pointer exception,
* LinearTransactionCostCriterionTest is giving different results, it's not consistent anymore with external data, or the test data, it looks to me like a precision diff between double/Num. Please check it. I think we should change the xls/tests to reflect the new outcome.


edlins:
Great!  This might take time to validate because someone should go through every modified criterion and make sure it is internally using Num properly, even if the tests pass.  I'll take a look at LTCC as soon as I can as I've had some experience with it.

MarijnKoesen:
Sure, i'll update the PR as feedback comes in 👍 

BTW, two other things worth mentioning is `Math.pow(double, double)` was used, for example in `AverageProfitCriterion`, but this is not supported in `Num`, as it's not supported in `BigDecimal`. So to keep the calculation the same, and the test green, I had to fallback to Math.pow in the Criterion:

```java
        return series.numOf(
            Math.pow(
                totalProfit.calculate(series, tradingRecord).doubleValue(),
                1d / bars.doubleValue()
            )
        );
```

Second: as mentioned in the notes above, passing timeseries as null will result in nullpointer exceptions in some criterions, as it's used to get a `numOf()`.

I assume in normal use a TimeSeries will be passed, but in the case of `calculate(TimeSeries, Trade)`, the TimeSeries is not always used (`NumberOfBarsCriterion`), so it could have been null before, but not anymore.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterion.java
```diff
@@ -67,50 +68,51 @@ public LinearTransactionCostCriterion(double initialAmount, double a, double b)
...
             //    - Remove the cost of the *first* order
             //    - Multiply by the profit ratio
             //    - Remove the cost of the *second* order
-            tradedAmount = (tradedAmount - getOrderCost(trade.getEntry(), tradedAmount));
-            tradedAmount *= profit.calculate(series, trade);
-            tradedAmount -= getOrderCost(trade.getExit(), tradedAmount);
+            tradedAmount =
+                tradedAmount.minus(getOrderCost(trade.getEntry(), tradedAmount))
+                .multipliedBy(profit.calculate(series, trade))
+                .minus(getOrderCost(trade.getExit(), tradedAmount));
```
edlins: Classic.  Someone else had made this mistake that I fixed a while back.  Look at the original three statements and see how they differ from your combined statement.  Use the XLS test file if you want to see the intermediate calculated values broken out.  Fixing this will get you past
```
assertNumEquals(xls.getFinalCriterionValue(1000d, 0.005, 0.2).doubleValue(), value);
```

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
```diff
@@ -44,81 +44,83 @@ public LinearTransactionCostCriterionTest(Function<Number, Num> numFunction) thr
...
         TimeSeries xlsSeries = xls.getSeries();
         TradingRecord xlsTradingRecord = xls.getTradingRecord();
-        double value;
+        Num value;
 
         value = getCriterion(1000d, 0.005, 0.2).calculate(xlsSeries, xlsTradingRecord);
-        assertEquals(xls.getFinalCriterionValue(1000d, 0.005, 0.2).doubleValue(), value, TestUtils.GENERAL_OFFSET);
-        assertEquals(843.5492, value, TestUtils.GENERAL_OFFSET);
+        assertNumEquals(xls.getFinalCriterionValue(1000d, 0.005, 0.2), value);
+        assertNumEquals(834.5492, value);
```
edlins: Not sure how you got the `834.5492` unless you just ran with the bad logic in the criterion and pasted in the actual value for expected.  If you fix the criterion, this should go back to the original value.  Also change
```
assertNumEquals(xls.getFinalCriterionValue(1000d, 0.005, 0.2), value);
```
~back~ to ~the original~
```
assertNumEquals(xls.getFinalCriterionValue(1000d, 0.005, 0.2).doubleValue(), value);
```
in order to get an inexact match.  We need inexact here due to the relatively low precision of XLS files compared to `Nums`.  These comments apply to several assertions but you'll see once you fix the criterion.

EDIT: Just noticed you also switched from `assertEquals()` to `assertNumEquals()` which is fine because `assertNumEquals(double, Num)` is the same as `assertEquals(double, double, double)` with `TestUtils.GENERAL_OFFSET` as the delta.  And I'd like to run everything through `TestUtils` before it hits `JUnit`.

edlins:
> Math.pow(double, double) was used, for example in AverageProfitCriterion, but this is not supported in Num

Hmm.  @team172011 should we add `Num pow(Num power)` to `Num` or just use `Math.pow(double base, double power)`?  Strangely `BigDecimal` has `pow(int power)` but no `pow(BigDecimal power)` so `BigDecimalNum.pow(Num power)` would end up using `Math.pow(double base, double power)`.
> passing timeseries as null will result in nullpointer exceptions in some criterions

Should we update all criteria to `Objects.requireNonNull(timeSeries)`?

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterion.java
MarijnKoesen: Damn, sharp, fixed it :)

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
MarijnKoesen: Reverted eveything back to the original values, after fixing the 'chaining' issues, then just changed the `assertEquals` to `assertNumEquals` and everything was fine  :)

MarijnKoesen:
Fixed the error in `LinearTransactionCostCriterionTest.java` I've just committed it without rebasing, so you can easily check the fix, i'll rebase when all comments are processed to a single commit 👍

edlins:
LTCC looks good now.  Haven't reviewed anything else yet.

MarijnKoesen:
My initial thought on the power was to not support it, if we need to cast to double we might loose precision, but give the user the idea it is supported and should work fine. But I'm the end we get weird effects.

But I've found another solution with which i think we can do it: https://stackoverflow.com/a/3590314

This way we do a smart pow, and although still not perfect, will probably be fine for our cases.

I will implement it so we can check/evaluate the result.

MarijnKoesen:
Ok, smart pow implemented, i've rebased it a bit already, but left the fixup with the change in the criterion there for easy review 👍

edlins:
This is getting a little off-topic, but how about also fixing sqrt to the Babylonian method:
https://en.wikipedia.org/wiki/Methods_of_computing_square_roots#Babylonian_method
https://stackoverflow.com/questions/19026761/mathematics-behind-babylonian-square-root-method

@team172011 might want a separate PR for `Num.pow(Num)` and `Num.sqrt(Num)`.  If so, probably leave them as double operations in this PR.

I'm going to spend some time looking at these criteria today.

team172011:
A while ago `Num#pow` and `Num#sqrt` had been dropped from `Decimal ` class because they had to convert the delegate to double. If we succeded in implementing adequate solutions for the `BigDecimal ` delegate, like Babylonian Method, we could add them back.

team172011:
@MarijnKoesen Yes, I think criterions need the null pointer check,

team172011:
@MarijnKoesen like @edlins mentioned your [Num#pow() commit](https://github.com/ta4j/ta4j/pull/208/commits/fed227704d8c4ee776381f6a69ec9339702d5deb) should be implemented in scope of a new PR.

MarijnKoesen:
> @MarijnKoesen Yes, I think criterions need the null pointer check,

@team172011 and what should we do in case of a NullPointer? 

The only thing we can do is return NaN, but i'm not sure that's semantically correct. Isn't it better to just let it crash with a RunTimeException (NullPointerException) instead, and don't do anything?

> @MarijnKoesen like @edlins mentioned your Num#pow() commit should be implemented in scope of a new PR.

Yeah i've created a new PR for this: https://github.com/ta4j/ta4j/pull/212

@edlins there are some implementation for Babylonian here (be careful though, some are only for BigInteger) https://stackoverflow.com/questions/13649703/square-root-of-bigdecimal-in-java might be worth checking out indeed. I'm not sure if you looked into it already, let me know, else I can take a look and create a PR for the Babylonian implementation in BigDecimal.

team172011:
@MarijnKoesen 
> what should we do in case of a NullPointer?

`Objects.requierNotNull(series, "Series must be not null") ` and Javadoc commet should do it!

> Yeah i've created a new PR for this: #212

👍

MarijnKoesen:
I've added non-null check &
 java doc [in this commit](https://github.com/ta4j/ta4j/pull/208/commits/7920ff5995b98c75fb1d8b02b28263bbb68fbe8c) I didn't do it in *every* criterion, as it would arguably be too defensive, and in that case we should do it in almost every method throughout the project, which would anyway be OOS for this PR again.

edlins:
> I can take a look and create a PR for the Babylonian implementation in BigDecimal.

That would be awesome!  Opened #213 .

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/AverageProfitCriterion.java
```diff
@@ -39,25 +40,27 @@ this software and associated documentation files (the "Software"), to deal in
...
+        return totalProfit.calculate(series, trade).pow(series.numOf(1).dividedBy(bars));
     }
 
     @Override
-    public boolean betterThan(double criterionValue1, double criterionValue2) {
-        return criterionValue1 > criterionValue2;
+    public boolean betterThan(Num criterionValue1, Num criterionValue2) {
+        return criterionValue1.isGreaterThan(criterionValue2);
     }
 }
```
edlins: AverageProfitCriterion looks good.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/AverageProfitableTradesCriterion.java
```diff
@@ -70,11 +70,11 @@ public double calculate(TimeSeries series, TradingRecord tradingRecord) {
...
+        return series.numOf(numberOfProfitable).dividedBy(series.numOf(tradingRecord.getTradeCount()));
     }
 
     @Override
-    public boolean betterThan(double criterionValue1, double criterionValue2) {
-        return criterionValue1 > criterionValue2;
+    public boolean betterThan(Num criterionValue1, Num criterionValue2) {
+        return criterionValue1.isGreaterThan(criterionValue2);
     }
 }
```
edlins: AverageProfitableTrades looks good.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/BuyAndHoldCriterion.java
```diff
@@ -34,24 +35,24 @@ this software and associated documentation files (the "Software"), to deal in
...
         }
     }
 
     @Override
-    public boolean betterThan(double criterionValue1, double criterionValue2) {
-        return criterionValue1 > criterionValue2;
+    public boolean betterThan(Num criterionValue1, Num criterionValue2) {
+        return criterionValue1.isGreaterThan(criterionValue2);
     }
 }
```
edlins: BuyAndHold looks good.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterion.java
```diff
@@ -121,17 +122,17 @@ private double getOrderCost(Order order, double tradedAmount) {
...
                     // To calculate the new traded amount:
                     //    - Remove the cost of the first order
                     //    - Multiply by the profit ratio
-                    double newTradedAmount = (initialAmount - totalTradeCost) * profit.calculate(series, trade);
-                    totalTradeCost += getOrderCost(trade.getExit(), newTradedAmount);
+                    Num newTradedAmount = initialAmount.minus(totalTradeCost).multipliedBy(profit.calculate(series, trade));
+                    totalTradeCost = totalTradeCost.plus(getOrderCost(trade.getExit(), newTradedAmount));
                 }
             }
         }
```
edlins: LTCC looks good.  :1st_place_medal:

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/MaximumDrawdownCriterion.java
```diff
@@ -35,25 +35,25 @@ this software and associated documentation files (the "Software"), to deal in
...
-        return maximumDrawdown.doubleValue();
+        return maximumDrawdown;
     }
 
     @Override
-    public double calculate(TimeSeries series, Trade trade) {
+    public Num calculate(TimeSeries series, Trade trade) {
         if (trade != null && trade.getEntry() != null && trade.getExit() != null) {
             CashFlow cashFlow = new CashFlow(series, trade);
             Num maximumDrawdown = calculateMaximumDrawdown(series, cashFlow);
```
edlins: MaximumDrawdown looks good.  Interesting that `calculateMaximumDrawdown()` was already converted to `Num`..

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfBarsCriterion.java
```diff
@@ -34,21 +37,25 @@ this software and associated documentation files (the "Software"), to deal in
...
+            nBars = nBars.plus(calculate(series, trade));
         }
         return nBars;
     }
 
     @Override
-    public double calculate(TimeSeries series, Trade trade) {
-        return (1 + trade.getExit().getIndex()) - trade.getEntry().getIndex();
+    public Num calculate(TimeSeries series, Trade trade) {
+        Objects.requireNonNull(series, "Series must be not null");
```
edlins, line 51: NumberOfBars looks good.  Still wondering about dealing with `nulls` more consistently..

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfTradesCriterion.java
```diff
@@ -25,24 +25,25 @@ this software and associated documentation files (the "Software"), to deal in
...
+        return series.numOf(1);
     }
 
     @Override
-    public boolean betterThan(double criterionValue1, double criterionValue2) {
-        return criterionValue1 < criterionValue2;
+    public boolean betterThan(Num criterionValue1, Num criterionValue2) {
+        return criterionValue1.isLessThan(criterionValue2);
     }
 }
```
edlins: NumberOfTrades looks good.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/RewardRiskRatioCriterion.java
```diff
@@ -39,17 +40,17 @@ this software and associated documentation files (the "Software"), to deal in
...
+        return criterionValue1.isGreaterThan(criterionValue2);
     }
 
     @Override
-    public double calculate(TimeSeries series, Trade trade) {
-        return totalProfit.calculate(series, trade) / maxDrawdown.calculate(series, trade);
+    public Num calculate(TimeSeries series, Trade trade) {
+        return totalProfit.calculate(series, trade).dividedBy(maxDrawdown.calculate(series, trade));
     }
 }
```
edlins: RewardRiskRatio looks good.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/TotalProfitCriterion.java
```diff
@@ -76,6 +76,6 @@ private double calculateProfit(TimeSeries series, Trade trade) {
                 profit = entryClosePrice.dividedBy(exitClosePrice);
             }
         }
-        return profit.doubleValue();
+        return profit;
     }
 }
```
edlins: TotalProfit looks good.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/VersusBuyAndHoldCriterion.java
```diff
@@ -42,26 +43,26 @@ public VersusBuyAndHoldCriterion(AnalysisCriterion criterion) {
...
     }
 
     @Override
-    public boolean betterThan(double criterionValue1, double criterionValue2) {
-        return criterionValue1 > criterionValue2;
+    public boolean betterThan(Num criterionValue1, Num criterionValue2) {
+        return criterionValue1.isGreaterThan(criterionValue2);
     }
 
     @Override
```
edlins: VersusBuyAndHold looks good.  Can probably remove `toString()`.  Would be a tiny change so I'm fine with making that change in this PR.

ta4j-core/src/main/java/org/ta4j/core/num/BigDecimalNum.java
```diff
@@ -181,6 +181,36 @@ public Num pow(int n) {
...
+        BigDecimal doublePow = new BigDecimal(Math.pow(delegate.doubleValue(), remainderOf2.doubleValue()));
+        result = intPow.multiply(doublePow);
+
+        if (signOfPower == -1) {
+            result = BigDecimal.ONE.divide(result, MATH_CONTEXT);
+        }
+
+        return new BigDecimalNum(result);
+    }
+
```
edlins: - [ ] Please remove per #212 .

ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
```diff
@@ -91,6 +91,11 @@ public Num pow(int n) {
         return new DoubleNum(Math.pow(delegate,n));
     }
 
+    @Override
+    public Num pow(Num n) {
+        return new DoubleNum(Math.pow(delegate,n.doubleValue()));
+    }
+
```
edlins: - [ ] Please remove per #212

ta4j-core/src/main/java/org/ta4j/core/num/NaN.java
```diff
@@ -107,6 +107,11 @@ public Num pow(int n) {
         return this;
     }
 
+    @Override
+    public Num pow(Num n) {
+        return this;
+    }
+
```
edlins: - [ ] Please remove per #212

ta4j-core/src/main/java/org/ta4j/core/num/Num.java
```diff
@@ -89,6 +89,13 @@ this software and associated documentation files (the "Software"), to deal in
      */
     Num pow(int n);
 
+    /**
+     * Returns a {@code num} whose value is <tt>(this<sup>n</sup>)</tt>.
+     * @param n power to raise this {@code num} to.
+     * @return <tt>this<sup>n</sup></tt>
+     */
+    Num pow(Num n);
+
```
edlins: - [ ] Please remove per #212

ta4j-core/src/test/java/org/ta4j/core/NumTest.java
```diff
@@ -105,6 +105,9 @@ public void testNaN(){
         mustBeNaN = a.pow(12);
         assertNumEquals(mustBeNaN,NaN);
 
+        mustBeNaN = a.pow(a);
+        assertNumEquals(mustBeNaN,NaN);
+
```
edlins: - [ ] Please remove per #212

ta4j-core/src/test/java/org/ta4j/core/NumTest.java
```diff
@@ -140,6 +143,8 @@ public void testArithmetic(){
         assertNumEquals(0,hundredMillion.remainder(five));
 
         assertNumEquals(0.00032, zeroDotTwo.pow(5));
+        assertNumEquals(0.7247796636776955, zeroDotTwo.pow(zeroDotTwo));
+        assertNumEquals(1.37972966146, zeroDotTwo.pow(numOf(-0.2)));
```
edlins: - [ ] Please remove per #212

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AbstractCriterionTest.java
```diff
@@ -35,7 +36,7 @@ this software and associated documentation files (the "Software"), to deal in
 import java.util.function.Function;
 
 @RunWith(Parameterized.class)
-public abstract class AbstractCriterionTest {
+public abstract class AbstractCriterionTest extends TestUtils {
```
edlins: - [ ] I don't believe this is necessary.  Remove `extends` clause and `import`.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AverageProfitCriterionTest.java
```diff
@@ -63,28 +63,28 @@ public void calculateOnlyWithLossTrades() {
...
 
     @Test
     public void betterThan() {
         AnalysisCriterion criterion = new AverageProfitCriterion();
-        assertTrue(criterion.betterThan(2.0, 1.5));
-        assertFalse(criterion.betterThan(1.5, 2.0));
+        assertTrue(criterion.betterThan(numOf(2.0), numOf(1.5)));
+        assertFalse(criterion.betterThan(numOf(1.5), numOf(2.0)));
     }
 }
```
edlins, line 92: AverageProfit Test looks good.  Like the use of `assertNumEquals()`.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AverageProfitableTradesCriterionTest.java
```diff
@@ -55,16 +55,16 @@ public void calculateWithOneTrade() {
...
 
     @Test
     public void betterThan() {
         AnalysisCriterion criterion = new AverageProfitableTradesCriterion();
-        assertTrue(criterion.betterThan(12, 8));
-        assertFalse(criterion.betterThan(8, 12));
+        assertTrue(criterion.betterThan(numOf(12), numOf(8)));
+        assertFalse(criterion.betterThan(numOf(8), numOf(12)));
     }
 }
```
edlins, line 73: AverageProfitableTrades Test looks good.  Am wondering if the `double` calculations like `2d/3` should be replaced with `Num` calculations..

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/BuyAndHoldCriterionTest.java
```diff
@@ -56,29 +56,29 @@ public void calculateOnlyWithLossTrades() {
...
 
     @Test
     public void betterThan() {
         AnalysisCriterion criterion = new BuyAndHoldCriterion();
-        assertTrue(criterion.betterThan(1.3, 1.1));
-        assertFalse(criterion.betterThan(0.6, 0.9));
+        assertTrue(criterion.betterThan(numOf(1.3), numOf(1.1)));
+        assertFalse(criterion.betterThan(numOf(0.6), numOf(0.9)));
     }
 }
```
edlins, line 86: BuyAndHold Test looks good.  Am wondering if `double` values like `0.7` should be replaced with `Nums`.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
```diff
@@ -44,81 +44,81 @@ public LinearTransactionCostCriterionTest(Function<Number, Num> numFunction) thr
...
 
     @Test
     public void betterThan() {
         AnalysisCriterion criterion = new LinearTransactionCostCriterion(1000, 0.5);
-        assertTrue(criterion.betterThan(3.1, 4.2));
-        assertFalse(criterion.betterThan(2.1, 1.9));
+        assertTrue(criterion.betterThan(numOf(3.1), numOf(4.2)));
+        assertFalse(criterion.betterThan(numOf(2.1), numOf(1.9)));
     }
 }
```
edlins, line 126: LTCC Test looks good.  :1st_place_medal:

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/MaximumDrawdownCriterionTest.java
```diff
@@ -96,13 +96,13 @@ public void withSimpleTrades() {
...
 
     @Test
     public void betterThan() {
         AnalysisCriterion criterion = new MaximumDrawdownCriterion();
-        assertTrue(criterion.betterThan(0.9, 1.5));
-        assertFalse(criterion.betterThan(1.2, 0.4));
+        assertTrue(criterion.betterThan(numOf(0.9), numOf(1.5)));
+        assertFalse(criterion.betterThan(numOf(1.2), numOf(0.4)));
     }
 }
```
edlins, line 110: MaximumDrawdown Test looks good.  Am wondering if `double` parametered `numOf(0.9)` should be `String` parametered `numOf("0.9")`

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/NumberOfBarsCriterionTest.java
```diff
@@ -53,21 +53,21 @@ public void calculateWithTwoTrades() {
...
 
     @Test
     public void betterThan() {
         AnalysisCriterion criterion = new NumberOfBarsCriterion();
-        assertTrue(criterion.betterThan(3, 6));
-        assertFalse(criterion.betterThan(6, 2));
+        assertTrue(criterion.betterThan(numOf(3), numOf(6)));
+        assertFalse(criterion.betterThan(numOf(6), numOf(2)));
     }
 }
```
edlins, line 75: NumberOfBars Test looks good.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/NumberOfTradesCriterionTest.java
```diff
@@ -53,21 +53,22 @@ public void calculateWithTwoTrades() {
...
         Trade trade = new Trade();
         NumberOfTradesCriterion tradesCriterion = new NumberOfTradesCriterion();
 
-        assertEquals(1d, tradesCriterion.calculate(null, trade), TestUtils.GENERAL_OFFSET);
+        assertNumEquals(1, tradesCriterion.calculate(series, trade));
     }
 
     @Test
     public void betterThan() {
         AnalysisCriterion criterion = new NumberOfTradesCriterion();
```
edlins: - [ ] Please change `new NumberOfTradesCriterion()` to `getCriterion()`.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/RewardRiskRatioCriterionTest.java
```diff
@@ -59,7 +59,7 @@ public void rewardRiskRatioCriterion() {
         double peak = (105d / 100) * (100d / 95);
         double low = (105d / 100) * (90d / 95) * (80d / 95);
 
-        assertEquals(totalProfit / ((peak - low) / peak), rrc.calculate(series, tradingRecord), TestUtils.GENERAL_OFFSET);
+        assertEquals(totalProfit / ((peak - low) / peak), rrc.calculate(series, tradingRecord).doubleValue(), GENERAL_OFFSET);
```
edlins: - [ ] Please change to use `assertNumEquals()`

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/RewardRiskRatioCriterionTest.java
```diff
@@ -68,30 +68,28 @@ public void rewardRiskRatioCriterionOnlyWithGain() {
...
-
-
         RewardRiskRatioCriterion ratioCriterion = new RewardRiskRatioCriterion();
-        assertEquals((95d/100) / ((1d - 0.95d)), TestUtils.GENERAL_OFFSET, ratioCriterion.calculate(series, trade));
+        assertEquals((95d/100) / ((1d - 0.95d)), ratioCriterion.calculate(series, trade).doubleValue(), GENERAL_OFFSET);
     }
 
     @Test
     public void betterThan() {
         AnalysisCriterion criterion = new RewardRiskRatioCriterion();
```
edlins, line 93: ~This criterion does not extend `AbstractCriterionTest`.  OOS?~ Wrong..

- [ ] Please change `new RewardRiskRatioCriterion()` to `getCriterion()`

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/TotalProfitCriterionTest.java
```diff
@@ -67,31 +67,31 @@ public void calculateProfitWithTradesThatStartSelling() {
...
 
     @Test
     public void betterThan() {
         AnalysisCriterion criterion = new TotalProfitCriterion();
-        assertTrue(criterion.betterThan(2.0, 1.5));
-        assertFalse(criterion.betterThan(1.5, 2.0));
+        assertTrue(criterion.betterThan(numOf(2.0), numOf(1.5)));
+        assertFalse(criterion.betterThan(numOf(1.5), numOf(2.0)));
     }
 }
```
edlins, line 99: TotalProfit Test looks good (aside from globally applicable previous comments).

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/VersusBuyAndHoldCriterionTest.java
```diff
@@ -98,13 +98,13 @@ public void calculateWithNumberOfBars() {
...
 
     @Test
     public void betterThan() {
         AnalysisCriterion criterion = new VersusBuyAndHoldCriterion(new TotalProfitCriterion());
-        assertTrue(criterion.betterThan(2.0, 1.5));
-        assertFalse(criterion.betterThan(1.5, 2.0));
+        assertTrue(criterion.betterThan(numOf(2.0), numOf(1.5)));
+        assertFalse(criterion.betterThan(numOf(1.5), numOf(2.0)));
     }
 }
```
edlins, line 112: VersusBuyAndHold Test looks good.  After `Num.pow(Num)` is merged we'll want to migrate all of the criteria away from `Math.pow`.

ta4j-core/src/test/java/org/ta4j/core/mocks/MockAnalysisCriterion.java
```diff
@@ -52,8 +52,8 @@ public double calculate(TimeSeries series, TradingRecord tradingRecord) {
      * @param criterionValue2 second value
      * @return boolean indicating first value is greater than second value
      */
-    public boolean betterThan(double criterionValue1, double criterionValue2) {
-        return (criterionValue1 > criterionValue2);
+    public boolean betterThan(Num criterionValue1, Num criterionValue2) {
+        return criterionValue1.isGreaterThan(criterionValue2);
     }
 
 }
```
edlins: MockAnalysisCriterion looks good.

ta4j-examples/src/main/java/ta4jexamples/num/DoubleNumVsBigDecimalNum.java
```diff
@@ -50,7 +51,7 @@ public static void test(TimeSeries series){
         TimeSeriesManager manager = new TimeSeriesManager(series);
         TradingRecord record1 = manager.run(strategy1);
         TotalProfitCriterion profit1 = new TotalProfitCriterion();
-        double profitResult1 = profit1.calculate(series, record1);
+        Num profitResult1 = profit1.calculate(series, record1);
         long end = System.currentTimeMillis();
 
         System.out.printf("[%s]\n" +
```
edlins: looks good

ta4j-examples/src/main/java/ta4jexamples/walkforward/WalkForward.java
```diff
@@ -174,7 +175,7 @@ public static void main(String[] args) {
                 String name = entry.getValue();
                 // For each strategy...
                 TradingRecord tradingRecord = sliceManager.run(strategy);
-                double profit = profitCriterion.calculate(slice, tradingRecord);
+                Num profit = profitCriterion.calculate(slice, tradingRecord);
                 System.out.println("\tProfit for " + name + ": " + profit);
             }
             Strategy bestStrategy = profitCriterion.chooseBest(sliceManager, new ArrayList<Strategy>(strategies.keySet()));
```
edlins: looks good

edlins:
I also saw a number of `numOf(0)` and `numOf(1)` calls that used to be `static` `Num.ZERO` and `Num.ONE` for performance reasons.  These invocation count is low and it's test, so probably doesn't matter.  Kind of wondering if we need to implement those for core usage..

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfBarsCriterion.java
MarijnKoesen: Yeah me too, I'm doubting if it's really worth it in the end.

The `Objects.requireNonNull(series, "Series must be not null");` is only to have a 'better readable' error message, but it's basically the same as not doing it, bar the extra message.

```java
java.lang.NullPointerException at 
org.ta4j.core.analysis.criteria.NumberOfBarsCriterion.calculate(NumberOfBarsCriterion.java:41)
```
Would be the error without the error message, so in that sense it *is* better readable. But it's at a cost of being more bloated and how far do you go?

Take this line in the calculation for example:

```java
            (series.numOf(1).plus(series.numOf(trade.getExit().getIndex())))
```

What if the trade is only created using this constructor:

```java
public Trade(OrderType startingType)
``` 

In that case there is no exit order, so we'd still get a null pointer exception.

So to 'properly' do the null checks in this method we'd need to do:

```java
    @Override
    public Num calculate(TimeSeries series, Trade trade) {
        Objects.requireNonNull(series, "Series must be not null");
        Objects.requireNonNull(trade, "Trade must be not null");
        Objects.requireNonNull(trade.getExit(), "Trade's Exit Order must be not null");
        Objects.requireNonNull(trade.getEntry(), "Trade's Entry Order must be not null");

        return
            (series.numOf(1).plus(series.numOf(trade.getExit().getIndex())))
            .minus(series.numOf(trade.getEntry().getIndex()));
    }
```

And then i'm not even talking about any criterion that's doing for loops over the `TradeRecord`. 

Maybe we should just trust on the user of the Criterions to pass the correct data, and if they don't, they'll just have to handle the NullPointerExceptions...

ta4j-core/src/main/java/org/ta4j/core/num/BigDecimalNum.java
MarijnKoesen: Yeah, my idea is to merge/get merged #212 first, so I can rebase this, and make this only 'clean'.

edlins: Okay.  The #212 requested changes should hold this PR until #212 is merged into develop.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/NumberOfTradesCriterionTest.java
MarijnKoesen, line 72: It's not possible as there is no factory defined in this test (I get a , also none of the other Criterions do it that way, am I missing something?

This is the exception I get if I change it:

```java
java.lang.NullPointerException
	at org.ta4j.core.analysis.criteria.AbstractCriterionTest.getCriterion(AbstractCriterionTest.java:71)
	at org.ta4j.core.analysis.criteria.NumberOfBarsCriterionTest.betterThan(NumberOfBarsCriterionTest.java:69)
[..]
```java

File: ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfBarsCriterion.java (hunk 8 above)
Comment (edlins) [reply ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfBarsCriterion.java:51]: Another possibility is this legacy logic in `MaximumDrawdownCriterion`:
```
    @Override
    public double calculate(TimeSeries series, Trade trade) {
        if (trade != null && trade.getEntry() != null && trade.getExit() != null) {
            CashFlow cashFlow = new CashFlow(series, trade);
            Num maximumDrawdown = calculateMaximumDrawdown(series, cashFlow);
            return maximumDrawdown.doubleValue();
        }
        return 0;
    }
```
But that just "hides" the null parameters and returns an actual `double`.  Very misleading!

I suggest we just leave all of these as they currently are and after this is merged open a new issue to standardize handling of criteria parameters regarding `nulls`.

File: ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/RewardRiskRatioCriterionTest.java (hunk 28 above)
Comment (MarijnKoesen) [reply ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/RewardRiskRatioCriterionTest.java:93]: Same as above, doesn't work, basically I see two options:

a)

```java
    @Before
    public void setUp() {
        rrc = new RewardRiskRatioCriterion();
    }

    @Test
    public void betterThan() {
        assertTrue(rrc.betterThan(numOf(3.5), numOf(2.2)));
        assertFalse(rrc.betterThan(numOf(1.5), numOf(2.7)));
    }
``

b)

```java
    public RewardRiskRatioCriterionTest(Function<Number, Num> numFunction) {
        super((params) -> new RewardRiskRatioCriterion(), numFunction);
    }

    @Test
    public void betterThan() {
        assertTrue(getCriterion().betterThan(numOf(3.5), numOf(2.2)));
        assertFalse(getCriterion().betterThan(numOf(1.5), numOf(2.7)));
    }
```

`A` is good enough for all tests minus 1, as only 1 of the Criterion takes arguments, and I think `A` is a little easier to read/understand.

If we do that we should change it on all other places in the tests as well, as there are a lot of places where we create an instance of the creation, so to keep it consistent I propose to change it in *all* the places.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/VersusBuyAndHoldCriterionTest.java
MarijnKoesen: Yeah there's 1 indicator left, and a couple of tests cases in the criterion, i'll fix those criterion tests in this PR already. Then we only have 1 indicator left to fix.

MarijnKoesen:
Thanks for the review @edlins , i've processed the feedback that needs no discussion, and gave a response on some of them, so there is still two things to do after we're aligned on the solution: the `getCriterion()` and `nullchecks` and rebasing with the Num.pow(Num) after that's merged.

And regarding the `Num.ZERO` and `Num.ONE`, I thought exactly the same when typing that often... I think it would be nice to implement them. But we could do it in another PR, and easily replaceall the cases (27 times numOf(1), and 69times numOf(0).

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/NumberOfTradesCriterionTest.java
edlins: I see.  `AbstractCriterionTest(Function)` should be removed and criteria fixed.  Separate issue.  Ignore this requested change.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfBarsCriterion.java
edlins: OOS, opened #216 .  Leaving this thread intact for reference but not holding up the PR.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/NumberOfTradesCriterionTest.java
edlins: OOS, opened #214

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/RewardRiskRatioCriterionTest.java
edlins: B.  Facilitates new test case creation by copy-and-paste from existing examples with minimal editing.  Consistent with more recently edited test classes.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/TotalProfitCriterionTest.java
edlins: OOS

edlins:
> there is still two things to do after we're aligned on the solution: the getCriterion() and nullchecks

OOS, #214 and #216 .

> regarding the Num.ZERO and Num.ONE

OOS, open a new PR/Issue if you like.  Perf impact is likely not significant because it's unlikely that user app will have a lot of these.

edlins:
> rebasing with the Num.pow(Num)

OOS, #217 after merge.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/VersusBuyAndHoldCriterionTest.java
edlins: OOS, #217

MarijnKoesen:
> @team172011 My suggestion now is to FIRST review and merge #212 , then re-review/merge this PR, then implement #213 , and finally #217 . #214 and #216 are totally separate. (I think) Been a long day, please double-check this suggestion.

Fully agree, seems like the best approach. 

This PR does what it says and should, the other issues are improvements of already existing structures. We can do that separately, also to avoid blowing this up further.

I've already rebased this PR against #212, and is now ready to merge (directly after #212 is merged).

team172011:
Okay, I have merged #212, so you can go on with your implementations.  I will only be able to get involved next weekend

edlins:
I can re-review this tomorrow.  However, I'd still like to see responses to my tardy review of #212 (after the merge).

MarijnKoesen:
Ok, i've rebased this one, all feedback is processed. This can be merged imo.

Any comments on #212 is OOS for this PR and should be addressed in a followup for #212 👍

edlins:
Thanks for rebasing.  I'll look at this today.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AbstractCriterionTest.java
edlins: I believe this comment still stands.  Please remove these changes.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/RewardRiskRatioCriterionTest.java
edlins: I'm still recommending B here.  Regardless, #214 will force option B so we can call this OOS and allow it.

MarijnKoesen:
> I believe this comment still stands. Please remove these changes.

I am confused, do you mean removing AbstractCriterionTest(Function)? That was said to be [OOS](https://github.com/ta4j/ta4j/issues/214). And if I don't extend TestUtils we don't have assertNumEquals.

> I'm still recommending B here. Regardless, #214 will force option B so we can call this OOS and allow it.

Yep, but to clarify, this code was already like this, I didn't touch it, and it was said to be OOS already. 

I had some other changes in this PR before and needed to create another PR for that as this was not in scope here. So let's keep this as is please, this is about updating the AnalysisCriterion to Num, not about refactoring all the criterion creation in all Criterion tests (although I agree it would be better) :)

edlins:
Right, all you have to do now is:
remove line 29 from AbstractCriterionTest
change line 39 from
```
public abstract class AbstractCriterionTest extends TestUtils {
```
back to
```
public abstract class AbstractCriterionTest {
```

MarijnKoesen:
As I mentioned in my comment https://github.com/ta4j/ta4j/pull/208#issuecomment-372772477:  

> If I don't extend TestUtils we don't have assertNumEquals.

edlins:
```
import static org.ta4j.core.TestUtils.assertNumEquals;
```

edlins:
For the record, the import was not at all required in AbstractCriterionTest, and the contributor has removed it.
**APPROVED!**

PR #212, opened 2018-03-08
Add support for `Num.pow(Num)`

MarijnKoesen:
Fixes #.

- Add support for `Num.pow(Num)`

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/num/BigDecimalNum.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/num/BigDecimalNum.java
@@ -182,6 +182,36 @@ public final class BigDecimalNum implements Num {
     }
 
     /**
+     * Returns a {@code Num} whose value is <tt>(this<sup>n</sup>)</tt>.
+     * @param n power to raise this {@code Num} to.
+     * @return <tt>this<sup>n</sup></tt>
+     * @see BigDecimal#pow(int, java.math.MathContext)
+     */
+    public Num pow(Num n) {
+        // Because BigDecimal.pow(BigDecimal) is not supported we need to find another way, we could do:
+        // Math.pow(a.doubleValue(), b.doubleValue()), but there is a better way:
+        // Perform X^(A+B)=X^A*X^B (B = remainder)
+        // As suggested: https://stackoverflow.com/a/3590314
+        BigDecimal result;
+        BigDecimal power = ((BigDecimalNum)n).delegate;
+        int signOfPower = power.signum();
+        power = power.multiply(new BigDecimal(signOfPower));
+
+        BigDecimal remainderOf2 = power.remainder(BigDecimal.ONE);
+        BigDecimal n2IntPart = power.subtract(remainderOf2);
+
+        BigDecimal intPow = delegate.pow(n2IntPart.intValueExact(), MATH_CONTEXT);
+        BigDecimal doublePow = new BigDecimal(Math.pow(delegate.doubleValue(), remainderOf2.doubleValue()));
+        result = intPow.multiply(doublePow);
+
+        if (signOfPower == -1) {
+            result = BigDecimal.ONE.divide(result, MATH_CONTEXT);
+        }
+
+        return new BigDecimalNum(result);
+    }
+
+    /**
      * Returns the correctly rounded positive square root of the <code>double</code> value of this {@code Num}.
      * /!\ Warning! Uses the {@code StrictMath#sqrt(double)} method under the hood.
      * @return the positive square root of {@code this}
--- a/ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
@@ -92,6 +92,11 @@ public class DoubleNum implements Num {
     }
 
     @Override
+    public Num pow(Num n) {
+        return new DoubleNum(Math.pow(delegate,n.doubleValue()));
+    }
+
+    @Override
     public Num abs() {
         return new DoubleNum(Math.abs(delegate));
     }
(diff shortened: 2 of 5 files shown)


edlins:
Note: after merge, refactor all/as-many-as-possible `Math.pow()` calls to `Num.pow()` calls.

edlins:
I'm out today but this is the priority.

team172011:
Merged because of high priority dependency

ta4j-core/src/main/java/org/ta4j/core/num/BigDecimalNum.java
```diff
@@ -181,6 +181,36 @@ public Num pow(int n) {
...
+     * @return <tt>this<sup>n</sup></tt>
+     * @see BigDecimal#pow(int, java.math.MathContext)
+     */
+    public Num pow(Num n) {
+        // Because BigDecimal.pow(BigDecimal) is not supported we need to find another way, we could do:
+        // Math.pow(a.doubleValue(), b.doubleValue()), but there is a better way:
+        // Perform X^(A+B)=X^A*X^B (B = remainder)
+        // As suggested: https://stackoverflow.com/a/3590314
+        BigDecimal result;
+        BigDecimal power = ((BigDecimalNum)n).delegate;
```
edlins, line 196: Put this in `NumTest` and it fails on `ClassCastException`:
```
        Num fiveTest = numOf(5);
        Num fiveDouble = DoubleNum.valueOf(5);
        assertNumEquals(3125, fiveTest.pow(fiveDouble));
```

edlins: Sometimes CCE is good.  Seems wrong here.  Open for discussion.  Any `Num` should be able to be raised to the `pow` of any other `Num` regardless of implementation.

ta4j-core/src/main/java/org/ta4j/core/num/BigDecimalNum.java
```diff
@@ -181,6 +181,36 @@ public Num pow(int n) {
...
+        int signOfPower = power.signum();
+        power = power.multiply(new BigDecimal(signOfPower));
```
edlins, line 198: ~Why not power = power.abs()?~

edlins: ~Or, however `power` is initially defined above, just add `.abs()` to the end of the definition.~
See my final comment on this file.

ta4j-core/src/main/java/org/ta4j/core/num/BigDecimalNum.java
```diff
@@ -181,6 +181,36 @@ public Num pow(int n) {
...
+
+        BigDecimal remainderOf2 = power.remainder(BigDecimal.ONE);
+        BigDecimal n2IntPart = power.subtract(remainderOf2);
```
edlins, line 201: OOS, but there should be a `Num Num.floor()`...

ta4j-core/src/main/java/org/ta4j/core/num/BigDecimalNum.java
```diff
@@ -181,6 +181,36 @@ public Num pow(int n) {
...
+        if (signOfPower == -1) {
+            result = BigDecimal.ONE.divide(result, MATH_CONTEXT);
+        }
+
+        return new BigDecimalNum(result);
+    }
+
     /**
      * Returns the correctly rounded positive square root of the <code>double</code> value of this {@code Num}.
      * /!\ Warning! Uses the {@code StrictMath#sqrt(double)} method under the hood.
```
edlins, line 216: I re-wrote this as:
```
        BigDecimal power = (((BigDecimalNum) n).delegate);
        BigDecimal remainderOf2 = power.remainder(BigDecimal.ONE);
        BigDecimal n2IntPart = power.subtract(remainderOf2);
        BigDecimal intPow = delegate.pow(n2IntPart.intValueExact(), MATH_CONTEXT);
        BigDecimal doublePow = new BigDecimal(Math.pow(delegate.doubleValue(), remainderOf2.doubleValue()));
        BigDecimal result = intPow.multiply(doublePow);
        return new BigDecimalNum(result);
```
And got the same unit test results.

ta4j-core/src/main/java/org/ta4j/core/num/BigDecimalNum.java
edlins: Just double checked and this may actually be consistent with all of the other operations.  It just seems wrong to me that you can't
```
Num a = DoubleNum.valueOf("1.0");
Num b = BigDecimalNum.valueOf("2.0");
assertNumEquals("3.0", a.plus(b));
```
But I'm willing to let this go for now.

edlins:
It would also be slick to see test cases for all of the overflow conditions.

ta4j-core/src/main/java/org/ta4j/core/num/BigDecimalNum.java
team172011: > Any Num should be able to be raised to the pow of any other Num regardless of implementation.

I do not agree with this assumption. `Num#pow` could be overloaded with int/double, but different `Num` implementations should not be compatible in any way.

edlins: Okay, I'll let it go.  :)  I now approve of line 196.

edlins:
My current assessment is that the unnecessary operations on line 198 and 208 should be removed per my [comment](https://github.com/ta4j/ta4j/pull/212#discussion_r173636408) above.

MarijnKoesen:
I think it would be nice to create a followup with the simplified implementation and tests 👍

PR #215, opened 2018-03-08
Fixed copyright notice from 2014-2017 to 2014-2018

MarijnKoesen:
Fixes #.

- Fixed copyright notice from 2014-2017 to 2014-2018
- 
- 

Minor doc thingy.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/AnalysisCriterion.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/AnalysisCriterion.java
@@ -1,7 +1,7 @@
 /*
   The MIT License (MIT)
 
-  Copyright (c) 2014-2017 Marc de Verdelhan, Ta4j Organization & respective authors (see AUTHORS)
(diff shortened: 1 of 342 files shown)


edlins:
@team172011 How does the license-maven-plugin work?

team172011:
@edlins I think the `mvn licence` command should replace all header comments with the content in the LICENCE file. But actually the old licence comment will not be replaced, but the new will just be added to each file. This is because of much manually changes in licence so that the maven plugin does not detect it as comment anymore

team172011:
I will update it with maven plugin

edlins:
Cool.  How do you recommend contributors deal with license headers?  Don't touch, and do not add a license to new files?

team172011:
I would say: Do not touch and add a exact copy of it to every new created java file. The eclipse Plug-In I use is very smart detecting changes, but other tools like our mvn plugin will not recognize it if some minor changes are done

PR #221, opened 2018-03-11
Remove unused in Quickstart

izeye:
This PR removes an unused statement in `Quickstart`.

### DIFF
--- a/ta4j-examples/src/main/java/ta4jexamples/Quickstart.java
+++ b/ta4j-examples/src/main/java/ta4jexamples/Quickstart.java
@@ -24,7 +24,6 @@
 package ta4jexamples;
 
 import org.ta4j.core.*;
-import org.ta4j.core.analysis.CashFlow;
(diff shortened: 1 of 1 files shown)


edlins:
Hello, and thank you for your interest!  Please read
https://ta4j.github.io/ta4j-wiki/How-to-contribute
PRs against `master` branch are not accepted.  Please PR against `develop`.

izeye:
@edlins Thanks for the feedback! I should have read it first. I'll try to change the target branch to the `develop` if possible here or create another one.

izeye:
@edlins Re-targeted to the `develop` branch now.

edlins:
Much better, thanks!
I've reviewed this and it's fine to merge.  However, what might be even better is to actually use the `CashFlow` and print a message with `getValue()` on the end index.  But that's optional.

team172011:
@edlins I also thought to use it instead of deleting it, but it would require more than just a getValue call (it must print several cash flow values to make sense and this is to much for a quickstart).

PR #222, opened 2018-03-13
Fix wrong indexing in Indicator.toDouble()

izeye:
- Fix wrong indexing in `Indicator.toDouble()`.
- Add a test for it and AssertJ and Mockito for easy tests.

### DIFF
--- a/ta4j-core/pom.xml
+++ b/ta4j-core/pom.xml
@@ -38,17 +38,5 @@
             <version>3.17</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.assertj</groupId>
-            <artifactId>assertj-core</artifactId>
-            <version>3.9.1</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.mockito</groupId>
-            <artifactId>mockito-all</artifactId>
-            <version>1.10.19</version>
-            <scope>test</scope>
-        </dependency>
     </dependencies>
 </project>
--- a/ta4j-core/pom.xml
+++ b/ta4j-core/pom.xml
@@ -38,5 +38,17 @@
             <version>3.17</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>3.9.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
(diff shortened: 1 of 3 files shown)


ta4j-core/pom.xml
```diff
@@ -38,5 +38,17 @@
...
+            <artifactId>assertj-core</artifactId>
+            <version>3.9.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
+            <version>1.10.19</version>
+            <scope>test</scope>
+        </dependency>
```
team172011: Adding additional dependency without consultation will not be accepted. Furthermore they are not needed for this PR. Please use existing functions of `TestUtils` or existing `junit.junit` dependency

ta4j-core/src/main/java/org/ta4j/core/Indicator.java
```diff
@@ -53,21 +53,22 @@
...
 	 */
 	static Double[] toDouble(Indicator<Num> ref, int index, int barCount) {
 
 		Double[] all = new Double[barCount];
 
-		for (int i = Math.max(0, index - barCount + 1); i <= index; i++) {
-			Num number = ref.getValue(index);
+		int startIndex = Math.max(0, index - barCount + 1);
+		for (int i = 0; i < barCount; i++) {
+			Num number = ref.getValue(i + startIndex);
```
team172011, line 71: Can you explain, please? What did you fix?

ta4j-core/pom.xml
izeye: @team172011 Okay, I just thought they are okay alongside this PR as they are commonly used in Java land. I'll change not to use them.

ta4j-core/src/main/java/org/ta4j/core/Indicator.java
izeye: @team172011 If I understand this method correctly, it should return an array of values ranged between `index - barCount + 1` and `index` but as you can see it's collecting the value at `index` multiple times and indexing `all` is also wrong. The added test shows what I expect from this method. Am I right?

izeye:
@team172011 I removed the dependencies in 0aa61adce27fa455b976603b2863d2e879f81738 as you requested.

Anyway, I'd like to suggest to use the libraries.

team172011:
> Anyway, I'd like to suggest to use the libraries.

The libraries look good, but adding those dependency just because of one test class is disproportional. And at the moment there are other changes in progress that affect all test classes, so we could keep it lean. Maybe we could come back to this in further releases

PR #226, opened 2018-03-17
Remove AbstractCriterionTest(Function)

### LINKED ISSUE #214: Remove AbstractCriterionTest(Function)
When used, this constructor leaves the factory `null`.  `getCriterion()` throws NPE in those cases.  Remove that constructor and fix criteria to use `AbstractCriterionTest(CriterionFactory, Function)`.

MarijnKoesen:
Fixes #214 

- Remove AbstractCriterionTest(Function)
- 
- 

### DIFF
--- a/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AbstractCriterionTest.java
+++ b/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AbstractCriterionTest.java
@@ -45,19 +45,15 @@ public abstract class AbstractCriterionTest {
     public static List<Function<Number, Num>> function(){
         return Arrays.asList(BigDecimalNum::valueOf, DoubleNum::valueOf);
     }
+
     /**
      * Constructor.
      * 
      * @param factory CriterionFactory for building an AnalysisCriterion given
      *            parameters
      */
-    public AbstractCriterionTest(CriterionFactory factory, Function<Number, Num> numFunction) {
+    protected AbstractCriterionTest(CriterionFactory factory, Function<Number, Num> numFunction) {
         this.factory = factory;
-        this.numFunction = numFunction;
-    }
-
-    public AbstractCriterionTest(Function<Number, Num> numFunction){
-        this.factory = null;
         this.numFunction = numFunction;
     }
 
@@ -69,6 +65,15 @@ public abstract class AbstractCriterionTest {
      */
     public AnalysisCriterion getCriterion(Object... params) {
         return factory.getCriterion(params);
+    }
+
+    /**
+     * Generates an AnalysisCriterion without any parameters
+     *
+     * @return AnalysisCriterion
+     */
+    public AnalysisCriterion getCriterion() {
+        return factory.getCriterion();
     }
 
     public Num numOf(Number n){
--- a/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AbstractCriterionTest.java
+++ b/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AbstractCriterionTest.java
@@ -26,7 +26,6 @@ package org.ta4j.core.analysis.criteria;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.ta4j.core.AnalysisCriterion;
-import org.ta4j.core.CriterionFactory;
 import org.ta4j.core.num.BigDecimalNum;
 import org.ta4j.core.num.DoubleNum;
 import org.ta4j.core.num.Num;
@@ -38,7 +37,7 @@ import java.util.function.Function;
 @RunWith(Parameterized.class)
 public abstract class AbstractCriterionTest {
 
-    private final CriterionFactory factory;
+    private Class indicatorClass;
     protected final Function<Number, Num> numFunction;
 
     @Parameterized.Parameters(name = "Test Case: {index} (0=BigDecimalNum, 1=DoubleNum)")
@@ -49,11 +48,10 @@ public abstract class AbstractCriterionTest {
     /**
      * Constructor.
      * 
-     * @param factory CriterionFactory for building an AnalysisCriterion given
-     *            parameters
+     * @param indicatorClass The indicator class to test
      */
-    protected AbstractCriterionTest(CriterionFactory factory, Function<Number, Num> numFunction) {
-        this.factory = factory;
+    protected AbstractCriterionTest(Class indicatorClass, Function<Number, Num> numFunction) {
+        this.indicatorClass = indicatorClass;
         this.numFunction = numFunction;
     }
 
@@ -64,16 +62,12 @@ public abstract class AbstractCriterionTest {
      * @return AnalysisCriterion given parameters
      */
     public AnalysisCriterion getCriterion(Object... params) {
-        return factory.getCriterion(params);
-    }
-
-    /**
-     * Generates an AnalysisCriterion without any parameters
-     *
-     * @return AnalysisCriterion
-     */
-    public AnalysisCriterion getCriterion() {
-        return factory.getCriterion();
+        try {
+            return (AnalysisCriterion)indicatorClass.getConstructor().newInstance(params);
+        } catch (Exception e) {
+            throw new RuntimeException("Cannot instantiate " + indicatorClass.getSimpleName()
+                + " implement getCriterion in your test class instead.");
+        }
     }
 
     public Num numOf(Number n){
--- a/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AverageProfitCriterionTest.java
+++ b/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AverageProfitCriterionTest.java
@@ -37,7 +37,10 @@ public class AverageProfitCriterionTest extends AbstractCriterionTest{
     private MockTimeSeries series;
 
     public AverageProfitCriterionTest(Function<Number, Num> numFunction) {
-        super(numFunction);
+        super(
+            (params) -> new AverageProfitCriterion(),
+            numFunction
+        );
     }
 
     @Test
@@ -46,7 +49,7 @@ public class AverageProfitCriterionTest extends AbstractCriterionTest{
         TradingRecord tradingRecord = new BaseTradingRecord(
                 Order.buyAt(0, series), Order.sellAt(2, series),
                 Order.buyAt(3, series), Order.sellAt(5, series));
-        AnalysisCriterion averageProfit = new AverageProfitCriterion();
+        AnalysisCriterion averageProfit = getCriterion();
         assertNumEquals(1.0243, averageProfit.calculate(series, tradingRecord));
     }
 
@@ -64,14 +67,14 @@ public class AverageProfitCriterionTest extends AbstractCriterionTest{
         TradingRecord tradingRecord = new BaseTradingRecord(
                 Order.buyAt(0, series), Order.sellAt(1, series),
                 Order.buyAt(2, series), Order.sellAt(5, series));
-        AnalysisCriterion averageProfit = new AverageProfitCriterion();
+        AnalysisCriterion averageProfit = getCriterion();
         assertNumEquals(numOf(95d/100 * 70d/100).pow(numOf(1d / 6)), averageProfit.calculate(series, tradingRecord));
     }
 
     @Test
     public void calculateWithNoBarsShouldReturn1() {
         series = new MockTimeSeries(numFunction,100, 95, 100, 80, 85, 70);
-        AnalysisCriterion averageProfit = new AverageProfitCriterion();
+        AnalysisCriterion averageProfit = getCriterion();
         assertNumEquals(1, averageProfit.calculate(series, new BaseTradingRecord()));
     }
 
@@ -79,13 +82,13 @@ public class AverageProfitCriterionTest extends AbstractCriterionTest{
     public void calculateWithOneTrade() {
         series = new MockTimeSeries(numFunction,100, 105);
         Trade trade = new Trade(Order.buyAt(0, series), Order.sellAt(1, series));
-        AnalysisCriterion average = new AverageProfitCriterion();
+        AnalysisCriterion average = getCriterion();
         assertNumEquals(numOf(105d / 100).pow(numOf(0.5)), average.calculate(series, trade));
     }
 
     @Test
     public void betterThan() {
-        AnalysisCriterion criterion = new AverageProfitCriterion();
+        AnalysisCriterion criterion = getCriterion();
         assertTrue(criterion.betterThan(numOf(2.0), numOf(1.5)));
         assertFalse(criterion.betterThan(numOf(1.5), numOf(2.0)));
     }
--- a/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AverageProfitCriterionTest.java
+++ b/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AverageProfitCriterionTest.java
@@ -37,10 +37,7 @@ public class AverageProfitCriterionTest extends AbstractCriterionTest{
     private MockTimeSeries series;
 
     public AverageProfitCriterionTest(Function<Number, Num> numFunction) {
-        super(
-            (params) -> new AverageProfitCriterion(),
-            numFunction
-        );
+        super(AverageProfitCriterion.class, numFunction);
     }
 
     @Test
--- a/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
+++ b/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
@@ -38,15 +38,13 @@ public class LinearTransactionCostCriterionTest extends AbstractCriterionTest{
     private ExternalCriterionTest xls;
 
     public LinearTransactionCostCriterionTest(Function<Number, Num> numFunction) throws Exception {
-        super(
-            (params) -> new LinearTransactionCostCriterion(
-                (double) params[0],
-                (double) params[1],
-                (double) params[2]
-            ),
-            numFunction
-        );
+        super(LinearTransactionCostCriterion.class, numFunction);
         xls = new XLSCriterionTest(this.getClass(), "LTC.xls", 16, 6, numFunction);
+    }
+
+    @Override
+    public AnalysisCriterion getCriterion(Object... params) {
+        return new LinearTransactionCostCriterion((double)params[0], (double)params[1], (double)params[2]);
     }
 
     @Test
--- a/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
+++ b/ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
@@ -38,7 +38,14 @@ public class LinearTransactionCostCriterionTest extends AbstractCriterionTest{
     private ExternalCriterionTest xls;
 
     public LinearTransactionCostCriterionTest(Function<Number, Num> numFunction) throws Exception {
-        super((params) -> new LinearTransactionCostCriterion((double) params[0], (double) params[1], (double) params[2]),numFunction);
+        super(
+            (params) -> new LinearTransactionCostCriterion(
+                (double) params[0],
+                (double) params[1],
+                (double) params[2]
+            ),
+            numFunction
+        );
         xls = new XLSCriterionTest(this.getClass(), "LTC.xls", 16, 6, numFunction);
     }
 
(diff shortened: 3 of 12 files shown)


edlins:
Thanks!! I'll review this today or tomorrow.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AbstractCriterionTest.java
```diff
@@ -71,6 +67,15 @@ public AnalysisCriterion getCriterion(Object... params) {
...
 
+    /**
+     * Generates an AnalysisCriterion without any parameters
+     *
+     * @return AnalysisCriterion
+     */
+    public AnalysisCriterion getCriterion() {
+        return factory.getCriterion();
+    }
+
```
edlins: This is not required.  The method that takes `Object... Params` can take `null` for `Params`.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AverageProfitCriterionTest.java
```diff
@@ -37,7 +37,10 @@
     private MockTimeSeries series;
 
     public AverageProfitCriterionTest(Function<Number, Num> numFunction) {
-        super(numFunction);
+        super(
+            (params) -> new AverageProfitCriterion(),
+            numFunction
+        );
```
edlins: This lambda that discards `params` doesn't feel right.  I was able to pass a method reference `AverageProfitCriterion::new` but only after defining an explicit constructor:
```
    public AverageProfitCriterion(Object... params) {
    }
```
But that seems even worse.  This is not a requested change, just an item for discussion.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
```diff
@@ -38,7 +38,14 @@
...
     public LinearTransactionCostCriterionTest(Function<Number, Num> numFunction) throws Exception {
-        super((params) -> new LinearTransactionCostCriterion((double) params[0], (double) params[1], (double) params[2]),numFunction);
+        super(
+            (params) -> new LinearTransactionCostCriterion(
+                (double) params[0],
+                (double) params[1],
+                (double) params[2]
+            ),
+            numFunction
+        );
```
edlins: Better.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/RewardRiskRatioCriterionTest.java
```diff
@@ -23,28 +23,23 @@
...
-        super(numFunction);
-    }
-
-    @Before
-    public void setUp() {
-        this.rrc = new RewardRiskRatioCriterion();
+        super(
+            (params) -> new RewardRiskRatioCriterion(),
+            numFunction
+        );
```
edlins: Better.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/RewardRiskRatioCriterionTest.java
```diff
@@ -61,7 +56,8 @@ public void rewardRiskRatioCriterion() {
         double peak = (105d / 100) * (100d / 95);
         double low = (105d / 100) * (90d / 95) * (80d / 95);
 
-        assertNumEquals(totalProfit / ((peak - low) / peak), rrc.calculate(series, tradingRecord));
+        AnalysisCriterion ratioCriterion = getCriterion();
```
edlins, line 56: Just a note - you could make the `ratioCriterion` an instance field and put the `getCriterion()` call in an `@Before` or in the constructor.  Not necessary, just a comment.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/VersusBuyAndHoldCriterionTest.java
```diff
@@ -37,7 +37,14 @@
...
     public VersusBuyAndHoldCriterionTest(Function<Number, Num> numFunction) {
-        super(numFunction);
+        super(
+            (params) ->
+                params.length == 0
+                    ? new VersusBuyAndHoldCriterion(new TotalProfitCriterion())
+                    : new VersusBuyAndHoldCriterion((AnalysisCriterion)params[0])
+            ,
+            numFunction
+        );
```
edlins: This doesn't look right to me.  I'd rather remove the conditional:
```
super(
    (params) -> new VersusBuyAndHoldCriterion((AnalysisCriterion) params[0]),
    numFunction);
 ```
And change the `getCriterion()` calls to `getCriterion(new TotalProfitCriterion())`.

File: ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AverageProfitCriterionTest.java (hunk 2 above)
Comment (MarijnKoesen) [reply ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AverageProfitCriterionTest.java]: Yeah, the solution I saw/see is just getting rid of the getCriterion altogether, or making it an abstract, although the latter solution will be almost the same in the end, as only 1 Crit requires params, all the others don't:

```java
    protected AverageProfitCriterionTest(Function<Number, Num> numFunction) {
        super(numFunction);
    }
    
    @Override
    protected AnalysisCriterion getCriterion(Object... params) {
        return new AverageProfitCriterion();
    }
```

or just plain:

```java
    protected AverageProfitCriterionTest(Function<Number, Num> numFunction) {
        super(numFunction);
    }

    protected AnalysisCriterion getCriterion() {
        return new AverageProfitCriterion();
    }
```

or even:

```java
    protected AverageProfitCriterionTest(Function<Number, Num> numFunction) {
        super(numFunction);
    }

    @Before
    protected void createCriterion() {
        criterion = new AverageProfitCriterion();
    }
```
As I'm not really sure what the idea behind the factory is.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/RewardRiskRatioCriterionTest.java
MarijnKoesen: Yeah, I think it's related to my comment above, what do you think about https://github.com/ta4j/ta4j/pull/226#discussion_r176988229 ?

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/VersusBuyAndHoldCriterionTest.java
MarijnKoesen: Done 👍

MarijnKoesen:
I've processed the feedback, still only some discussion stuff left, let's see what we want to do with the factory, keep it like this, or get rid of the factory altogether.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/AverageProfitCriterionTest.java
edlins: The factory was added so that the unit tests can call `getCriterion(Object... params)` consistently in all `Criterion` unit tests.  Also, when implementing a new `Criterion's` unit test you can select an appropriate "parent" unit test, copy it, rename it, and make it work with minimal editing.  Otherwise you'd have to refactor the name of the `Criterion`.  It's also easier to `diff` two different `Criteria` unit tests.  Finally we have the ability to develop a generic unit test framework that can work on all `Criteria`.

See [XLS Testing](https://ta4j.github.io/ta4j-wiki/XLS-Testing.html) and read the `Copy-Paste-Edit Quickstart` section.  Looks I may have to edit some of the formatting on that page..

MarijnKoesen: Ah I see, in practice though it doesn't really work out though, I diffed some of the test classes, but the tests are pretty different, the only thing that remains the same is `betterThan`, the rest is pretty different.

The lambda that disregards is the parameters is the result of having the 'abstract' getCriterion. I really had to look and see how it's working the first time I touched on of these tests. I think it's easier to understand if all the test classes are just like this:

```    
public class AverageProfitCriterionTest extends AbstractCriterionTest{
    private Analysis criterion;

    protected AverageProfitCriterionTest(Function<Number, Num> numFunction) {
        super(numFunction);
    }

    @Before
    protected void createCriterion() {
        criterion = new AverageProfitCriterion();
    }

    // ...
}
```

If all of them are like this, they will c/p exactly this from another one, and we'll get the same result in the end, consistent tests.

But i'm fine to let it rest, and let it be/merge it like this as well.

(Side note, still working on sqrt, mainly with making a good implementation for the initial guess, and trying to understand that algebra in the wiki :))

MarijnKoesen: Actually, I have another approach, see the last commit. It's up for discussion.

edlins: Okay, I'll take a look today.  If there's a better way, I'm all for it.

MarijnKoesen:
Yeah the idea is that we don't need to create a factory in every test class if it's just a plain constructor without parameters, when we can do it in the abstract already. (I wouldn't use the reflection approach in 'production' code, but in the tests I think it's fine, it's makes it simpler in this case).

We still get the benefit of just extending the `AbstractCriterionTest` with a pretty straight forward implementation if you c/p it, and only in the case of a special constructor do you need to specify the factory. No more need for implementing a factory with params that we don't need. Only if you need them, you'll need to specify the factory. Looks to be the best of both worlds.

edlins:
Okay, I see, in the case of `AnalysisCriterion` testing.  Before we commit to this approach make sure you look at `AbstractIndicatorTest` and consider how that uses a factory in a more useful manner.  That was actually the first factory implementation and it was basically copied to `AbstractCriterionTest`.  LMK if you think there's a problem with one using a factory and the other using a class (in the context of making unit test development as simple as possible for contributors).  Consistency is typically simpler to learn even if it just looks a little weird.  Good discussion.

MarijnKoesen:
Hmm... in the AbstractIndicatorTest there are actually 2 constructors possible `(Factory, NumFunction)` and `(NumFunction)`.

So there it's not enforced to use the `Factory`, and quite some tests also don't use it. 

For example in the `DoubleEMAIndicatorTest` it's not used. I think it's a simple straight forward test case now, simple and easy to understand. I actually like it.

A class that uses the factory for example is `EMAIndicatorTest` which does multiple calls to `getIndicator(closePrice, 1)`. Looking at it like this, I assume it's getting an EMAIndicator with the 2 parameters, it makes sense.
 
But to understand this I must take a look and mentally parse this single line (formatted exactly likes this) `super((data, params) -> new EMAIndicator(data, (int) params[0]), numFunction);` because of the single line, all the brackets and casting it's a bit hard to read and takes me some effort to see what is data, what is params, where are they coming from, what is numFunction etc. 
The first time I had to understand how this worked with the `LinearTransactionCostCriterionTest` it took me some time.

A simple `new EmaIndicator(data, closePRice, 1)` is just a little easier to understand for me / a newcomer to the codebase. But maybe this is just me.

In the end I can live with both solutions, although I have a slight personal preference for the simple way and not have the factory for readability :)

What do you think?

edlins:
> What do you think?

I'll mull this and give a response today.  @team172011 , any thoughts?

MarijnKoesen:
What shall we do? @edlins @team172011

edlins:
I'd just like to make sure we are consistent between `AbstractCriterionTest` and `AbstractIndicatorTest`.  So, if one switches from factory to class and reflection, then the other should too.  In the end, I just want to remove all calls to either class that only pass a `Function` in such a way that the `Criterion` or `Indicator` cannot be instantiated via `getIndicator` or `getCriterion`.

team172011:
Sorry guys for the late response. I was up to my ears with work and could not track this discussion so far. 

In a first look I can not detect any really improvements by using reflection instead of the factory approach (that is not really a [factory pattern](https://en.wikipedia.org/wiki/Factory_method_pattern)). Both ways are hard to understand. I also do not like the handling of the constructor parameter in reflection. I think in this approach the user does not even get IDE support by choosing type and order of the parameters.

How about replacing the factory interface by a java8 `Function` interface? It is not a factory and this naming could be very confusing to users, other than a `Function<Object, AnalysisCriterion>`. Using a function would be cleaner because it is more standardized and avoids an additional complex sounding factory interface in the ta4j lib.

Same for the ` IndicatorFactory<D ,I>`: the java8 library already provides a [`BiFunction<D,I,Indicator`](https://docs.oracle.com/javase/8/docs/api/java/util/function/BiFunction.html) interface that could be used here and that would make thinks a liitle bit clearer.

team172011:
You should also note that it could be easier to **not** using lambda statements for Criterions/Indicators with many parameters. You can also handle a "factory" implementing class to the constructor

In `ATRIndicatorTest`:

`super((data, params) -> new ATRIndicator( data, (int) params[0]),numFunction);`

would be the same as

`super(ATRTestFactory, numFunction);`

with implementation:

```
class ATRTestFactory implements IndicatorFactory<TimeSeries, Num>{
	@Override
	public Indicator<Num> getIndicator(TimeSeries data, Object... params) {
		return new ATRIndicator(data, (int)params[0]);
	}
 }
```

MarijnKoesen:
I agree that not using a lambda would be easier, again because it's just simpler, but In that case i'd probably just go for:

```
@RunWith(Parameterized.class)
public abstract class AbstractCriterionTest implements CriterionFactory {
    protected final Function<Number, Num> numFunction;

    @Parameterized.Parameters(name = "Test Case: {index} (0=BigDecimalNum, 1=DoubleNum)")
    public static List<Function<Number, Num>> function(){
        return Arrays.asList(BigDecimalNum::valueOf, DoubleNum::valueOf);
    }

    protected AbstractCriterionTest(Function<Number, Num> numFunction) {
        this.numFunction = numFunction;
    }

    public Num numOf(Number n){
        return numFunction.apply(n);
    }
}
```

So implementing the interface, but not the method, forcing it to the subclasses:

That way all tests must implement:

```
public AnalysisCriterion getCriterion(Object... params){}
```

Which will make it straight forward.

You guys agree to go this way?

edlins:
Sounds good to me, but I will defer on structural design decisions.  I might toss one more item in the mix (of course).  We have `AbstractIndicatorTest` and `AbstractCriterionTest`, which should be very similar.  How about an `AbstractCoreTest` for things like `NumTest`?  Right now it extends `AbstractIndicatorTest` but it's not an `Indicator`.

edlins:
> Could be an AbstractTest (for thinks like NumTest) and extending classes like AbstractCriterionTest and AbstractIndicatorTest

Sounds good to me.  I'll leave the details for the design team but happy to give feedback and help review changes.  :grin:

MarijnKoesen:
I did an attempt into making this, but i'm giving up. I really don't see forcing a generate interface into specific implementations....

Take a look at FisherIndicator:

```java
    /**
     * Constructor.
     *
     * @param series the series
     */
    public FisherIndicator(TimeSeries series) {
        this(new MedianPriceIndicator(series), 10);
    }

    /**
     * Constructor (with alpha 0.33, beta 0.67, gamma 0.5, delta 0.5).
     *
     * @param price the price indicator (usually {@link MedianPriceIndicator})
     * @param barCount the time frame (usually 10)
     */
    public FisherIndicator(Indicator<Num> price, int barCount) {
        this(price, barCount, 0.33, 0.67, ZERO_DOT_FIVE, ZERO_DOT_FIVE, 1, true);
    }
```

There we already have 2 data types, `Indicator<Num>` and `TimeSeries`, how can you now make an AbstractIndicator like this:

```java
public abstract class AbstractIndicatorTest<D, I> extends AbstractCoreTest implements IndicatorFactory<D, I>
```

You'll need to implement the `IndicatorFactory` twice, once like `Indicator<Num, TimeSeries`, and once like `Indicator<Num>, Indicator<Num>`. But afaik this is not possible in the abstract, especially because most of them have only 1 data parameter.

Then some actually have 2 Data:

```java
    public DifferenceIndicator(Indicator<Num> first, Indicator<Num> second) {
```

Further more there are multiple indicators that have the same number of parameters, but different types.

Take Fisher again:

```java
    public FisherIndicator(Indicator<Num> price, int barCount, double alpha, double beta)
    public FisherIndicator(Indicator<Num> ref, int barCount, double densityFactor, boolean isPriceIndicator)
```

So how will this generic `getIndicator(Object data...)` work?

A simple case can be like this:

```java    @Test
    public void withSma1AndSma2() {
        Indicator<Num> acceleration = getIndicator(series, 1, 2);

        assertNumEquals(0, acceleration.getValue(0));
        assertNumEquals(0, acceleration.getValue(1));
        assertNumEquals(0, acceleration.getValue(2));
        assertNumEquals(0, acceleration.getValue(3));
        assertNumEquals(0, acceleration.getValue(4));
    }

    @Test
    public void withSmaDefault() {
        Indicator<Num> acceleration = getIndicator(series);

        assertNumEquals(0, acceleration.getValue(0));
        assertNumEquals(0, acceleration.getValue(1));
        assertNumEquals(0, acceleration.getValue(2));
        assertNumEquals(0, acceleration.getValue(3));
        assertNumEquals(0, acceleration.getValue(4));
    }

    @Override
    public Indicator<Num> getIndicator(TimeSeries data, Object... params) {
        switch (params.length) {
            case 2:
                return new AccelerationDecelerationIndicator(data, (int) params[0], (int) params[1]);

            case 0:
                return new AccelerationDecelerationIndicator(data);

            default:
                throw new RuntimeException("Only 0 or 2 parameters are supported");
        }
    }
```

But how do you do it for the fisher? It gets soo ugly... the only thing I can think of to do it in a better way is reflection, but then we get that discussion again, which makes sense, because that is also not a nice solution.

```java
    @Override
    public Indicator<Num> getIndicator(TimeSeries data, Object... params) {
       // this gets sooo ugly...
        if (params.length == 4 && params[3].getClass().isAssignableFrom(Double.class) || (params[3].getClass().isPrimitive() && params[3].getClass().getTypeName().equals("double"))) {
            // create FisherIndicator(Indicator<Num> price, int barCount, double alpha, double beta)
        } else if (params.length == 4 && params[3].getClass().isAssignableFrom(Boolean.class) || (params[3].getClass().isPrimitive()
        params[3].getClass().getTypeName().equals("boolean")){
            // craete FisherIndicator(Indicator<Num> ref, int barCount, double densityFactor, boolean isPriceIndicator)
        }
        
        // and now make all this logic for the following:
        public 
            public FisherIndicator(TimeSeries series) {
            this(new MedianPriceIndicator(series), 10);
        }

    public FisherIndicator(Indicator < Num > price, int barCount){
            this(price, barCount, 0.33, 0.67, ZERO_DOT_FIVE, ZERO_DOT_FIVE, 1, true);
        }
    public FisherIndicator(Indicator < Num > price, int barCount, double alpha, double beta){
            this(price, barCount, alpha, beta, ZERO_DOT_FIVE, ZERO_DOT_FIVE, 1, true);
        }
    public FisherIndicator(Indicator < Num > price, int barCount, double alpha, double beta, double gamma, double delta)
        {
            this(price, barCount, alpha, beta, gamma, delta, 1, true);
        }
    public FisherIndicator(Indicator < Num > ref, int barCount, boolean isPriceIndicator){
            this(ref, barCount, 0.33, 0.67, ZERO_DOT_FIVE, ZERO_DOT_FIVE, 1, isPriceIndicator);
        }
    public FisherIndicator(Indicator < Num > ref, int barCount, double densityFactor, boolean isPriceIndicator){
            this(ref, barCount, 0.33, 0.67, ZERO_DOT_FIVE, ZERO_DOT_FIVE, densityFactor, isPriceIndicator);
        }
    public FisherIndicator(Indicator < Num > ref, int barCount, final double alphaD, final double betaD,
        final double gammaD, final double deltaD, double densityFactorD, boolean isPriceIndicator){
        }
    }
```

In the end I just think it's a bad design that there is `Object data...` thing that we force upon a whole set of different parameters which is just something that will not result in something good.

There should be (and there is) an interface: `Indicator<T>` which defines a couple of methods an Indicator should implement. That way you can implement it anyway you want as long as you adhere to the interface / implement these methods. The way you get there is to the implementer. Because of that we have the flexibility that some indicators take on TimeSeries, other another indicator, or two, whatever, that doesn't matter. 

Now we are kind of forcing the way you should get there, which is not ending up in a flexible and extendable solution imho.

I think that whole IndicatorFactory should be ditched, and the tests should just create the objects themselves.

The AbstractCoreTest is fine and can be like this:

```
@RunWith(Parameterized.class)
public abstract class AbstractCoreTest {
    public final Function<Number, Num> numFunction;

    @Parameterized.Parameters(name = "Test Case: {index} (0=BigDecimalNum, 1=DoubleNum)")
    public static List<Function<Number, Num>> function() {
        return Arrays.asList(BigDecimalNum::valueOf, DoubleNum::valueOf);
    }

    /**
     * Constructor<p/>
     *
     * @param numFunction the function to convert a Number into a Num implementation (automatically insertet by Junit)
     */
    public AbstractCoreTest(Function<Number, Num> numFunction) {
        this.numFunction = numFunction;
    }

    protected Num numOf(Number n) {
        return numFunction.apply(n);
    }
}
```

But leave the instantiation of the Indicator/Criterion to the test classes themselves.

edlins:
I'll circle back on the original issue here tomorrow and see how to whittle the scope back down to the original intent.  This is good discussion but there should probably be a separate issue to discuss potential design changes as I think that's gone OOS for #214 .

team172011:
replaced by #269

Issue #223, opened 2018-03-15
Enhance Num pow(Num)


edlins:
per https://github.com/ta4j/ta4j/pull/212#discussion_r173636408.  Also add test cases for overflow conditions.  @MarijnKoesen.

MarijnKoesen:
I agree, we should have those.

Furthermore I think we should first make a decision on whether to replace bigdecimalnum with precisionnum or not. Otherwise any work on BDN is a waste of time.

team172011:
Yes we should replace it. The differences are too little to hold two seperate implementations with BigDecimal as delegate

MarijnKoesen:
Let's first finalize that  PR then, and make the proper pow and sqrt implementations with all fixes and tests for PrecisionNum :+1:

edlins:
FYI - I'm updating #198 to replace `BigDecimalNum`.

edlins:
Disregard my previous comment.  This can wait or move forward with BDN.  There is value in moving forward with BDN as the logic is relatively easily copied to PN.  But it can wait as well.  Do leave this issue open regardless.

team172011:
#228

PR #228, opened 2018-03-18
Enhance/precision num

edlins:
Fixes #198 .

- PrecisionNum introduction
- add Num numOf(String, int)
- more tests in NumTest
- add TestUtils.assertIndicatorEquals() with delta for inexact matching
- logging
- rename DoubleNumVsBigDecimalNum to CompareNumTypes, add error percentage reporting


edlins:
Please hold, one or two small details to work out..

edlins:
Okay, I think this is good now.  To see how the rubber meets the road, take a look first at `PrecisionNumTest`.  Also, try running the `CompareNumTypes` example to see differences in time and accuracy as impacted by different precisions.

team172011:
@edlins do you want to replace BigDecimal with PrecisionNum within this PR?

edlins:
> do you want to replace BigDecimal with PrecisionNum within this PR?

I'd prefer not to because this is already pretty big, and the replacement is also bigger than I thought and touches a lot of files in ways I was not expecting.  As soon as this is merged, I am happy to perform the replacement.

edlins:
I can work on the new Conflicts today or tomorrow.

team172011:
> I'd prefer not to because this is already pretty big, and the replacement is also bigger than I thought and touches a lot of files in ways I was not expecting. As soon as this is merged, I am happy to perform the replacement.

okay

edlins:
That rebase was a little messy.  Plz double-check my manual merging.

edlins:
Fixed! (I think)

edlins:
Just wondering if anyone has had a chance to review this, please.

team172011:
> Just wondering if anyone has had a chance to review this, please.
I will do, asap

ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
```diff
@@ -230,7 +232,11 @@ public void addTrade(Num tradeVolume, Num tradePrice) {
     }
 
     @Override
-    public void addPrice(Num price) {
+    public void addPrice(Num inPrice) {
+        Num price = inPrice;
+        if (numFunction != null) {
+            price = numFunction.apply(inPrice.getDelegate());
```
team172011: The new instance variable is only needed for this statement? If `numFunction != null` and the user tries to add another `Num ` implementation the `inPrice` will be converted to the `Num ` implementation of the `BaseBar` class. Is this the correct behaviour? The user can get unexpected rounding deviatations. I think we should not implicit transform different Num implementations, but throw a `ClassCastException ` if the user tries to mix up different `Num ` implementations. There are also default functions in `Bar` interface that support adding and transforming a `Number` or `String` into the corresponding `Num` implementation of the `BaseBar` class:
```
    default void addPrice(String price, Function<Number, Num> numFunction){
        addPrice(numFunction.apply(new BigDecimal(price)));
    }

    default void addPrice(Number price, Function<Number, Num> numFunction){
        addPrice(numFunction.apply(price));
}
```

ta4j-core/src/test/java/org/ta4j/core/TestUtils.java
```diff
@@ -157,4 +160,38 @@ public static void assertNumNotEquals(double expected, Num actual) {
...
+                String expString = exp.toString();
+                String actString = act.toString();
+                int minLen = Math.min(expString.length(), actString.length());
+                if (expString.length() > minLen)
+                    expString = expString.substring(0, minLen) + "..";
+                if (actString.length() > minLen)
+                    actString = actString.substring(0, minLen) + "..";
+                throw new AssertionError(String.format("Failed at index %s: expected %s but actual was %s",
+                        i, expString, actString));
+            } else {
```
team172011: I think if the result is correct we do not have to log a message each time?

edlins: They are trace logs that would not normally be shown.  They are properly parametrized so performance impact is negligible.  I found them useful for examining exactly where and by how much the `Indicators` matched and deviated.  But I'll remove them for being excessively verbose.

ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
edlins: If `numFunction` is not stored and used then what is the purpose of the `numFunction` parameter in the constructor?
```
public BaseBar(Duration timePeriod, ZonedDateTime endTime, Function<Number, Num> numFunction)
```

edlins:
> Looks good, I only have some minor requests

Thanks!  Responses added.

ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
team172011: You need the `numFunction` parameter to initialize some values:

```
this.volume = numFunction.apply(0);
this.amount = numFunction.apply(0);
```

edlins: When I remove the change to `BaseBar` the following unit test in `PrecisionNumTest` (line 186) fails
```
assertEquals(32, ((BigDecimal) bigDecimalSeries.getBar(0).getClosePrice().getDelegate()).precision());
```
with
```
java.lang.AssertionError: expected:<32> but was:<120>
```

edlins:
LMK if I should remove or hack the failing unit test.

team172011:
> LMK if I should remove or hack the failing unit test.
Why does it fail? 

Failed tests:   precisionNumTest(org.ta4j.core.num.PrecisionNumTest): expected:<32> but was:<120>
Tests run: 933, Failures: 1, Errors: 0, Skipped: 0

edlins:
> Why does it fail?

Because `BaseBar.AddPrice(Num)` doesn't use a `numFunction`.  It just overwrites the `Num` fields with the `price` parameter.  It should use a `numFunction`. This could probably be done implicitly with
```
closePrice = closePrice.numOf(price)
```
This would need to be done on all four assignments (open, close, max, min).

edlins:
> I think managing the Num logic in TimeSeries is the best way

Okay, I'll look at changing my test class to use `TimeSeries` instead of `Bar`. Do you want me to deprecate `Bar#addPrice` or just leave that alone for now?

edlins:
Found a `BaseBar` constructor that actually uses `numFunction.apply()` on `String` parameters.  It's hardcoded to a one-day `Duration` but that's how I wrote the tests originally so I lucked out.

team172011:
> Do you want me to deprecate Bar#addPrice or just leave that alone for now?
We could also delete the addPrice function from Bar interface, but that would be much work in all exiting test cases. I have to think about it

edlins:
> We could also delete the addPrice function from Bar interface, but that would be much work in all exiting test cases. I have to think about it

Let's leave that out of scope for this PR.  I just added more test cases and a low precision (0.001) test.  The low precision was much less precise than `DoubleNum` and ran in 300% of the time.  I like `DoubleNum`.  I also added `assertIndiatorNotEquals(Indicator, Indicator, Num)` for negative testing.  I should probably add some test cases to `TestUtilsTest` but the tests in `PrecisionNumTest` cover it pretty well for now.

edlins:
FYI - this also implements #223 so that can be closed.

PR #233, opened 2018-03-19
Polish

izeye:
This PR fixes some typos and polishes trivial stuff.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/TimeSeriesManager.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/TimeSeriesManager.java
@@ -98,7 +98,7 @@ public class TimeSeriesManager {
     /**
      * Runs the provided strategy over the managed series.
      * <p>
-     * Opens the trades with the specified {@code orderType} order.
+     * Opens the trades with the specified {@link OrderType orderType} order.
      * @param strategy the trading strategy
      * @param orderType the {@link OrderType} used to open the trades
      * @return the trading record coming from the run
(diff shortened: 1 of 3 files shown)


ta4j-core/src/main/java/org/ta4j/core/TimeSeriesManager.java
```diff
@@ -97,7 +98,7 @@ public TradingRecord run(Strategy strategy, int startIndex, int finishIndex) {
     /**
      * Runs the provided strategy over the managed series.
      * <p>
-     * Opens the trades with {@link OrderType} BUY orders.
+     * Opens the trades with the specified {@code orderType} order.
```
edlins: @team172011 what's the preferred method of embedding `@link` and `@code`?  From the Oracle [Comments for Javadoc](http://www.oracle.com/technetwork/articles/java/index-137868.html):
```
Use in-line links economically
You are encouraged to add links for API names (listed immediately above) using the {@link} tag. It is not
necessary to add links for all API names in a doc comment. Because links call attention to themselves
(by their color and underline in HTML, and by their length in source code doc comments), it can make 
the comments more difficult to read if used profusely. We therefore recommend adding a link to an API 
name if:

    The user might actually want to click on it for more information (in your judgment), and
    Only for the first occurrence of each API name in the doc comment (don't bother repeating a link)

Our audience is advanced (not novice) programmers, so it is generally not necessary to link to API 
in the java.lang package (such as String), or other API you feel would be well-known.
```

edlins: My question is OOS and I approve this PR.  But I'd like to address the Javadoc comments across the entire project in a separate PR.  And this is a convenient place to discuss that.

team172011: I think we should use `@link` annontation like oracle suggests and the `@code` annontation optionally forl for code examples

izeye: @edlins @team172011 Thanks for the feedback!

I just tried to point the parameter `orderType` via `@code` and didn't intended to remove `@link`. I restored them in 147bcdb.

PR #238, opened 2018-03-21
Use stream API for TotalProfitCriterion.calculate()

izeye:
This PR changes to use stream API for `TotalProfitCriterion.calculate()`.


team172011:
looks good, thanks!

PR #250, opened 2018-04-05
added trailing stop loss rule

Chr1st0ph:

    add a trailing stop-loss rule ( similar to StopLossRule )
    added an entry to the unreleased section of CHANGES.md

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/Trade.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/Trade.java
@@ -102,7 +102,8 @@ public class Trade implements Serializable {
     public boolean equals(Object obj) {
         if (obj instanceof Trade) {
             Trade t = (Trade) obj;
-            return entry.equals(t.getEntry()) && exit.equals(t.getExit());
+            return (entry == null ? t.getEntry() == null : entry.equals(t.getEntry()))
+            		&& (exit == null ? t.getExit() == null : exit.equals(t.getExit()));
         }
         return false;
     }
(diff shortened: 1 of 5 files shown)


Chr1st0ph:
I still found a flaw if this rule is combined with other rules. I am working on a fix

Chr1st0ph:
I added another commit to handle combined rules. Please review

ta4j-core/src/main/java/org/ta4j/core/Trade.java
```diff
@@ -102,7 +102,8 @@ public Order getExit() {
     public boolean equals(Object obj) {
         if (obj instanceof Trade) {
             Trade t = (Trade) obj;
-            return entry.equals(t.getEntry()) && exit.equals(t.getExit());
+            return (entry == null ? t.getEntry() == null : entry.equals(t.getEntry()))
+            		&& (exit == null ? t.getExit() == null : exit.equals(t.getExit()));
         }
         return false;
     }
```
edlins, line 109: Should this throw a `ClassCastException` if not `instanceof Trade`?  Also would be nice to see at least one new test case in the appropriate test class for any change to a pervasive core class.

team172011: Test case would be appropriate. If obj is not instance of Trade the equals function returns false, that is standard behaviour of equals function

Chr1st0ph: equals-method-policy is defined in java.lang.Object#equals. I added some test-cases...

team172011:
Thanks! Resolved conflicts and merged local with [ef4c2d4](https://github.com/ta4j/ta4j/commit/ef4c2d4bca750e9aecdb6a4c4ee1b91126f2fb69)

PR #251, opened 2018-04-05
bug fix:  ATR(1) was built as ATR(chopTimeFrame)

jamestomk:
Fixes #.

- bug fix:  ATR(timeframe=1) was built as ATR(timeframe=chopTimeFrame)

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/ChopIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/ChopIndicator.java
@@ -45,7 +45,7 @@ public class ChopIndicator extends CachedIndicator<Num> {
 	 */
 	public ChopIndicator( TimeSeries timeseries, int ciTimeFrame, int scaleTo ) {
 		super( timeseries );
-        this.atrIndicator = new ATRIndicator( timeseries, ciTimeFrame );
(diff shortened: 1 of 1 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/ChopIndicator.java
```diff
@@ -45,7 +45,7 @@
 	 */
 	public ChopIndicator( TimeSeries timeseries, int ciTimeFrame, int scaleTo ) {
 		super( timeseries );
-        this.atrIndicator = new ATRIndicator( timeseries, ciTimeFrame );
+        this.atrIndicator = new ATRIndicator( timeseries, 1 );	// ATR(1) = Average True Range (Period of 1)
         hvi = new HighestValueIndicator( new MaxPriceIndicator(timeseries), ciTimeFrame );
         lvi = new LowestValueIndicator( new MinPriceIndicator(timeseries), ciTimeFrame );
         this.timeFrame = ciTimeFrame;
```
edlins, line 51: Is ATR(1) *always* used?  If so, this is okay.  If not, it should be added as a parameter.  Also, the test class should include cases that *must* be modified to pass with this change.  If it does not, they should be added.

jamestomk: ATR(1) is always used.  It is part of the definition of CI.  So a constant is fine.

team172011:
thanks!

PR #265, opened 2018-05-08
Set default num type

### LINKED ISSUE #247: Set default Num type
I think it would be nice to have a method of setting the default `Num` type throughout an application with a single call and then not have to use `SeriesBuilder.withNumTypeOf(Function)`.  That way a developer could quickly and easily change all of the `Num` types in an application with a one line code change (or a String resource or whatever).  I'm thinking about this as in-scope for the replacement of `BigDecimalNum` with `PrecisionNum`.

edlins:
Fixes #247 .

- add setDefaultFunction() to BaseTimeSeries.SeriesBuilder
- update BuildTimeSeries example app to use it
- 


petoncle:
@edlins The changes you made looks fine to me, even though I don't agree with some initial design decisions about Num in TimeSeries (like: TimeSeries being a Num factory).

PR #270, opened 2018-05-14
Remove time series

### LINKED ISSUE #242: Redundant TimeSeries setting
This seem to be redundant and written **in lots of Indicators**. Should we leave this as is?
```
public class ClosePriceIndicator extends CachedIndicator<Num> {

    private TimeSeries series;

    public ClosePriceIndicator(TimeSeries series) {
        super(series);
        this.series = series;
    }
```
**I can imagine two possibilities:**
1st - Here we would use the `getTimeSeries()` method inside the class instead of `series` reference.
```
public class ClosePriceIndicator extends CachedIndicator<Num> {

    public ClosePriceIndicator(TimeSeries series) {
        super(series);
    }
```
2nd - We wouldn't use the parent's constructor to set the TimeSeries.
```
public class ClosePriceIndicator extends CachedIndicator<Num> {

    private TimeSeries series;

    public ClosePriceIndicator(TimeSeries series) {
        this.series = series;
    }
```

edlins:
ChopIndicator: various simple fixes
StochasticRSIIndicator: comments and param names
ConvergenceDivergenceIndicator: remove unused method
ChopIndicatorTest: spelling, TODO

Fixes #242 .

- 
- 
- 

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/StochasticRSIIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/StochasticRSIIndicator.java
@@ -42,7 +42,8 @@ public class StochasticRSIIndicator extends CachedIndicator<Num> {
     private final HighestValueIndicator maxRsi;
 
     /**
-     * Constructor.
+     * Constructor.  In most cases, this should be used to avoid confusion
+     * over what Indicator parameters should be used.
      * @param series the series
      * @param barCount the time frame
      */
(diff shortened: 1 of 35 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/StochasticRSIIndicator.java
```diff
@@ -52,23 +53,23 @@ public StochasticRSIIndicator(TimeSeries series, int barCount) {
 
     /**
      * Constructor.
-     * @param indicator the indicator
+     * @param closePriceIndicator the close prices
      * @param barCount the time frame
      */
-    public StochasticRSIIndicator(Indicator<Num> indicator, int barCount) {
-        this(new RSIIndicator(indicator, barCount), barCount);
+    public StochasticRSIIndicator(Indicator<Num> closePriceIndicator, int barCount) {
```
team172011: Why renaming to CosePriceIndicator? StochRSI could also be based on High, Low, Open or so

edlins: > Why renaming to CosePriceIndicator? StochRSI could also be based on High, Low, Open or so

But in practice it never is, and the ambiguity has generated confusion that has spawned multiple Issues.  And this is a semantic change so you can still use any `Indicator`.  Users need guidance when they see `Indicator<Num> indicator` and `@param indicator the Indicator` as a parameter or they will use it wrong and open Issues.  However, I have moved the hint to the javadoc comment to get the PR accepted.

Issue #217, opened 2018-03-08
refactor Math.pow and Math.sqrt to Num methods


edlins:
After those changes are merged.

MarijnKoesen:
I'm working on this

PR #276, opened 2018-05-26
remove doubles

### LINKED ISSUE #271: Replace all doubles with Nums
in all indicator and criteria calculation.  Agree or disagree?

edlins:
Fixes #271 #217 .

- <*>Indicator: store `double` params as `Nums`, remove `double` math
- <*>Indicator: replace `Math.sqrt()` with `Num.sqrt()`
- ConvergenceDivergenceIndicator: spelling
- PeriodicalGrowthRateIndicator: utilize ONE, return value of `Num`
- IIIIndicator: var name fix
- PrecisionNum: fix sqrt()

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
@@ -25,6 +25,8 @@ package org.ta4j.core.num;
 import java.math.BigDecimal;
 import java.math.MathContext;
 import java.math.RoundingMode;
+import java.text.DecimalFormat;
+import java.text.ParseException;
 import java.util.Objects;
 import java.util.function.Function;
 
@@ -268,7 +270,7 @@ public final class PrecisionNum implements Num {
             return NaN;
 
         case 0:
-            return BigDecimalNum.valueOf(0);
+            return PrecisionNum.valueOf(0);
         }
 
         // Direct implementation of the example in:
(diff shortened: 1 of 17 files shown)


edlins:
My Eclipse formatter is breaking indentation in some files.  I'll correct this prior to merge, but other comments are welcome at this time.

ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
```diff
@@ -290,7 +292,14 @@ public Num sqrt(int precision) {
...
             String estimateString = String.format("%sE%s", estimatedMantissa, estimatedExponent);
             log.trace("x[0] =~ sqrt({}...*10^{}) =~ {}", mantissa, exponent, estimateString);
-            estimate = new BigDecimal(estimateString);
+            DecimalFormat format = new DecimalFormat();
+            format.setParseBigDecimal(true);
+            try {
+                estimate = (BigDecimal) format.parse(estimateString);
+            } catch (ParseException e) {
+                // TODO Auto-generated catch block
+                e.printStackTrace();
```
laviua, line 301: It is not good idea to print exception to the console during precision calculation.
will be better to wrap it into kind of RuntimeException.
in this case we can come when estimate stay still as a previous value

edlins: `DecimalFormat.parse(String)` throws `Exception`.  Do you mean `try/catch` and `throw RuntimeException(e.getMessage())`?

team172011: Yes, no imaginable way to handle this should throw RuntimeException("Could not format Number", e),

petoncle: Technically there is no user input that could lead to this exception being thrown, correct? If that's correct then:
When I read the message "Could not format Number" I might think that it is due to an invalid user input, but in this case such exception could only be due to a bug in the `sqrt(int)` implementation. So we can either:
- make the message more explicit to make the user aware that he found a bug in `sqrt(int)`
- or if we're confident enough that there's no condition under which an exception can be thrown, swallow the exception

laviua: 1. throw RuntimeException("Unable to parse into the DecimalFormat the following string: " + estimateString, e);
actually we don't care for original ParseException, so it can be even without ",e"
2. static code analyzer gives bad mark for this approach :)

edlins: Happy to go any way with this.  Just be clear on the preferred solution and I'll make the change.  @team172011 makes the final call.

team172011: I would preffer using a more explicit message like @laviua suggests because I cannot exculding any potential fault. If there will be any errors the programm should stop and not calculate wrong results whoes cause is not transparent.

PR #277, opened 2018-05-30
Revert "remove doubles"

team172011:
Reverts ta4j/ta4j#276


team172011:
I reverted the merge commit, because I overlooked open reviews. Unfortunatley they have been deleted durign the merge. There was an interesting question regarding printing an exception I think?

team172011:
replaced by #280

PR #280, opened 2018-06-06
Removedoubles

### LINKED ISSUE #271: Replace all doubles with Nums

edlins:
Fixes #271 #217 #276 .

- same as #276 but with new RuntimeException in PrecisionNum.sqrt

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterion.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterion.java
@@ -127,7 +127,6 @@ public class LinearTransactionCostCriterion extends AbstractAnalysisCriterion {
      * @return the absolute total cost of all orders in the trade
      */
     private Num getTradeCost(TimeSeries series, Trade trade, Num initialAmount) {
-        //Num totalTradeCost = SeriesBuilder.getDefaultFunction().apply(0);
         Num totalTradeCost = series.numOf(0);
         if (trade != null) {
             if (trade.getEntry() != null) {
--- a/ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterion.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterion.java
@@ -23,10 +23,13 @@
  *******************************************************************************/
 package org.ta4j.core.analysis.criteria;
 
+import java.util.function.Function;
+
 import org.ta4j.core.Order;
 import org.ta4j.core.TimeSeries;
 import org.ta4j.core.Trade;
 import org.ta4j.core.TradingRecord;
+import org.ta4j.core.BaseTimeSeries.SeriesBuilder;
 import org.ta4j.core.num.Num;
 
 /**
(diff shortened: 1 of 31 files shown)


edlins:
I fully expect there to be discussion regarding the changes to LTCC.  It feels messy and probably begs for some redesign.

team172011:
Does this replace #276  ?

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterion.java
```diff
@@ -124,7 +127,7 @@ private Num getOrderCost(Order order, Num tradedAmount) {
      * @return the absolute total cost of all orders in the trade
      */
     private Num getTradeCost(TimeSeries series, Trade trade, Num initialAmount) {
-        Num totalTradeCost = series.numOf(0);
+        Num totalTradeCost = SeriesBuilder.getDefaultFunction().apply(0);
```
team172011: So LTCC calculates only with the defaultFunction of `SeriesBuilder`? What if the `numFunction ` of `TimeSeries ` parameter is different? My first thought is to drop the Constructor initialization and only provide static `LTCC #calculate `functions with additional parameter `a`, `b` and `initialAmount` (here we have a time series with corresponding `numFunction`)

ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
```diff
@@ -297,8 +297,7 @@ public Num sqrt(int precision) {
             try {
                 estimate = (BigDecimal) format.parse(estimateString);
             } catch (ParseException e) {
-                // TODO Auto-generated catch block
-                e.printStackTrace();
+                throw new RuntimeException("Unable to parse into DecimalFormat the following string:" + estimateString + ": " + e);
```
team172011, line 300: looks good

edlins:
> Does this replace #276 ?

#276 was already merged.  Wasn't sure if you wanted something new or not.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterion.java
edlins: I don't remember now why I made this change.  I think it had to do with a `ClassCastException` but when I undo that one line the tests run fine.  So I've changed it back.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/LinearTransactionCostCriterionTest.java
```diff
@@ -38,7 +38,7 @@
     private ExternalCriterionTest xls;
 
     public LinearTransactionCostCriterionTest(Function<Number, Num> numFunction) throws Exception {
-        super((params) -> new LinearTransactionCostCriterion((double) params[0], (double) params[1], (double) params[2]),numFunction);
+        super((params) -> new LinearTransactionCostCriterion((double) params[0], (double) params[1], (double) params[2], numFunction), numFunction);
```
edlins, line 41: This is starting to feel hideous.  May need a redesign on how `numFunctions` are handled.  They are now
1) stored as an API default
2) stored in TimeSeries
3) passed explicitly

Maybe 2) and 3) should be removed and only 1) used..?  Regardless it feels like it's getting unwieldy.

PR #297, opened 2018-07-31
add update lastet bar method in time series

belerweb:
Some exchange provide continuous new bar data in the time period (eg. 1s in 1m Duration). So add update the latest bar method to support real time time series.

According to issue #149 and PR #181, the latest bar indicator is uncached, so update latest bar should not cause issues.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/BaseTimeSeries.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BaseTimeSeries.java
@@ -373,6 +373,15 @@ public class BaseTimeSeries implements TimeSeries {
     }
 
     @Override
+    public void updateBar(Bar bar) {
+        if (bars.isEmpty()) {
+            addBar(bar);
+        } else {
+            bars.set(bars.size() - 1, bar);
+        }
+    }
+
+    @Override
     public void addTrade(Number price, Number amount) {
         addTrade(numOf(price), numOf(amount));
     }
--- a/ta4j-core/src/main/java/org/ta4j/core/BaseTimeSeries.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BaseTimeSeries.java
@@ -319,13 +319,17 @@ public class BaseTimeSeries implements TimeSeries {
      * @apiNote to add bar data directly use #addBar(Duration, ZonedDateTime, Num, Num, Num, Num, Num)
      */
     @Override
-    public void addBar(Bar bar) {
+    public void addBar(Bar bar, boolean replace) {
         Objects.requireNonNull(bar);
         if(!checkBar(bar)){
             throw new IllegalArgumentException(String.format("Cannot add Bar with data type: %s to series with data" +
                     "type: %s",bar.getClosePrice().getClass(), numOf(1).getClass()));
         }
         if (!bars.isEmpty()) {
+            if (replace) {
+                bars.set(bars.size() - 1, bar);
+                return;
+            }
             final int lastBarIndex = bars.size() - 1;
             ZonedDateTime seriesEndTime = bars.get(lastBarIndex).getEndTime();
             if (!bar.getEndTime().isAfter(seriesEndTime)) {
@@ -370,15 +374,6 @@ public class BaseTimeSeries implements TimeSeries {
     public void addBar(Duration timePeriod, ZonedDateTime endTime, Num openPrice, Num highPrice, Num lowPrice,
                        Num closePrice, Num volume, Num amount) {
         this.addBar(new BaseBar(timePeriod, endTime, openPrice,highPrice,lowPrice,closePrice,volume, amount));
-    }
-
-    @Override
-    public void updateBar(Bar bar) {
-        if (bars.isEmpty()) {
-            addBar(bar);
-        } else {
-            bars.set(bars.size() - 1, bar);
-        }
     }
 
     @Override
--- a/ta4j-core/src/main/java/org/ta4j/core/TimeSeries.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/TimeSeries.java
@@ -148,7 +148,22 @@ public interface TimeSeries extends Serializable {
      * @see TimeSeries#setMaximumBarCount(int)
      * @apiNote use #addBar(Duration, ZonedDateTime, Num, Num, Num, Num, Num) to add bar data directly
      */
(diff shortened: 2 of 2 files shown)


team172011:
Hi @belerweb,
thank you for your contribution. I hesitated with this PR because I want to avoid letting the user generate a `Bar` instance. It is opaque, because you need a `numFunction` parameter that is the same as the internal `numFunction` of the `TimeSeries`. Maybe we could add a `boolean` parameter to the `addBar` functions, that replace the last bar if it is true?

belerweb:
@team172011 I have change to code. Please review it. Thanks.

dgosset:
Since your last commit on /develop, i had this classcast exception !

series.addTrade(BigDecimalNum.valueOf(volume),  BigDecimalNum.valueOf(bid));

volume & bid are BigDecimal

java.lang.ClassCastException: org.ta4j.core.num.BigDecimalNum cannot be cast to org.ta4j.core.num.PrecisionNum
	at org.ta4j.core.num.PrecisionNum.plus(PrecisionNum.java:144)
	at org.ta4j.core.BaseBar.addTrade(BaseBar.java:227)
	at org.ta4j.core.BaseTimeSeries.addTrade(BaseTimeSeries.java:391)

team172011:
@dgosset this must be caused by different `Num ` types. Afaics you try to mix `BigDecimalNum` with the `PrecisionNum` of the time series. Why not use `series.addTrade(Number, Number);` and let the series convert the `Number` (could be `Integer`, `Double`, `Long `etc.) into the coresponding `Num` type?

dgosset:
I am building with this issue from today. Yesterday this code was OK.
So it looks like a bug from previous commit on 0.12-SNAPSHOT (repo maven)

My local maven respo has many 0.12-SNAPSHOT  : 
This code works with ta4j-core-0.12-SNAPSHOT.jar from 01/05/2018 (ta4j-core-0.12-20180428.131229-6)
This code is KO with ta4j-core-0.12-SNAPSHOT.jar from 05/09/2018 (ta4j-core-0.12-20180905.103425-7)

I quickly tried a getDelegate() on my BigDecimal, which returns a Number to test an other way = doesn't work

team172011:
I don't know why your code did work yesterday, but as far as I see it is correct by throwing this exception. I have added the following **working** test case. Each test case is called twice (once with `PrecisionNum `and once with `DoubleNum `as num function):

```java
    /**
     * Tests if the {@link BaseTimeSeries#addTrade(Number, Number)} method works correct.
     */
    @Test
    public void addTradeTest() {
    	TimeSeries series = new BaseTimeSeries.SeriesBuilder().withNumTypeOf(numFunction).build();
    	series.addBar(new MockBar(ZonedDateTime.now(ZoneId.systemDefault()), 1d,numFunction));
    	series.addTrade(200, 11.5);
    	series.addTrade(BigDecimal.valueOf(200), BigDecimal.valueOf(11.5));
    }
```

If I add your line, this test case will **fail**:

```java
    /**
     * Tests if the {@link BaseTimeSeries#addTrade(Number, Number)} method works correct.
     */
    @Test
    public void addTradeTest() {
    	TimeSeries series = new BaseTimeSeries.SeriesBuilder().withNumTypeOf(numFunction).build();
    	series.addBar(new MockBar(ZonedDateTime.now(ZoneId.systemDefault()), 1d,numFunction));
    	series.addTrade(200, 11.5);
    	series.addTrade(BigDecimal.valueOf(200), BigDecimal.valueOf(11.5));
        // the following line will add BigDecimalNum to series with PrecisionNum or DoubleNum
        series.addTrade(BigDecimalNum.valueOf(200), BigDecimalNum.valueOf(11.5));
    }
```
Errors: 2 

Test Case 0:
-    java.lang.ClassCastException: org.ta4j.core.num.BigDecimalNum cannot be cast to org.ta4j.core.num.DoubleNum

Test Case 1:

-     java.lang.ClassCastException: org.ta4j.core.num.BigDecimalNum cannot be cast to org.ta4j.core.num.PrecisionNum

team172011:
@dgosset if you still think there is a bug, I would be greatful if you could open a new issue with working example code.

dgosset:
Hi @team172011, i will. For me this is a bug : 

I used BigDecimalNum which implements Num, and both class BaseTimeSeries & BaseBar have methods which are signed to accept classes which implements Num. Right?

https://github.com/ta4j/ta4j/blob/develop/ta4j-core/src/main/java/org/ta4j/core/BaseTimeSeries.java
391 public void addTrade(Num tradeVolume, Num tradePrice) {

https://github.com/ta4j/ta4j/blob/develop/ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
224 public void addTrade(Num tradeVolume, Num tradePrice) {

I don't have time to write a clean issue right now. I will stay on my previous snapshot (which works) for the run, and build Maven offline.

team172011:
@dgosset 

> and both class `BaseTimeSeries` & `BaseBar `have methods which are signed to accept classes which implements Num. Right?

Once you have choose a `Num` implementation (`PrecisionNum`, `DoubleNum`) by creating a `TimeSeries` you have to stay at your choosen implementation. It is not possible to add a `DoubleNum` value to a `TimeSeries` or `Bar` that uses `PrecisionNum`. The simple reason for this behaviour is, that converting one `Num` implementation to another `Num` implementation can cause significant rounding issuses.

Furthermore the `BigDecimalNum` implementation will be deleted in #297, because it acts the same like the `PrecisionNum` implementation (both use `BigDecimal `as delegate).

I don't know your code, but I would recommend to use a `TimeSeries` with `PrecisionNum`. You can create it like:

`TimeSeries series = new BaseTimeSeries.SeriesBuilder().withNumTypeOf(PrecisionNum.class).build();`

subsequently you can do:

`series.addTrade(PrecisionNum.valueOf(volume), PrecisionNum.valueOf(bid));`

dgosset:
I will try to do what you said on my next refactoring.

But, i used snapshot which means /develop and /develop still has BigDecimalNum.
I will follow the next release, but i think there is an issue right now for /develop (maybe fix at the next merge for the 0.12, when all the code will be clean and BigDecimalNum deleted)
I create an issue https://github.com/ta4j/ta4j/issues/313 if some others peoples have the same behavior.

I will check if i made this mistake to mix sometimes BigDecimalNum and PrecisionNum.

PR #301, opened 2018-08-19
Fix waitForRule's default satisfied value

kduhyun:
### Summary

I found that WaitForRule just satisfies when it meets a certain condition.
Because of that, I couldn't use the rule to buy like the following.

```
//This rule couldn't run because WaitForRule sets false at first.
Rule entrySignal = new AndRule(new WaitForRule(Order.OrderType.SELL, 2),
                    new UnderIndicatorRule(differencePercentageIndicator, -0.05));
```

So I fixed it.

Moreover, WaitForRule has a comment. 'No trading history, no need to wait.'
That means it has to return true. 

### Test
WaitForRuleTest.java

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/trading/rules/WaitForRule.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/trading/rules/WaitForRule.java
@@ -54,8 +54,8 @@ public class WaitForRule extends AbstractRule {
 
     @Override
     public boolean isSatisfied(int index, TradingRecord tradingRecord) {
-        boolean satisfied = false;
-        // No trading history, no need to wait
+        boolean satisfied = true;
+        // No trading history, no need to wait. Therefore it have to return true.
         if (tradingRecord != null) {
             Order lastOrder = tradingRecord.getLastOrder(orderType);
             if (lastOrder != null) {
--- a/ta4j-core/src/main/java/org/ta4j/core/trading/rules/WaitForRule.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/trading/rules/WaitForRule.java
@@ -55,7 +55,7 @@ public class WaitForRule extends AbstractRule {
     @Override
     public boolean isSatisfied(int index, TradingRecord tradingRecord) {
         boolean satisfied = true;
-        // No trading history, no need to wait. Therefore it have to return true.
+        // No trading history, no need to wait. Therefore it has to return true.
         if (tradingRecord != null) {
             Order lastOrder = tradingRecord.getLastOrder(orderType);
             if (lastOrder != null) {
--- a/ta4j-core/src/test/java/org/ta4j/core/trading/rules/WaitForRuleTest.java
+++ b/ta4j-core/src/test/java/org/ta4j/core/trading/rules/WaitForRuleTest.java
@@ -47,8 +47,8 @@ public class WaitForRuleTest {
         // Waits for 3 bars since last buy order
         rule = new WaitForRule(Order.OrderType.BUY, 3);
 
-        assertFalse(rule.isSatisfied(0, null));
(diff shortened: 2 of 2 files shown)


team172011:
> Moreover, WaitForRule has a comment. 'No trading history, no need to wait.'

Could also mean 'no need to apply this rule (at this index)'. In my opinion, if there is no `TradingRecord` or no `Order` of the determined `OrderType`, the rule should **not be satisfied**. Otherwise `isSatisfied(int)` and `isSatisfied(int, record)` methods will return always true at the beginning of a run.

ta4j-core/src/main/java/org/ta4j/core/trading/rules/WaitForRule.java
```diff
@@ -54,8 +54,8 @@ public WaitForRule(OrderType orderType, int numberOfBars) {
 
     @Override
     public boolean isSatisfied(int index, TradingRecord tradingRecord) {
-        boolean satisfied = false;
```
team172011, line 57: I think this was 'more convenient'. Please take a look at my comment under this PR

kduhyun:
Hi @team172011 ,
Thanks for the reply. Your assumption is also right, but if you see my entrySignal rule, it can not be run because of default value False.
What do you think about this case? And is there any example to support your assumption?

Thanks!

team172011:
Regarding your problem 

```
//This rule couldn't run because WaitForRule sets false at first.
Rule entrySignal = new AndRule(new WaitForRule(Order.OrderType.SELL, 2),
                    new UnderIndicatorRule(differencePercentageIndicator, -0.05));
```

Your `WaitForRule` means _wait two bars after a sell order has been executed_. If there is no sell order it is not satisfied, what I think is the correct behaviour.

kduhyun:
@team172011 , ok. I understand. 
That means I have to use WaitForRule after selling, not for entry signal. 
That makes sense.

But WaitForRule also has a problem.

If you see WaitForRuleTest.java:99, it doesn't meet the test case.
assertTrue(rule.isSatisfied(23, tradingRecord));

So I suggest the following.
```
        boolean satisfied = false;
        // No trading history, no need to wait. Therefore it has to return true.
        if (tradingRecord != null) {
            Order lastOrder = tradingRecord.getLastOrder(orderType);
            if (lastOrder != null) {
                int currentNumberOfBars = index - lastOrder.getIndex();
                satisfied = currentNumberOfBars >= numberOfBars;
            } else {
                satisfied = true;
            }
        }
        return satisfied;
    }
```

kduhyun:
difference is the following. Thanks.
```
 else {
                satisfied = true;
            }
```

team172011:
This would mean, that the rule is always satisfied if there is no `Order ` in `TradingRecord`. Why? Afaics there is another bug in the `WaitForRule` it ignores the `OrderType ` contructor parameter.

team172011:
Ah, I have overlooked the `getLastOrder(orderType)` call. I think the rule works correct...

team172011:
> but if you see my entrySignal rule, it can not be run because of default value False.

Yes, that is not because of wrong `WaitForRule `design. Your rule means "buy if we are two bars after a sell order and if the `UnderIndicatorRule ` is satisfied". You have to add another rule for at least the first entry/exit order, otherwise you will never get an order signal, because there is no sell order...

>  And is there any example to support your assumption

If you start backtesting your strategy, your `TradingRecord ` is generally empty. If `WaitForRule ` would always return true if the record is null or the record is empty, your strategy would always place orders at the beginning

kduhyun:
@team172011 , I understand what you mean.
I think WaitForRule's use case is specific, so my case is wrong with the rule.
Eventually, you noticed that WaitForRule doesn't care about the construct parameter.

You can close this Pull Request. Thanks!

team172011:
Your are welcome

PR #302, opened 2018-08-19
Feature for AndRule and OrRule to accept multiple rules

kduhyun:
### Summary
This feature is to accept multiple rules for AndRule and OrRule.
It will be useful to combine multiple rules like the following.

```
Rule entryRule = new OrRule(satisfiedRule, BooleanRule.TRUE, BooleanRule.TRUE);
Rule exitRule = new AndRule(satisfiedRule, BooleanRule.TRUE, BooleanRule.TRUE);
Rule complexRule = new OrRule(satisfiedRule.and(rule1, rule2, rule3), satisfiedRule2.and(rule1,rule2));
```

### Test
AndRuleTest
OrRuleTest

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/Rule.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/Rule.java
@@ -28,9 +28,12 @@ import org.ta4j.core.trading.rules.NotRule;
 import org.ta4j.core.trading.rules.OrRule;
 import org.ta4j.core.trading.rules.XorRule;
 
+import java.util.Arrays;
+import java.util.List;
+
 /**
  * A rule for strategy building.
- * </p>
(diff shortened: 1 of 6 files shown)


team172011:
@kduhyun thanks for your contribution to this repository. I am unsure if this `little` enhancement justifies such a big change. I like the present approach that enables to link rules without the use of arrays and the need of iterating

milczarekIT:
I agree with @team172011, old implementation is more cohesive

yurivin:
I also like old implementation. May be there is a way to just add new method and constructor to old implementation.

kduhyun:
Sorry. I was late.
Ok! As we agreed, I will close this PR. 

Thanks!

PR #319, opened 2018-09-15
Returning correct types for static methods valueOf

milczarekIT:
Fixes #.

- Corrected return types in method signature for DoubleNum.valueOf and PrecisionNum.valueOf
- Clean up code in PrecisionNumTest
- added tests for PrecisionNum.valueOf(Float.NaN), PrecisionNum.valueOf(Double.NaN)

- [Corrected return types in method signature for DoubleNum.valueOf and PrecisionNum.valueOf] added an entry to the unreleased section of `CHANGES.md`


milczarekIT:
In current implementation invoking: PrecisionNum.valueOf(Float.NaN) will throw NumberFormatException. From my point of view exception is correct, because also trying to create Nan as BigDecimal will throw the same exception. But current implementation suggest that NaN.NaN will be returned. This is not correct, because inside method valueOf you are using: 'val == Float.NaN'. It always return false

team172011:
already included in #323

PR #329, opened 2018-09-21
FeatureExtension Value At Risk, Expected Shortfall

damian-t:
- Added class to keep track of the returns of a strategy (log and arithmetic)
- Added Value At Risk as an analysis criterion
- Added Expected Shortfall as an analysis criterion

- [x ] added an entry to the unreleased section of `CHANGES.md`

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/analysis/Returns.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/analysis/Returns.java
@@ -1,0 +1,176 @@
+/*******************************************************************************
+ *   The MIT License (MIT)
+ *
+ *   Copyright (c) 2014-2017 Marc de Verdelhan, 2017-2018 Ta4j Organization 
+ *   & respective authors (see AUTHORS)
+ *
+ *   Permission is hereby granted, free of charge, to any person obtaining a copy of
+ *   this software and associated documentation files (the "Software"), to deal in
+ *   the Software without restriction, including without limitation the rights to
+ *   use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ *   the Software, and to permit persons to whom the Software is furnished to do so,
+ *   subject to the following conditions:
+ *
+ *   The above copyright notice and this permission notice shall be included in all
+ *   copies or substantial portions of the Software.
+ *
+ *   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ *   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ *   FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ *   COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ *   IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ *   CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *******************************************************************************/
+package org.ta4j.core.analysis;
+
+import org.ta4j.core.Indicator;
+import org.ta4j.core.TimeSeries;
+import org.ta4j.core.Trade;
+import org.ta4j.core.TradingRecord;
+import org.ta4j.core.num.DoubleNum;
+import org.ta4j.core.num.NaN;
+import org.ta4j.core.num.Num;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * The return rates.
+ * </p>
+ * This class allows to compute the return rate of a price time-series
+ */
+public class Returns implements Indicator<Num> {
+
+    public enum ReturnType {
+        LOG {
+            public Num calculate(Num x_new, Num x_old) {
+                // r_i = ln(P_i/P_(i-1))
+                return ((DoubleNum) x_new.dividedBy(x_old)).log();
+            }
+        },
+        ARITHMETIC {
+            @Override
+            public Num calculate(Num x_new, Num x_old) {
+                // r_i = P_i/P_(i-1) - 1
+                return x_new.dividedBy(x_old).minus(x_new.numOf(1));
+            }
+        };
+
+        /**
+         * @return calculate a single return rate
+         */
+        public abstract Num calculate(Num x_new, Num x_old);
+    }
+
+    private final ReturnType type;
+
+    /** The time series */
+    private final TimeSeries timeSeries;
+
+    /** The return rates */
+    private List<Num> values;
+
+    /**
+     * Constructor.
+     * @param timeSeries the time series
+     * @param trade a single trade
+     */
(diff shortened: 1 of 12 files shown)


ta4j-core/src/main/java/org/ta4j/core/analysis/Returns.java
```diff
@@ -0,0 +1,176 @@
...
+            public Num calculate(Num x_new, Num x_old) {
+                // r_i = ln(P_i/P_(i-1))
+                return ((DoubleNum) x_new.dividedBy(x_old)).log();
+            }
+        },
+        ARITHMETIC {
+            @Override
+            public Num calculate(Num x_new, Num x_old) {
+                // r_i = P_i/P_(i-1) - 1
+                return x_new.dividedBy(x_old).minus(x_new.numOf(1));
```
team172011: Performance would be better if you store `numOf(1)` as a class field instead of converting `int` to `num ` in every calculation

damian-t:
Included your remark of storing `numOf(1)` as a class variable. Also consistently compute the returns in Double precision (added another test for that)

team172011:
Why are your specifying `Double` precision? You could just take the `numFunction` of the `TimeSeries`, so that the precision of the underlying TimeSeries will be guaranteed.

damian-t:
> Why are your specifying `Double` precision? You could just take the `numFunction` of the `TimeSeries`, so that the precision of the underlying TimeSeries will be guaranteed.

The `TimeSeries` object is only given in the constructor of `Returns` but the ReturnType is part of the `Return` class definition where it doesn't know about `TimeSeries`. Is there a way to overcome this?

team172011:
>  Is there a way to overcome this?

At least for ARITHMETIC ReturnType you could write: 
`x_new.dividedBy(x_old).minus(one);`

And just specify for LOG the double precision

damian-t:
Thanks for the feedback, I updated the PR.
I also changed that the `Num` implementation of the LOG return is the same as for the corresponding timeseries, even though the computation is performed in `Double` precision.

damian-t:
@team172011 can this be merged?

ta4j-core/src/main/java/org/ta4j/core/analysis/Returns.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * The return rates.
+ * </p>
+ * This class allows to compute the return rate of a price time-series
+ */
+public class Returns implements Indicator<Num> {
```
milczarekIT, line 42: Why this indicator does not extend CachedIndicator?

ta4j-core/src/main/java/org/ta4j/core/analysis/Returns.java
```diff
@@ -0,0 +1,184 @@
...
+
+    public enum ReturnType {
+        LOG {
+            @Override
+            public Num calculate(Num x_new, Num x_old) {
+                // r_i = ln(P_i/P_(i-1))
+                return x_new.numOf(DoubleNum.valueOf(x_new.dividedBy(x_old).doubleValue()).log().doubleValue());
```
milczarekIT: Please use library convention name for variable, so variable should not have underscores

milczarekIT: Please do not use specific impl of Num: DoubleNum.valueOf should be removed

ta4j-core/src/main/java/org/ta4j/core/analysis/Returns.java
```diff
@@ -0,0 +1,184 @@
...
+            values.add(strategyReturn);
+        }
+    }
+
+    /**
+     * Calculates the returns for a trading record.
+     * @param tradingRecord the trading record
+     */
+    private void calculate(TradingRecord tradingRecord) {
+        for (Trade trade : tradingRecord.getTrades()) {
```
milczarekIT: We are using Java 8, so for loop can be replaced with foreach

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/ValueAtRiskCriterion.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public Num calculate(TimeSeries series, TradingRecord tradingRecord) {
+        Returns returns = new Returns(series, tradingRecord, Returns.ReturnType.LOG);
+        return calculateVaR(returns, confidence);
+    }
+
+    @Override
+    public Num calculate(TimeSeries series, Trade trade) {
+        if (trade != null && trade.getEntry() != null && trade.getExit() != null) {
```
milczarekIT: Instead of 
`trade.getEntry() != null && trade.getExit() != null`

please use:

`trade.isClosed()`

ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
```diff
@@ -168,6 +168,13 @@ public boolean isEqual(Num other) {
         return !other.isNaN() && delegate == ((DoubleNum) other).delegate;
     }
 
+    public Num log() {
```
milczarekIT, line 171: Impl should be provided also for `PrecisionNum`

ta4j-core/src/test/java/org/ta4j/core/mocks/MockTimeSeries.java
```diff
@@ -58,6 +62,15 @@ public MockTimeSeries(Function<Number, Num> nf) {
         super(arbitraryBars(nf));
     }
 
+    private static List<Bar> doublesToBars(Function<Number, Num> nf, List<Double> data) {
+        ArrayList<Bar> bars = new ArrayList<>();
+        for (int i = 0; i < data.size(); i++) {
+            //bars.add(new MockBar(ZonedDateTime.now().with(ChronoField.MILLI_OF_SECOND, i), data[i]));
```
milczarekIT: Please remove unnecessary comment

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/ValueAtRiskCriterion.java
```diff
@@ -0,0 +1,76 @@
...
+            var = returnRates.get(nInTail-1);
+
+            // VaR is non-positive
+            if (var.isGreaterThan(zero)) { var = zero; }
+        }
+        return var;
+    }
+
+    @Override
+    public boolean betterThan(Num criterionValue1, Num criterionValue2) {
```
milczarekIT, line 73: Greater ValueAtRisk is better? Hmm, I am not sure

ta4j-core/src/main/java/org/ta4j/core/analysis/Returns.java
damian-t: I tried to keep it consistent with the CashFlow class. That one implements Indicator<Num>

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/ValueAtRiskCriterion.java
damian-t: Losses are negative, hence VaR is negative. Better means less negative, meaning "greater is better". I wrote a comment in `calculateVaR()`. I will also add a comment to the `betterThan()` function.

team172011:
thanks!

PR #337, opened 2018-10-07
Removed unnecessary license in file

jansoren:
The lincence is here https://github.com/ta4j/ta4j/blob/develop/LICENSE and should be sufficient.


team172011:
Sorry, but I dont agree with removing the license header of all source files. Having a license header in each source code file is absolute standard

team172011:
Adding and checking licence header is part of current maven build cycle. At least every source file must include author and point to the licence file (if not mentioned).  This MIT license is not that big that it is disturbing the reader and it can be automatically managed by maven license plugin or eclipse license plugin. Just removing the license header is not a useful contribution.

jansoren:
@team172011 - With github you can see who have committed the code. I do agree that the IDE handles the header very vell but github does not, examples:
- https://github.com/ta4j/ta4j/blob/master/ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
- https://github.com/ta4j/ta4j/blob/master/ta4j-core/src/main/java/org/ta4j/core/BaseStrategy.java
- https://github.com/ta4j/ta4j/blob/master/ta4j-core/src/main/java/org/ta4j/core/TimeSeriesBuilder.java
- https://github.com/ta4j/ta4j/blob/master/ta4j-core/src/main/java/org/ta4j/core/TimeSeriesBuilder.java
- https://github.com/ta4j/ta4j/blob/master/ta4j-core/src/main/java/org/ta4j/core/TimeSeriesManager.java

PR #339, opened 2018-10-08
Fixed returning wrong profit value when starting type == SELL

milczarekIT:
Fixes #.
Fixed returning wrong profit value when starting type == SELL


team172011:
Thanks!

PR #340, opened 2018-10-08
Removed code smells in Num class

milczarekIT:
Removed code smells in Num class

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/num/Num.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/num/Num.java
@@ -35,7 +35,7 @@ import java.util.function.Function;
  * @see Num
  * @see Num#function()
  * @see DoubleNum
- * @See PrecisionNum
+ * @see PrecisionNum
 
  */
 public interface Num extends Comparable<Num>, Serializable {
@@ -204,9 +204,9 @@ public interface Num extends Comparable<Num>, Serializable {
     Num max(Num other);
 
     /**
-     * Retruns the {@link Function} to convert a number instance to
+     * Returns the {@link Function} to convert a number instance to
      * the corresponding Num instance
-     * @return
+     * @return function which convert a number instance to the corresponding Num instance
      */
     Function<Number, Num> function();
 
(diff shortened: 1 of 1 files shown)


ta4j-core/src/main/java/org/ta4j/core/num/Num.java
```diff
@@ -204,9 +204,9 @@
     Num max(Num other);
 
     /**
-     * Retruns the {@link Function} to convert a number instance to
+     * Returns the {@link Function} to convert a number instance to
      * the corresponding Num instance
-     * @return
+     * @return function which convert a number instance to the corresponding Num instance
```
team172011: converts

team172011: into

ta4j-core/src/main/java/org/ta4j/core/num/Num.java
```diff
@@ -260,18 +260,4 @@ default float floatValue(){
...
-    @Override
-    String toString();
-
-    /**
-     * {@inheritDoc}
-     * @apiNote: This method should return true if `this` and `obj` are both NaN.
-     */
-    @Override
-    boolean equals(Object obj);
-
```
team172011: I think forcing the implementing class to override the `hashCode()` method is necessary, because we also override the `equals()` method. It is also a consistent approach to have an explicit `toString()` method that converts the `Num` implementation to a human friendly string.

ta4j-core/src/main/java/org/ta4j/core/num/Num.java
milczarekIT: Fixed. Also in description

ta4j-core/src/main/java/org/ta4j/core/num/Num.java
milczarekIT: Ok, I've already reverted this change. Only one thing done here: "@apiNote: This method should return true if `this` and `obj` are both NaN." removed from javadoc, because it is not sufficient for this class

PR #350, opened 2018-10-15
Add hashCode method

tmoreira2020:


tmoreira2020:
1. NPE won't happen because there is treatment for that, take a closer look into the if statements.
2. maxPrice and minPrice fields still exist, the change were made only in the methods.

milczarekIT:
Sorry @tmoreira2020 - my fault, you are right.

Yes, maxPrice and minPrice fields still exist, but the question is why...

tmoreira2020:
Yeah, it is really strange that the interface changed but the implementation didn't. For sure the developer missed that.

team172011:
Thanks, @tmoreira2020! Yes, fields should be refactored, too.

PR #352, opened 2018-10-16
Minor fixes

tmoreira2020:


ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
```diff
@@ -376,8 +376,8 @@ public int hashCode() {
         result = prime * result + ((beginTime == null) ? 0 : beginTime.hashCode());
         result = prime * result + ((closePrice == null) ? 0 : closePrice.hashCode());
         result = prime * result + ((endTime == null) ? 0 : endTime.hashCode());
-        result = prime * result + ((maxPrice == null) ? 0 : maxPrice.hashCode());
-        result = prime * result + ((minPrice == null) ? 0 : minPrice.hashCode());
+        result = prime * result + ((highPrice == null) ? 0 : highPrice.hashCode());
+        result = prime * result + ((lowPrice == null) ? 0 : lowPrice.hashCode());
         result = prime * result + ((openPrice == null) ? 0 : openPrice.hashCode());
```
team172011: I have already changed min/max via hotfix commit in develop

ta4j-examples/src/main/java/ta4jexamples/indicators/IndicatorsToCsv.java
```diff
@@ -108,7 +109,7 @@ public static void main(String[] args) {
          */
         BufferedWriter writer = null;
         try {
-            writer = new BufferedWriter(new FileWriter("indicators.csv"));
+            writer = new BufferedWriter(new FileWriter(new File("target", "indicators.csv")));
```
team172011, line 112: why?

tmoreira2020: Because the `indicators.csv` is a generated file and by Maven's convention everything that is generated must be placed in the `target` directory, without this fix git complains that the file is not being tracked.

team172011: ah, I wondered myself sometimes about this behaviour...

Issue #327, opened 2018-09-20
Costs in Trade class


damian-t:
Hi guys
I am implementing a backtesting application using ta4j. One weakness as I see it, is the handling of trading costs in the model. As far as I can tell, a linear transaction costs model is implemented but only applicable as a analysis criterion (`LinearTransactionCostCriterion`).

Given the direct implication of trading costs (e.g. asset borrowing fees for short trades or transaction costs) on performance, it would be interesting to have them tied to the respective trades.

It can be "hacked" in by directly modifying the prices stored in the entry & exit orders of a trade but I think it would be cleaner to have cost within the `Trade` class that is updated whenever an order is added (enter and/or exit). It could incorporate holding & transaction costs.

What do you think?

milczarekIT:
I agree with you, this is missing for me too

damian-t:
I am working on a proposed implementation and will open a PR soon.

PR #362, opened 2018-11-02
Trading cost models

damian-t:
- Modelling of trading costs. Inclusion of two models
- LinearTransactionCostModel: Calculates the transaction cost as a linear function of the traded value (amount * price). Transaction costs arise at order level, hence they are part of the `Order` class.
- LinearBorrowingCostModel: Calculates holding costs (for borrowing assets) as a linear function of the traded amount and the holding period. Holding costs arise at trade level, hence they are part of the `Trade` class.
- Added a demonstrative example

Issue: https://github.com/ta4j/ta4j/issues/327

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 ### Changed
 
 ### Added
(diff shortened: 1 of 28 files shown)


team172011:
thanks!

PR #371, opened 2018-12-17
Construct BaseTradingRecord from List

joelmo:
I want to have the ability to choose List implementation of orders in BaseTradingRecord. The array constructor can be supported as well, tell me what you think.


team172011:
In my opinion there is to little added value in this big api change. You can also use the varargs constructor with lists:

`BaseTradingRecord(myOrdersList.asArray(myOrdersList.size()));`

PR #372, opened 2018-12-25
StopLossRule and StopGainRule now correctly handle sale orders. (issue  #373)

developmentMLPN:
From my point of view the StopLossRule and StopGainRule did not function correctly for Sale orders.
Both rules when calculating the threshold ratio where using the percentage with a minus operation which resulted in thresholds ratio like 0.97 for 3%. To make it work correctly for sale orders the value of the threshold ratio should be 1.03 for 3%.

- fixes for both rules
- tests to support the assumption and the fix


team172011:
thanks!

PR #394, opened 2019-03-23
feat: Adding LWMA indicator.

KealanLooby:
Fixes #.

- Adding Linearly Weighted Moving Average (LWMA) Indicator and unit test.

- [*] added an entry to the unreleased section of `CHANGES.md`

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/LWMAIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/LWMAIndicator.java
@@ -40,6 +40,9 @@ public class LWMAIndicator extends CachedIndicator<Num> {
 
     private final int barCount;
 
+    private final Num zero = numOf(0);
+
+
     public LWMAIndicator(Indicator<Num> indicator, int barCount) {
         super(indicator);
         this.indicator = indicator;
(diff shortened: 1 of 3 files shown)


KealanLooby:
Hello, 
I would like to add HeikinAshi and HeikinAshiSmoothed.
Just adding LWMA PR first, to follow with SMMAIndicator PR, any reviewers available ? 
thanks,
Kealan

ta4j-core/src/main/java/org/ta4j/core/indicators/LWMAIndicator.java
```diff
@@ -0,0 +1,72 @@
...
+    protected Num calculate(int index) {
+        Num sum = numOf(0);
+        Num denominator = numOf(0);
+        int count = 0;
+
+        if ((index + 1) < barCount) {
+            return numOf(0);
+        }
+
+        int startIndex = Math.max(0, index - barCount + 1);
```
team172011: no need for `Math#max` because index is `always >= barCount`

ta4j-core/src/main/java/org/ta4j/core/indicators/LWMAIndicator.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public LWMAIndicator(Indicator<Num> indicator, int barCount) {
+        super(indicator);
+        this.indicator = indicator;
+        this.barCount = barCount;
+    }
+
+    @Override
+    protected Num calculate(int index) {
+        Num sum = numOf(0);
```
team172011: it would be more effizient to use an instance field like `private final Num zero = numOf(0)` and reuse it, instead of creating a new one in every method call several times

ta4j-core/src/main/java/org/ta4j/core/indicators/LWMAIndicator.java
KealanLooby: Done.

ta4j-core/src/main/java/org/ta4j/core/indicators/LWMAIndicator.java
KealanLooby: Done, thanks.

PR #416, opened 2019-06-25
BaseBar: add constructor for BigDecimal

nimo23:

- add convenient constructor for BigDecimal.
- add javadoc for numFunction
- clean code for addPrice


ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
```diff
@@ -230,23 +272,19 @@ public void addTrade(Num tradeVolume, Num tradePrice) {
...
-            maxPrice = price;
-        }
-        if (minPrice == null) {
-            minPrice = price;
-        } else if(minPrice.isGreaterThan(price)){
-            minPrice = price;
-        }
-    }
+		closePrice = price;
+		if (maxPrice == null || maxPrice.isLessThan(price)) {
```
milczarekIT: Please sync your code with current ta4j develop branch. maxPrice and minPrice do not exists

ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
```diff
@@ -90,6 +92,47 @@ public BaseBar(ZonedDateTime endTime, double openPrice, double highPrice, double
...
+     * @param endTime the end time of the bar period
+     * @param openPrice the open price of the bar period
+     * @param highPrice the highest price of the bar period
+     * @param lowPrice the lowest price of the bar period
+     * @param closePrice the close price of the bar period
+     * @param volume the volume of the bar period
+     * @param numFunction the numbers precision
+     */
+    public BaseBar(Duration timePeriod, ZonedDateTime endTime, BigDecimal openPrice, BigDecimal highPrice, BigDecimal lowPrice, BigDecimal closePrice, BigDecimal volume, Function<Number, Num> numFunction) {
+    	this(timePeriod, endTime, openPrice, highPrice, lowPrice, closePrice, volume, numFunction.apply(0), numFunction);
```
team172011: Is it necessary to provide a `numFunction` parameter? Or can we use `PrecisionNum` here so that you do not have to create an own `Function<Number, Num>`?

nimo23: yes, good idea. I changed constructor so "PrecisionNum.valueOf()" is used for BigDecimal.

nimo23:
I removed restrictive constructors for double and string values with duration of 1 day. It should be allowed to put any duration and amount in any string, double, bigdecimal constructors. I also included defaults to use PrecisionNum for BigDecimal/String-Constructors and DoubleNum for double-constructor.

In my opinion, the following constructor could be removed, because it is redundant:

```
public BaseBar(ZonedDateTime endTime, double openPrice, double highPrice, double lowPrice, double closePrice, double volume, Function<Number, Num> numFunction) {
    	this(Duration.ofDays(1), endTime, openPrice, highPrice, lowPrice, closePrice, volume, 0.0, numFunction);
    }
```

milczarekIT:
Is still valid this pull request?
Some of us agreed (https://github.com/ta4j/ta4j/issues/419), we want to decline telescopic constructors of BaseBar. BaseBar is too long and it has too many constructors.

You can achieve your goal by using BaseBar.builder(Function<T, Num> conversionFunction, Class<T> clazz).
Example of usage you can find in ta4jexamples.backtesting.SimpleMovingAverageBacktest or ta4jexamples.timeSeries.BuildTimeSeries

nimo23:
@milczarekIT how can BaseBarBuilder replace the constructor of String values? 

BaseBarBuilder can be a good complement but if someone needs constructor for any other reasons (because they want to avoid Builder needing more memory/cpu cycles than constructor..)..?

nimo23:
I finished with BaseBar. I think, there is no more need to put any other constructors in BaseBar in future. For all other cases BaseBarBuilder should be used..

nimo23:
BaseBar is now finished. I added "trades" in constructor and removed all restricted or redundant constructors..For all other cases BaseBarBuilder should be used..

milczarekIT:
Hi @nimo23,
I didn't mention about `BaseBarBuilder `, I mentioned about `BaseBar.builder(Function<T, Num> conversionFunction, Class clazz)`, so I suggested to use `ConvertibleBaseBarBuilder`.

I agree that the library should allow for high flexibility and it is a reason why `BaseBarBuilder` and `ConvertibleBaseBarBuilder` were created.
Yes, you can use `ConvertibleBaseBarBuilder` with `String` and I give you an example of usage with Strings in my previous post. Please check once again: `ta4jexamples.timeSeries.BuildTimeSeries.barBuilderFromString` 

You can also use other types (e.g. your own types) with this bar builder.

milczarekIT:
BTW: "This branch has conflicts that must be resolved"

milczarekIT:
Build failed

nimo23:
There are a few test cases depending on custom constructors of BaseBar. I did not have time to adapt each test, at the moment I left these constructors within BaseBar..

nimo23:
Hi @milczarekIT 

Why the unused property of `Class<T> clazz` ?

```
public static <T> ConvertibleBaseBarBuilder<T> builder(Function<T, Num> conversionFunction, Class<T> clazz) {
    }
```

nimo23:
replaced by https://github.com/ta4j/ta4j/pull/428

PR #418, opened 2019-06-26
Bar aggregator

nimo23:
- add new methods for bar aggregation (upscaling): aggregateTimeSeries(), aggregateBars()

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,28 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
+---
+
+**I have checked existing issues and wiki**
+- [ ] I could not find similar [issues](https://github.com/ta4j/ta4j/issues?utf8=%E2%9C%93&q=)
+- [ ] I could not find a solution in the [wiki](https://ta4j.github.io/ta4j-wiki/) or [faq section](https://ta4j.github.io/ta4j-wiki/FAQ.html)
+
+**Describe the bug**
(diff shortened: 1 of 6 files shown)


nimo23:
Method allows to upscale a list of bars from lower timePeriod to higher timePeriod.

If you want, add more test cases or refine or optimise code.

decebals:
Another idea is to move all these static helper methods in a separate utils class (`TimeSeriesUtils` for example). In this mode the size of `BaseTimeSeries` class decrease considerable.

nimo23:
Yes, good idea. TimeSeriesUtils (or in future: BarSeriesUtils).

milczarekIT:
I agree with @decebals. These changes should not be included in `BaseTimeSeries`. It can be done in `TimeSeriesUtils `, but in my opinion, there can be a separate component responsible for bars aggregation e.g. `BarAggregator` without static methods.

Please notice that a current method signature looks like:
`public static List<Bar> aggregateBars(List<Bar> bars, Duration timePeriod)`

It has `Bar` in the signature so it is not specific for a BaseBar. What is more, please remember that it is a library so it should be flexible. In this case, it makes sense to have BarAggregator interface with one (for this moment) implementation, It can be easily extended by user in its application.

nimo23:
We can create a new package **"core/utils"** in which we can store such things: "core.utils.BarAggregator" with the provided static methods . Such things should not be within the root "core"..

@milczarekIT  I dont see any sense in providing an interface for BarAggregator. As you saw, the "Interface" is already build in by using the "Bar"-Interface within method signature. And I also dont see any sense in providing an object for BarAggregator..why one should create a constructor only to invoke a method?

nimo23:
https://github.com/ta4j/ta4j/pull/420

PR #422, opened 2019-07-03
Feature/base bar builder

milczarekIT:
- Added BaseBarBuilder 
- Added ConvertibleBaseBarBuilder 
- Example usage BaseBarBuilder and ConvertibleBaseBarBuilder  in ta4j-examples 


ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
```diff
@@ -257,12 +299,12 @@ public void addPrice(Num price) {
         closePrice = price;
         if (highPrice == null) {
```
nimo23: replace "if-else-if" with:

```
if (highPrice == null || highPrice.isLessThan(price)) {
highPrice = price;
}
if (lowPrice == null || lowPrice.isGreaterThan(price)) {
lowPrice = price;
}

```

ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
```diff
@@ -257,12 +299,12 @@ public void addPrice(Num price) {
...
             highPrice = price;
-        } else if(highPrice.isLessThan(price)) {
```
nimo23: Replace if-else-if with:

```
if (highPrice == null || highPrice.isLessThan(price)) {
highPrice = price;
}
if (lowPrice == null || lowPrice.isGreaterThan(price)) {
lowPrice = price;
}
```

milczarekIT: Due to existing pull request: https://github.com/ta4j/ta4j/pull/423/files, this comment is no longer valid. It will be fixed in #423

nimo23: oh ok, forget it;)

nimo23:
Hi @milczarekIT 

Why the **unused** property of `Class<T> clazz` ?

```
public static <T> ConvertibleBaseBarBuilder<T> builder(Function<T, Num> conversionFunction, Class<T> clazz) {
    }
```

PR #424, opened 2019-07-10
Clean PrecisionNum

nimo23:
- use log.error instead of printstacktrace..
- use "BigDecimal.valueOf(double)" instead of "new BigDecimal(double)"
- minor cleanings..

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
@@ -136,7 +136,7 @@ public final class PrecisionNum implements Num {
      * @return the {@code Num}
      */
     public static PrecisionNum valueOf(String val) {
-        if (val.toUpperCase().equals("NAN")) {
+        if (val.equalsIgnoreCase("NAN")) {
             throw new NumberFormatException();
         }
         return new PrecisionNum(val);
@@ -150,7 +150,7 @@ public final class PrecisionNum implements Num {
      * @return the {@code Num}
      */
     public static PrecisionNum valueOf(String val, int precision) {
-        if (val.toUpperCase().equals("NAN")) {
+        if (val.equalsIgnoreCase("NAN")) {
             throw new NumberFormatException();
         }
         return new PrecisionNum(val, precision);
--- a/ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
@@ -441,8 +441,7 @@ public final class PrecisionNum implements Num {
             try {
                 estimate = (BigDecimal) format.parse(estimateString);
             } catch (ParseException e) {
-                // TODO Auto-generated catch block
-                e.printStackTrace();
+                log.error("PrecicionNum ParseException:", e);
             }
         }
         BigDecimal delta;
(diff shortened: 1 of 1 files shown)


nimo23:
why a mathContext for floating poins in BigDecimal?

```
private PrecisionNum(float val) {
        mathContext = new MathContext(DEFAULT_PRECISION, RoundingMode.HALF_UP);
        delegate = new BigDecimal(val, mathContext);
    }

    private PrecisionNum(double val) {
        mathContext = new MathContext(DEFAULT_PRECISION, RoundingMode.HALF_UP);
        delegate = BigDecimal.valueOf(val);
    }
```

Should I replace this by:

```
private PrecisionNum(float val) {
        delegate = BigDecimal.valueOf(val);
    }
private PrecisionNum(double val) {
        delegate = BigDecimal.valueOf(val);
    }
```

nimo23:
How can I rewrite this

`estimateString = String.format("%1." + precision + "e", estimate);  (line 465)`

to something like:

`estimateString = String.format("%1.e", precision, estimate);  `

nimo23:
Why is MathContext used for float?

```
private PrecisionNum(float val) {
        mathContext = new MathContext(DEFAULT_PRECISION, RoundingMode.HALF_UP);
        delegate = new BigDecimal(val, mathContext); // BigDecimal.valueOf(val); ?
    }

    private PrecisionNum(double val) {
        mathContext = new MathContext(DEFAULT_PRECISION, RoundingMode.HALF_UP);
        delegate = BigDecimal.valueOf(val);
    }
```

milczarekIT:
> MathContext for float ok?

MathContext is for BigDecimal (output), float is an input

ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
```diff
@@ -744,7 +743,7 @@ public Num pow(Num n) {
         // use double pow(double, double)
         double xpowb = Math.pow(delegate.doubleValue(), bDouble);
         // use PrecisionNum.multiply(PrecisionNum)
-        BigDecimal result = xpowa.multiply(new BigDecimal(xpowb));
```
team172011, line 747: I am not sure if replacing `new` with `valueOf ` does make sense here. Did you compare the results or why are you changing this? Using new or valueOf do not the same and in this case I can imagine that cutting off digits (that is what the valueOf method does) is not the preffered behaviour.

ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
```diff
@@ -136,7 +136,7 @@ private PrecisionNum(BigDecimal val, int precision) {
      * @return the {@code Num}
      */
     public static PrecisionNum valueOf(String val) {
-        if (val.toUpperCase().equals("NAN")) {
```
team172011, line 139: Why do you want to change this? `toUpperCase `+ `equals` is not clearer than just one method call `equalsIgnoreCase`. Also the performance of `equalsIgnoreCase` is a little bit better.

nimo23:
@team172011 yes, you are right. this PR does not make any sense. There is no question that PrecicionNum can be improved..but this PR does not. So I close it.

ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
nimo23: Yes, you are right. Did not know that..

ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
nimo23: ? `val.toUpperCase().equals` was the code **before**..I wanted to replace to `equalsIgnoreCase`..

Issue #419, opened 2019-06-27
create a new class "BarBuilder"


nimo23:
Maybe we can add another class **BaseBarBuilder** which has builder methods to create different kind of Bars (coming from different Numbers, double, string, bigdecimal, precicionNum, doubleNum, Num). As this class is long it should not be embeeded within BaseBar but include in seperate class-file..

team172011:
Having a `BaseBarBuilder `could be an option. But so far I tried to not let the user create `Bar` objects on its own, because of the possible conflict with other Bars or time series that hold other `Num ` implementation. If possilbe the user should use `TimeSeries.addBar(..)` and the correct bar will be added to the

nimo23:
>But so far I tried to not let the user create Bar objects on its own

There are cases where bar(s) must be created before creating a barSeries or where bars and BarSeries must be loosely coupled..for example, to allow to build two barSeries with different nums but with same bars..so please dont try to restrict bar creation without barSeries..

Would be nice to have have a BaseBarBuilder which simplifies bar creation. Actually there are too many constructors with many parameters of BaseBar which tries to permutate any available options..boilerplate which can be eliminated by BaseBarBuilder. And it s cumbersome to oversee which constructor is actually used by BaseBar or series.addBar(,,,,).

decebals:
I don't like telescopic constructors but I admit that the signature of `TimeSeries.addBar` must be simple.
A possible solution is to keep the actual signature of `TimeSeries.addBar` method (not break the API) but inside (implementation) of method to use a `BarBuilder` and to remove the telescopic constructors from `BaseBar`.

milczarekIT:
> A possible solution is to keep the actual signature of `TimeSeries.addBar` method (not break the API) but inside (implementation) of method to use a `BarBuilder` and to remove the telescopic constructors from `BaseBar`.

Yes, in my opinion, it is the best option. It also can solve the issue that trades (count) in a bar can be also updated by `BaseBar.addTrade` method. Described here: https://github.com/ta4j/ta4j/issues/414

milczarekIT:
I created pull request with a proposal of BarBuilder: https://github.com/ta4j/ta4j/pull/422.

It also solves an issue with an update of 'trade' property in BaseBar described here: https://github.com/ta4j/ta4j/issues/414.

[...]

PR #428, opened 2019-07-11
BaseBar:  cleanup constructors

nimo23:
- removed all redundant constructors from BaseBar()
- adapt the used BaseBar constructors in some "src/test" and "ta4jexamples"
- added default constructors for double, BigDecimal, String
- added property "trades" in constructor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **BaseTimeSeriesBuilder** moved from inner class to own class
 
 ### Added
+- :tada: **Enhancement** Added common constructors in BaseBar for BigDecimal, Double and String values
+- :tada: **Enhancement** Added constructor in BaseBar with trades property
 - :tada: **Enhancement** Added BaseBarBuilder and ConvertibleBaseBarBuilder - BaseBar builder classes
 - :tada: **Enhancement** Added BarAggregator and TimeSeriesAggregator to allow aggregates bars and time series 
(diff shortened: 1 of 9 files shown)


nimo23:
All **test cases** and **ta4jexamples** which make use of a **BaseBar()**-constructor could be replaced by BaseBarBuilder in the future.

nimo23:
I added entries to changelog.

nimo23:
I did notice the `Class<T> clazz` property of the following method  in `BaseBar.java` is not used within method. Am I right? If so, I can remove it this property from the arguments list..

public static <T> ConvertibleBaseBarBuilder<T> builder(
    Function<T, Num> conversionFunction, 
    Class<T> clazz) 
{
}

milczarekIT:
Hi 
`Class<T> clazz` is not used, but it is mandatory to deduce by compiler which type/function should be  used. 
You can try to remove it, but then code will not compile.

milczarekIT:
> * added default constructors for double, BigDecimal, String
> * added property "trades" in constructor

In following discussion: https://github.com/ta4j/ta4j/issues/419 @team172011  and @decebals agreed that BarBuilder approach is better option. I agree with guys too.

Your PR duplicates features provided in https://github.com/ta4j/ta4j/pull/422, but you are using your approach with telescopic constructors.

What is more, you removed some constructor(s), so you introduced unncessary braking changes which are not compatible with version 0.12

nimo23:
I added an entry to the changelog to notice users that unneccessary constructors are removed. This PR is finished and can be merged.

PR #429, opened 2019-07-14
Update PrecisionNum.java

nimo23:
- replace "toUpperCase.equals" to "equalsIgnoreCase"
- replace printStackTrace to log.error

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/num/PrecisionNum.java
@@ -136,7 +136,7 @@ public final class PrecisionNum implements Num {
      * @return the {@code Num}
      */
     public static PrecisionNum valueOf(String val) {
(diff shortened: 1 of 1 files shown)


nimo23:
Look at line 465:


```
// + should be avoided in String.format
estimateString = String.format("%1." + precision + "e", estimate);  
```

How can I rewrite this to something like: 

`estimateString = String.format("%1.e", precision, estimate); `

milczarekIT:
> How can I rewrite the above to something like

`String.format` is not suitable for dynamic formatting

You can use `NumberFormat`:

`
NumberFormat formatter = NumberFormat.getInstance(); 
formatter.setMaximumFractionDigits(precision);
formatter.setMinimumFractionDigits(precision);
estimateString = formatter.format(estimate);
`

nimo23:
hmm..if so then we should replace all these `String.format() ` things to `NumberFormat` in a seperated PR because there are other code files with String.format()), too.

PR #431, opened 2019-08-04
Add Strategy object to the list of parameters of a TradeStatement constructor

tmoreira2020:
When you are running several backtest it is hard to map the results to a given Strategy without passing along the strategy object to TradingStatement

- [ ] Add Strategy object to the list of parameters of a TradeStatement constructor


tmoreira2020:
👍

PR #434, opened 2019-08-07
BarSeries verify

nimo23:
**Added methods to correct or detect corrupted, invalidated or incompleted barSeries:**

- TimeSeries#replaceBarIfChanged()
- TimeSeries#findMissingBars()
- Tests provided by new class TimeSeriesVerifyTest


nimo23:
build failed because of 

`Expected feature release number in range of 9 to 14, but got: 8`

Did I miss to see any java 9 code in my PR?

tmoreira2020:
The problem with the build is fixed here https://github.com/ta4j/ta4j/pull/436

nimo23:
yes, would be better to outsource this to a new class...TimeSeriesUtils with static methods?

nimo23:
I think, I will do a new class "`TimeSeriesUtils`" where we can put:
- such small things related to barSeries
- shortcuts (e.g. static methods for barAggregations which will eliminate boilerplate code to write) - - and other utilities not needed for a class/interface-design.

I think, if one static method one time deserves object/interface, we can refactor it anyway..

 If one thinks, static util for such cases is bad, please suggest.

PR #449, opened 2019-09-08
Handle NaN from CloseLocationValueIndicator as zero

### LINKED ISSUE #375: CloseLocationValueIndicator returns NaN if max price = min price
CloseLocationValueIndicator returns NaN if max price = min price, since it's dividing 0
max price could equal to min price for 1 min chart during quiet hours 
due to the cached nature of EMA, it means all subsequent values will all become NaN for indicators like ChaikinOscillatorIndicator

team172011:
Fixes #375.

- CloseLocationValueIndicator returns zero instead of NaN if high==low

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CloseLocationValueIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CloseLocationValueIndicator.java
@@ -36,6 +36,8 @@ import org.ta4j.core.num.Num;
  */
 public class CloseLocationValueIndicator extends CachedIndicator<Num> {
 
+    private final Num zero = numOf(0);
+
     public CloseLocationValueIndicator(TimeSeries series) {
         super(series);
     }
@@ -43,11 +45,12 @@ public class CloseLocationValueIndicator extends CachedIndicator<Num> {
     @Override
     protected Num calculate(int index) {
         final Bar bar = getTimeSeries().getBar(index);
-
-        final Num close = bar.getClosePrice();
         final Num low = bar.getLowPrice();
         final Num high = bar.getHighPrice();
+        final Num close = bar.getClosePrice();
 
-        return ((close.minus(low)).minus(high.minus(close))).dividedBy(high.minus(low));
+        final Num diffHighLow = high.minus(low);
+
+        return diffHighLow.isNaN() ? zero : ((close.minus(low)).minus(high.minus(close))).dividedBy(diffHighLow);
     }
 }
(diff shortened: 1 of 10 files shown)


team172011:
Maybe we should just fix the `CloseValueIndicators` returning zero instead of NaN?

nimo23:
I guess, returning zero for such cases is the right way:

- maxPrice>minPrice -> return > 0
- maxPrice<minPrice -> return < 0
- maxPrice=minPrice -> return 0

nimo23:
> Maybe we should just fix the CloseValueIndicators returning zero instead of NaN?

There is no `CloseValueIndicator`. You mean `CloseLocationValueIndicator`?

nimo23:
>returning zero instead of NaN

Yes.

Normally, CLV returns 0, only if the close is exactly halfway between high and low.
However, it is not wrong to return 0 if open=high=low=close.

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CloseLocationValueIndicator.java
```diff
@@ -40,6 +40,11 @@ public CloseLocationValueIndicator(TimeSeries series) {
         super(series);
     }
 
+    /**
+     *
+     * @param index the bar index
+     * @return the CLC value or {@link org.ta4j.core.num.NaN} if high == low
```
milczarekIT: should be 'the CL**V** value'

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CloseLocationValueIndicator.java
```diff
@@ -40,6 +40,11 @@ public CloseLocationValueIndicator(TimeSeries series) {
...
+     */
     @Override
     protected Num calculate(int index) {
```
milczarekIT, line 46: CLV oscillates around 0 (zero) in the range from -1 to +1.
NaN is not a correct value for this indicator. For this case when high == low, 0 should be returned.

It is exactly the same case when we have balanced bullish and bearish pressure. Neither Bulls nor Bears were able to push price neither up nor down from the middle level.

Please add missing test cases in CloseLocationValueIndicatorTest:
1) high == low
2) close - low == high - close (e.g. low: 100, close: 120, high: 140)

ta4j-core/src/test/java/org/ta4j/core/indicators/volume/ChaikinMoneyFlowIndicatorTest.java
```diff
@@ -33,12 +33,12 @@
 
 public class ChaikinMoneyFlowIndicatorTest {
 
+    ZonedDateTime now = ZonedDateTime.now();
```
milczarekIT: Access can be private

milczarekIT:
Hi
@nimo23  is right. 0 should be returned.

@team172011 - implementation is not correct. Please fix it. 
Also, I requested for missing tests:
high == low
close - low == high - close (e.g. low: 100, close: 120, high: 140)

In both cases, 0 should be returned.

PR #456, opened 2019-09-15
added code formatting to travis script

team172011:
- added mvn goal to travis script

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,4 @@ jdk:
   - openjdk8
 
 script:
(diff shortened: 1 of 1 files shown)


nimo23:
> mvn formatter:format cannot cause bugs within the source code, because it **only** formats the source code. Nothing else. 

>Users (contributors) must be aware that to submit code they need to follow some guidelines and format the code and have the license applied is one of them.

I dont think so. Such guidelines can, as you see, be automated on the server. So no need to add complexitiy to users..

PR #477, opened 2019-09-30
DoubleNum.class: changed isEqual()-method

nimo23:
- changed isEqual()-method of DoubleNum.class

There were the "==" in isEqual()-method. Was this made intentionally? I changed it to use `Double.compare() == 0`. Maybe we should be careful: `-0.0 != +0.0` and `NaN != NaN`, I dont know if this matters with `Double.compare()` ?

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
@@ -164,16 +164,16 @@ public class DoubleNum implements Num {
         return delegate <= 0;
     }
 
-    @Override
-    public boolean isEqual(Num other) {
-        return !other.isNaN() && delegate == ((DoubleNum) other).delegate;
-    }
-
     public Num log() {
         if (delegate <= 0) {
             return NaN;
         }
         return new DoubleNum(Math.log(delegate));
+    }
+    
+    @Override
+    public boolean isEqual(Num other) {
+        return !other.isNaN() && compareTo(other) == 0;
     }
 
     /**


nimo23:
Hi @milczarekIT 

>Please check comment from my review.

I can not see any code review.

As I said before, the new isEqual()-method sees `-0.0` as a different value than `+0.0`.

`// -0.0 != +0.0 is FALSE for this method`

which is why 

```
@ Test
public void test() {
final Num num1 = DoubleNum.valueOf(-0.0);
final Num num2 = DoubleNum.valueOf(0.0);

assertTrue(num2.isEqual(num1));
}
```

failed. I will include this test case into `DoubleNumTest.class`.

We could add this special case into the new isEqual()-method:

@Override
// -0.0 != +0.0 is FALSE for this method
// all NaN are equal, so NaN == NaN is always TRUE for this method
public boolean isEqual(Num other) {
        return !other.isNaN() && Double.doubleToLongBits(delegate) == Double.doubleToLongBits(((DoubleNum) other).delegate) ;
    }


Btw, how can the bits of `+0.0` be the same as `-0.0`? It would always be false for those cases.
I think we should normalize the double values in this special case within the isEqual()-method: Using x+0 converts -0.0 to 0.0. What do you think?

nimo23:
The previous equals()-method seems wrong to me, so I changed it.

`return !other.isNaN() && delegate == ((DoubleNum) other).delegate;`

team172011:
@nimo23  I think the old and your new equals method are both correct ways of comparision. It depends on the use case, what is the reason why it seemed wrong for you? 
With this change we would invalidate a lot of calculations that were correct so far because of 0 != -0

nimo23:
@team172011 yes, you are right. No need to change. However, we should merge the `DoubleNumTest.java` and ignore the rest.

PR #495, opened 2019-11-30
Fixed Ichimoku span A and span B calculations

### LINKED ISSUE #448: IchimokuChikouSpanIndicator calculation is wrong ?
IchimokuChikouSpanIndicator is today's price value plotted 26 period ago and not price 26 periods ago ? Has somebody used this indicator and please provide some explanation if possible.

Warm Regards,
Sudeep

jurepetrovic:
Fixes #448.

Fixed Ichimoku span A and span B calculations.
Fixed unit tests.
Added calculation spreadsheet file
IchimokuLineIndicator not abstract anymore, because needs instantiation


stoxalpha:
Hey @jurepetrovic  great work. Looked quickly at high level, but we would also need a easier way to get the current value of Span A and Span B which would be derived from 26 days and 52 days earlier values for today. That would be a very typical use case.

jurepetrovic:
I don't understand what do you mean. 
You get today's values by invoking the indicator on today's index. 
If you want to get future values for spanA and spanB you invoke it with future index ;)

I'd like to get that in main branch if possible. 
Who is responsible for that? @team172011  ? 
Please advise.

team172011:
thanks @jurepetrovic !

stoxalpha:
This is still pending add to master branch :(

stoxalpha:
Method threw 'java.lang.IndexOutOfBoundsException' exception.. Also we need to create a handle for this

victorchung0325:
Just read the fix and found below line of code in IchimokuSenkouSpanAIndicator.java and IchimokuSenkouSpanBIndicator.java

int spanIndex = index - offset + 1;

I wonder why need to add one for the spanIndex? why not just minus the index by offset?

PR #498, opened 2019-12-09
Added TradeOpenedMinimumBarCountRule - rule to specify minimum bar count for opened trade.

milczarekIT:
Fixes #.

- Added TradeOpenedMinimumBarCountRule - rule to specify minimum bar count for opened trade.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 ### Removed/Deprecated
 
 ### Added
-- :tada: **Enhancement** Added AtLeastBarCountRule - rule to specify minimum bar count for opened trade.
+- :tada: **Enhancement** Added TradeOpenedMinimumBarCountRule - rule to specify minimum bar count for opened trade.
 - :tada: **Enhancement** Added DateTimeIndicator a new Indicator for dates.
 - :tada: **Enhancement** Added DayOfWeekRule for specifying days of the week to trade.
 - :tada: **Enhancement** Added TimeRangeRule for trading within time ranges.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 ### Removed/Deprecated
 
 ### Added
+- :tada: **Enhancement** Added AtLeastBarCountRule - rule to specify minimum bar count for opened trade.
 - :tada: **Enhancement** Added DateTimeIndicator a new Indicator for dates.
 - :tada: **Enhancement** Added DayOfWeekRule for specifying days of the week to trade.
 - :tada: **Enhancement** Added TimeRangeRule for trading within time ranges.
--- a/ta4j-core/src/main/java/org/ta4j/core/trading/rules/AtLeastBarCountRule.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/trading/rules/AtLeastBarCountRule.java
@@ -1,0 +1,70 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2014-2017 Marc de Verdelhan, 2017-2019 Ta4j Organization & respective
+ * authors (see AUTHORS)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package org.ta4j.core.trading.rules;
+
+import org.ta4j.core.TradingRecord;
+
+/**
+ * Rule to specify minimum bar count for opened trade. Using this rule makes
+ * sense only for exit rule (for entry rule
+ * {@link AtLeastBarCountRule#isSatisfied(int, TradingRecord)} always return
+ * false.
+ */
+public class AtLeastBarCountRule extends AbstractRule {
+
+    /**
+     * Minimum bar count for opened trade.
+     */
+    private final int barCount;
+
+    public AtLeastBarCountRule(int barCount) {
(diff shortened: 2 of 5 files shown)


nimo23:
Good idea. 

>Using this rule makes sense only for exit rule (for entry rule {@link AtLeastBarCountRule#isSatisfied(int, TradingRecord)} always return false.

Why restrict this rule only for exit? It could also make sense for entry rule to provide a rule for **the minimum number of bars required between this entry and the prior entry**. Maybe think about renaming to `MinimumBarCountRule` or `MinBarCountRule` .

milczarekIT:
>  the minimum number of bars required between this entry and the prior entry.

Sounds interesting, but looks like the idea for the new rule (single responsibility principle).

My rule depends on opened trade and bar count for opened trade.

Your proposal depends on closed trade and bar count between the last exit.

After your suggestion, I added a more descriptive name, so `AtLeastBarCountRule` was renamed to `TradeOpenedMinimumBarCountRule`.

So your proposal can be `MinimumBarCountBetweenTradesRule` or sth similar.

nimo23:
>My rule depends on opened trade and bar count for opened trade.

Can you provide an use case?


>Your proposal depends on closed trade and bar count between the last exit.

Not exactly. For example: If `distanceMin = 5`, then the bar distance between **this entry and the prior entry** must be at least 3. Where `distanceMin` is the required minimum number of bars (minBarCount). With such a rule you can avoid too many excessive entries which can occur by whipsaws. Would be nice if you add such a rule also because it is similar to the logic of your provided rule.

nimo23:
>Your proposal depends on closed trade and bar count between the last exit.

Not exactly. For example: If `distanceMin = 5`, then the bar distance between **this entry and the prior entry** must be at least 5. Where `distanceMin` is the required minimum number of bars (**minBarCount**, therefore `MinBarCountRule`). With such a rule you can avoid too many excessive entries which can occur by whipsaws. Would be nice if you add such a rule also because it is similar to the logic of your provided rule.

**I recommend to use the following names or similar for the rules:**

- `MinBarsAfterOpenedTradeRule` or `MinBarCountAfterOpenedTradeRule` (this is your actual PR)
    _returns true if a minimum number of bars exist after the last opened trade_
- `MinBarsBetweenTradesRule` (this could be another PR)
    _returns true if a minimum number of bars are between the actual possible entry and the prior entry_


To think further, we can also provide a `Duration` property as a complement of `barCount` for these rules. With this, we can have something like "_exit trades after 5 hours_",  "_exit trades after 5 bars_" and similar. The two rules can either take a `barCount` or a `Duration` property. Both properties (`barCount` and `Duration`) represent **time**. So a better naming for these two rules could be:

- `TimeAfterOpenedTradeRule` (this is your actual PR)
    _returns true if a minimum number of bars or a minimum duration exist after the last opened trade_
- `TimeBetweenTradesRule` (this could be another PR)
    _returns true if a minimum number of bars or a minimum duration exist between the actual possible entry and the prior entry_

nimo23:
>Your proposal depends on closed trade and bar count between the last exit.

Not exactly. For example: If `distanceMin = 5`, then the bar distance between **this entry and the prior entry** must be at least 5. Where `distanceMin` is the required minimum number of bars (**minBarCount**, therefore `MinBarCountRule`). With such a rule you can, for example, avoid too many excessive entries which can occur by whipsaws. Would be nice if you add such a rule also because it is similar to the logic of your provided rule.

**I recommend to use the following names or similar for the rules:**

- `MinBarsAfterOpenedTradeRule` or `MinBarCountAfterOpenedTradeRule` (this is your actual PR)
    _returns true if a minimum number of bars exist after the last opened trade_
- `MinBarsBetweenTradesRule` (this could be another PR)
    _returns true if a minimum number of bars are between the actual possible entry and the prior entry_


To think further, we can also provide a `Duration` property as a complement of `barCount` for these rules. With this, we can have something like "_exit trades after 5 hours_",  "_exit trades after 5 bars_" and similar. The two rules can either take a `barCount` or a `Duration` property. You could derive the `barCount` by `Duration` property (barCount = Floor(Duration/timePeriod)) but it is maybe better to use the date of the bar and compare that with `Duration`, if given. So you have two different constructors in your rule: one with `barCount` and the other with `Duration`. Both properties (`barCount` and `Duration`) represent **time**. (Yes, `barCount` also represents time: `1 barCount = 1 x barDuration = 1 x bar.getTimePeriod()`.) So a better naming for these two rules could be:

- `TimeAfterEntryRule`(this is your actual PR)
    _returns true if a minimum number of bars or a minimum duration exist after the last entry_
- `TimeBetweenEntryRule`(this could be another PR)
    _returns true if a minimum number of bars or a minimum duration exist between the actual possible entry and the prior entry_

nimo23:
>Your proposal depends on closed trade and bar count between the last exit.

Not exactly. For example: If `distanceMin = 5`, then the bar distance between **this entry and the prior entry** must be at least 5. Where `distanceMin` is the required minimum number of bars (**minBarCount**, therefore `MinBarCountRule`). With such a rule you can, for example, avoid too many excessive entries which can occur by whipsaws. Would be nice if you add such a rule also because it is similar to the logic of your provided rule;

- `MinBarsAfterOpenedTradeRule` or `MinBarCountAfterOpenedTradeRule` (this is your actual PR)
    _returns true if a minimum number of bars exist after the last opened trade_
- `MinBarsBetweenTradesRule` (this could be another PR)
    _returns true if a minimum number of bars are between the actual possible entry and the prior entry_


To think further, we can also provide a `Duration` property as a complement of `barCount` for these rules. With this, we can have something like "_exit trades after 5 hours_",  "_exit trades after 5 bars_" and similar. The two rules can either take a `barCount` or a `Duration` property. You could derive the `barCount` by `Duration` property (`barCount = Floor(Duration/timePeriod)`) but it is maybe better to use the date of the bar and compare that with `Duration`, if given. 

So you have two different constructors in your rule: one with `barCount` and the other with `Duration`. Both properties (`barCount` and `Duration`) represent **time**. (Yes, `barCount` also represents time: `1 barCount = 1 x barDuration = 1 x bar.getTimePeriod()`.) So a better naming for these two rules could be:

**I recommend to use the following names or similar for the the 2 rules:**

- `TimeAfterEntryRule`(this is your actual PR)
    _returns true if a minimum number of bars or a minimum duration exist after the last entry_
- `TimeBetweenEntryRule`(this could be another PR)
    _returns true if a minimum number of bars or a minimum duration exist between the actual possible entry and the prior entry_

milczarekIT:
> Not exactly. For example: If `distanceMin = 5`, then the bar distance between **this entry and the prior entry** must be at least 5. Where `distanceMin` is the required minimum number of bars (**minBarCount**, therefore `MinBarCountRule`).

Ok, so it is an idea for the next rule. Please do not create one big rule for everything - please remember about Single Responsibility Principle. If you think it can be useful for ta4j users - feel free to provide impl.


> - MinBarsBetweenTradesRule (this could be another PR)
_returns true if a minimum number of bars are between the actual possible entry and the prior entry_

Please be precise... MinBarsBetweenTradesRule - between trades means between last exit and next entry, but in the description you proposed 'between entries', so the name should contains 'BetweenEntries'.



> I recommend to use the following names or similar for the the 2 rules:
- TimeAfterEntryRule(this is your actual PR)
_returns true if a minimum number of bars or a minimum duration exist after the last entry_

This rule is not time-based, but it depends on a bar count, so proposed name is misleading - I will not change it.

nimo23:
>MinBarsBetweenTradesRule

`TimeAfterEntryRule` is the new name for that..please read fully before arguing.

>This rule is not time-based, but it depends on a bar count

yes, I understand..barCount can be determined by barCount = Floor(Duration/timePeriod), where Duration is the given Duration from the constructor. If you dont want that, then we have to introduce 4 new rules even if 2 rules have the same logic:

- `TimeAfterEntryRule` (accepts duration in constructor, same logic as TradeOpenedMinimumBarCountRule)
- `TimeBetweenEntryRule` (accepts duration in constructor)
- `TradeOpenedMinimumBarCountRule` (accepts barCount) - the name is aweful long but not precise)
- `BarsBetweenEntryRule` (accepts barCount, but same logic as TimeBetweenEntryRule)

milczarekIT:
Up. This PR is waiting over 8 weeks

milczarekIT:
Up. This PR is waiting over 12 weeks

PR #511, opened 2019-12-29
Log result from CachedIndicator on trace level

milczarekIT:
Fixes #.

- Log result from CachedIndicator on trace level

- [ ] added an entry to the unreleased section of `CHANGES.md` - no need to add, only logging improvement

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,16 @@
 		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 	</properties>
 
+	<dependencyManagement>
+		<dependencies>
+			<dependency>
+				<groupId>ch.qos.logback</groupId>
+				<artifactId>logback-classic</artifactId>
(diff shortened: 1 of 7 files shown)


milczarekIT:
Up. This PR is waiting for 2 months.

pom.xml
```diff
@@ -66,6 +66,16 @@
...
 
+	<dependencyManagement>
+		<dependencies>
+			<dependency>
+				<groupId>ch.qos.logback</groupId>
+				<artifactId>logback-classic</artifactId>
+				<version>1.2.3</version>
+			</dependency>
+		</dependencies>
+	</dependencyManagement>
```
team172011, line 77: I am ok with adding a specific logger implementation for test cases and for examples. But not here in parent pom

milczarekIT: Please check it once again: it is only dependencyManagement, not a dependecy included

PR #508, opened 2019-12-23
Configured properly formatter validation in Travis build

milczarekIT:
Fixes #.

PR https://github.com/ta4j/ta4j/pull/456 turned of formatter validation, so this PR is fixing it.

Changes proposed in this pull request:
- Restored travis config for formatter validation/checking

- [ ] added an entry to the unreleased section of `CHANGES.md`

Issue #320, opened 2018-09-15
Common code formatter for project


milczarekIT:
I noticed that code is pushed by few contributors, I see everyone is using other code formatter (or even sometimes code formatter is not used). My suggestion is to start using common code formatter for whole project.

I am using IntelliJ (v. 2018.2) and think default formatting in this tool is very well configured. Of course we can adjust this formatter to project's need. I think it is a good idea to export formatter from IntelliJ to xml file, add them somewhere and add this info to Wiki for contributors page.

How do you think about this?

team172011:
In general I have no opposition to design and add a code formatter, as far at is easy to import into eclipse, intellij and netbeans. I want to avoid that contributors are dissuaded by complying to much conditions for a PR

yurivin:
I think it is better to make auto code format during CI.

team172011:
@yurivin that sound great. Do you have any examples experience?

milczarekIT:
A common practice is to push code to the repository formatted. IDE like IntelliJ allow to format code  during commit/push, so I you forgot to format code, IDE will do this.

I think we should choose one IDE which is used/supported by contributors and select/configure formatter for them.

yurivin:
@team172011 
Maven plugin that may help. Check format (http://maven.apache.org/plugins/maven-checkstyle-plugin/).

milczarekIT:
@yurivin, @team172011 which IDE do you use?
Me and @jansoren prefer IntelliJ. Community edition which is free is enough, but for open source projects you can request for ultimate licenses: https://www.jetbrains.com/buy/opensource/

In my opinion Eclipse is dying (https://qiita.com/dunebookcom/items/4df79f5c390abe000ff3).

Maven Checkstyle Plugin is not for code formatting, this plugin is rather for code style, so for example it is checking if you have correct modifiers ordering e.g. 'private static final' (this one is in Java Sepcification Language) instead of 'private final static'.

formatter-maven-plugin supports only Eclipse code formatter.

yurivin:
@milczarekIT I use IntelliJ Idea.

It is not a problem that formatter-maven-plugin supports only Eclipse code formatter. https://github.com/HPI-Information-Systems/Metanome/wiki/Installing-the-google-styleguide-settings-in-intellij-and-eclipse

It has eclise style and it is possible import in to IntelliJ Idea to.

milczarekIT:
[...]

PR #525, opened 2020-02-29
Update BarSeriesUtils.class

nimo23:
- Added javadoc


nimo23:
Due to https://github.com/ta4j/ta4j/pull/508/commits/2c878381ea5a090f1483e1a2539926cbcf799be8 

> The command "mvn -B test license:check formatter:validate" exited with 1.

Why not add `formatter:format` ? 
`mvn -B test license:check formatter:format formatter:validate`

Actually, user cannot do the task manually when they create a PR within github editor.

milczarekIT:
Because `formatter:format` only formats code and it is not committed and pushed to repo.
If `formatter:format` will be added, then it is required to add additional step to travis (to push changes to repo). Code should be formatted by user, it can be done automatically by maven plugin, build only checks if code is formatted correctly.

nimo23:
@team172011 maybe you can improve the build step as suggested in https://github.com/ta4j/ta4j/pull/456#issuecomment-531818365. I will see other contributers having issues with that so better to shift the formatting to the server..

nimo23:
The code formatting issues which will lead to contributing issues for others has already discussed in https://github.com/ta4j/ta4j/issues/320 and I see that there is no progess in that. Btw, I someone uses GitHub-Editor then the code formatting issue always happens. Not nice!

nimo23:
replaced by https://github.com/ta4j/ta4j/pull/553

PR #532, opened 2020-03-14
Added support for arbitrary number of arguments on AndRule and OrRule.

kennethjor:
- Added support for arbitrary number of arguments on AndRule and OrRule.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/trading/rules/AndRule.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/trading/rules/AndRule.java
@@ -34,32 +34,45 @@ import org.ta4j.core.TradingRecord;
  */
 public class AndRule extends AbstractRule {
 
-    private final Rule rule1;
-    private final Rule rule2;
+    private final Rule[] rules;
 
     /**
      * Constructor
      *
-     * @param rule1 a trading rule
-     * @param rule2 another trading rule
+     * @param rules all trading rules
      */
-    public AndRule(Rule rule1, Rule rule2) {
-        this.rule1 = rule1;
-        this.rule2 = rule2;
+    public AndRule(Rule... rules) {
+        if (rules == null || rules.length < 2) {
+            throw new IllegalArgumentException(
+                    String.format("At least 2 rules must be supplied, %s seen", rules == null ? 0 : rules.length));
+        }
+        this.rules = rules;
     }
 
     @Override
     public boolean isSatisfied(int index, TradingRecord tradingRecord) {
-        final boolean satisfied = rule1.isSatisfied(index, tradingRecord) && rule2.isSatisfied(index, tradingRecord);
+        boolean satisfied = true;
+        for (Rule rule : rules) {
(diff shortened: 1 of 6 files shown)


ta4j-core/src/main/java/org/ta4j/core/trading/rules/OrRule.java
```diff
@@ -34,38 +34,46 @@
  */
 public class OrRule extends AbstractRule {
 
-    private final Rule rule1;
-    private final Rule rule2;
+    private final Rule[] rules;
```
nimo23, line 37: Please provide getter for rules:

```
public Rule[] getRules() {
      return rules;
}
```

nimo23:
I really like this PR, thanks!

ta4j-core/src/main/java/org/ta4j/core/trading/rules/OrRule.java
kennethjor: Done!

nimo23:
Is it better to add a hint to `IllegalArgumentException` in AndRule/OrRule ? For example:

`IllegalArgumentException("You must add at least 2 rules.") `

kennethjor:
> Is it better to add a hint to `IllegalArgumentException` in AndRule/OrRule ? For example:
> 
> `IllegalArgumentException("You must add at least 2 rules.") `

Done, and also added a null check for good measure.

ta4j-core/src/main/java/org/ta4j/core/trading/rules/AndRule.java
```diff
@@ -34,32 +34,46 @@
...
      *
-     * @param rule1 a trading rule
-     * @param rule2 another trading rule
+     * @param rules all trading rules
      */
-    public AndRule(Rule rule1, Rule rule2) {
-        this.rule1 = rule1;
-        this.rule2 = rule2;
+    public AndRule(Rule... rules) {
+        if (rules == null || rules.length < 2) {
```
nimo23: should be 

`rules == null || rules[0] == null || rules.length < 2`

because the array can contain a null value, which must be considered.

milczarekIT:
Hi @kennethjor,
Some time ago there was a proposal to introduce this feature, but the idea was rejected by community: https://github.com/ta4j/ta4j/pull/302

kennethjor:
> Hi @kennethjor,
> Some time ago there was a proposal to introduce this feature, but the idea was rejected by community: #302

@milczarekIT Looking at the other PR, it seems it was rejected mainly because the implementation was superfluous? Or do you feel it's just the general concept that was frowned upon?

milczarekIT:
In my opinion, the general concept was rejected.

If you need multiple rules you can still use sth like that:
`rule1.and(rule2).and(rule3)`.

In current impl you have easier access to getters.

kennethjor:
All right.

nimo23:
@kennethjor please don't close this issue only because of one opinion. Wait for others what they have to say. Thanks.

Issue #487, opened 2019-10-22
TradingRecord: rule.isSatisfied(tradingRecord) with splitted orders


nimo23:
I have a question:

Some `rules` need a `tradingRecord` to get a valid `isSatisfied(tradingRecord)`.

Does these rules also work with splitted orders?

According to https://github.com/ta4j/ta4j/issues/466

> you could very well execute a buy order with a lot size of 3, and then a later execute a sell order of size 2, then later sell order size 1. 

**Does ta4j handles such scenarios within `rule.isSatisfied(tradingRecord)`?**

**Following example.** 

The **tradingRecord** has these orders:

```
Order 1: enter with volume=3 (enter signal comes from rule.isSatisfied(tradingRecord))
Order 2: exit with volume=2 (exit signal comes from rule.isSatisfied(tradingRecord))
(I am waiting for another exit with volume=1 as I could not close the whole position within "Order 2")
```

The exit for "Order 2" was triggered by `rule.isSatisfied(tradingRecord)`. What happens now? Does the tradingRecord propagate the needed exit `volume=1` to `rule.isSatisfied(tradingRecord)`? Does the rule know that there is another need for an exit?

milczarekIT:
> 

> According to #466
> 
> > you could very well execute a buy order with a lot size of 3, and then a later execute a sell order of size 2, then later sell order size 1.

This is not supported by `BarSeriesManager`.
`BarSeriesManager` is very simply a strategy runner for very simply backtesting.
Described scenarios are of course possible with custom strategy runners.


> Does ta4j handles such scenarios within rule.isSatisfied(tradingRecord)?

`BarSeriesManager` does not allow to close position partially.

nimo23:
@milczarekIT thanks for the answer.

Actually, users excepts that such scenarios described above already work. Not only because I can use someting like the following code which is actually valid and does not throw any errors:

```
// enter with volume=3
recorder.enter(endIndex, price, 3);
// enter with volume=2
recorder.exit(endIndex, price, 2);
```

Hence, it is a bug within BarSeriesManager which must be resolved to handle such orders.

milczarekIT:
> Actually, users excepts that such scenarios described above already work.

They work but not with `BarSeriesManager`.

> // enter with volume=3
> recorder.enter(endIndex, price, 3);
> // exit with volume=2
> recorder.exit(endIndex, price, 2);

[...]

PR #534, opened 2020-03-14
Profit loss percentage fix

kennethjor:
- Fixed incorrect calculations over multiple trades on ProfitLossPercentageCriterion.


kennethjor:
So while the original calculations over `TradingRecord`, I'm not entirely sure these are correct either. Perhaps someone can confirm? I'm happy to submit a fix with the proper calculations for profit percentage over multiple trades.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/ProfitLossPercentageCriterion.java
```diff
@@ -36,10 +36,20 @@
...
+
     @Override
     public Num calculate(BarSeries series, TradingRecord tradingRecord) {
-        return tradingRecord.getTrades().stream().filter(Trade::isClosed).map(trade -> calculate(series, trade))
-                .reduce(series.numOf(0), Num::plus);
+        Num zero = series.numOf(0);
+        Num profit = profitLoss.calculate(series, tradingRecord);
+        Num purchaseCost = tradingRecord.getTrades().stream()
+            .filter(Trade::isClosed)
+            .map(trade -> series.getBar(trade.getEntry().getIndex()).getClosePrice())
```
nimo23: Summing up the `closePrice` is not correct, the `closePrice` is only the close price from the bar and nothing else. If you want to do it that way, then you must sum up the total cost of the trade (by using pricePerAsset, amount, costModel of the actual `Order.class`).

kennethjor: @nimo23 I've added the trade amount into the mix. I don't think the cost model should be taken into account here, because that wasn't included in the original implementation, nor do any of the others do it.

However, I am working on a PR to make all the analysis implementations use the `Trade::getProfit()` method, rather than them all calculating it themselves in various ways.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/ProfitLossPercentageCriterion.java
```diff
@@ -36,10 +36,22 @@
...
+
     @Override
     public Num calculate(BarSeries series, TradingRecord tradingRecord) {
-        return tradingRecord.getTrades().stream().filter(Trade::isClosed).map(trade -> calculate(series, trade))
-                .reduce(series.numOf(0), Num::plus);
+        Num zero = series.numOf(0);
+        Num profit = profitLoss.calculate(series, tradingRecord);
+        Num purchaseCost = tradingRecord.getTrades().stream()
+            .filter(Trade::isClosed)
+            .map(trade -> series.getBar(trade.getEntry().getIndex()).getClosePrice()
```
milczarekIT: Calculations based on close price are incorrect. @nimo23  is right

milczarekIT:
@kennethjor your implementation is wrong.
You wrote: "incorrect calculations over multiple trades on ProfitLossPercentageCriterion".
Could you describe the problem?

milczarekIT:
Travis build fails due to:

> Failed to execute goal net.revelc.code.formatter:formatter-maven-plugin:2.10.0:validate (default-cli) on project ta4j-core: File '/home/travis/build/ta4j/ta4j/ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/ProfitLossPercentageCriterion.java' has not been previously formatted.  Please format file and commit before running validation

kennethjor:
> You wrote: "incorrect calculations over multiple trades on ProfitLossPercentageCriterion".
> Could you describe the problem?

@milczarekIT @nimo23 You're both absolutely right, the "fixed" implementation I submitted is incorrect. I opened this PR at 4 AM.

I don't believe it's possible to fix it currently. Let me explain why:

This class is supposed to calculate the profit-loss percentage of either a single trade or a series of trades (portfolio?). The implementation per trade is just fine. Enter at 200 and exit at 220, it returns 10, as expected. The implementation for a trading record, however, just summed up all the profit-loss percentages for all the trades. For instance, if you had two trades:

1. enter at 5,000 and exit at 6,000
2. enter at 200 and exit at 220

It would return 30 (20 for the first trade plus 10 for the second trade). This number isn't really useful. [To calculate the profit-loss percentage for an entire portfolio](https://www.investopedia.com/ask/answer/07/portfolio_calculations.asp), you have to look at the thing as a whole and take `end value - start value / start value`, just like you would for a single trade. Since trades have an amount, ideally you would take that amount into account and try to approximate the entire portfolio value. Assuming we start with 5,000 in cash (minimum to cover both trades) and all trades have a fixed amount of 1 (as per `BarSeriesManager::run`), this is what would happen:

1. Start = shared: 0, cash: 5,000
2. Enter first trade - shares: 1, cash: 0
3. Exit first trade - shares: 0, cash: 6,000
4. Enter second trade - shares: 1, cash: 5,800
5. Exit second trade - shares: 0, cash: 6,020

Total profit for the entire trade series was 1,020, which is 20.4% of the original 5,000.

However, since ta4j doesn't have the concept of cash, I think it's not possible to fix this implementation properly. Supposed you start out with 10,000 in cash, your profit would still be 1,020, meaning the profit percentage should be 10.2%.

_Alternatively_, when running over several trades, we could calculate the average profit-loss percentage per trade, perhaps? So in the above example, the result would be 14.89% (`(1.1 * 1.2)^(1 / 2) - 1`). Thoughts?

nimo23:
>when running over several trades, we could calculate the average profit-loss percentage per trade, perhaps?

I think it is **not** a good idea to use the weighted average. You already have entry and exit trades with settled prices so you can use that. What we should take into account is that the trading record could have an opened trade at the end, which is not considered in the calculation by  `Trade.getProfit()`. Or?

nimo23:
> you have to look at the thing as a whole and take end value - start value / start value

This formula is not suitable because each trade has a entry and exit. What you want to do with your formula behaves like BuyAndHold..the actual calculation is fine.

victorchung0325:
In ProfitLossPercentageCriterion calculation, the calculation divided by "entryPrice" no matter long order or short order is not correct, for short position, should divided by "exitPrice" instead of "entryPrice" (similiar logic can be found in TotalProfitCriterion.calculateProfit())

kennethjor:
Tests are failing, don't have time to look into it properly right now.

nimo23:
@team172011 this PR is waiting more than n half n year. Would be nice to finish these issues.

team172011:
I am not sure if this PR still is finished or correct. I do not understand the example from @kennethjor 
If you have 6000 cash, why should you only spend 200 for the second trade?
I think the basic assumption in this class and its calculations is the fact that you are always 'in or out' with all of your cash.

kennethjor:
If I remember correctly, in order to calculate this correctly over multiple trades, you need the concept of cash and portfolio value. This was just me trying to approximate that. And yeah, it's been ages, so it's all a bit fuzzy.

> If you have 6000 cash, why should you only spend 200 for the second trade?
Because practicing position sizing it a good thing :) And yeah, I agree, the classes assume you're all-in. It would be a major refactor to remove that assumption.

nimo23:
@team172011 @kennethjor I think we should close this issue: the current impl of PL-percentage with the TradingRecord is just fine. Sure, we can or must better this when we work on https://github.com/ta4j/ta4j/issues/487. But actually, we can close this issue.

PR #535, opened 2020-03-14
Fixed return calculations on short trades.

kennethjor:
- Fixed TotalProfitCriterion calculations on short trades.

The previous return on short trades was based on the exit price, this is incorrect. It should still be based on the entry price since that represents the initial investment made. A short trade selling at 100 and then buying 95 will have the same 5% return as a long trade buying at 100 and selling at 105.


nimo23:
So you see, as I already said in https://github.com/ta4j/ta4j/pull/525:

> **I will see other contributers having issues with that** so better to shift the formatting to the server..

And as I said a few times ago: I will not adapt my code formatter within my IDE only to suit your code formatter of your IDE..so maybe it can be shifted to the server. Would be nice!

milczarekIT:
@nimo23  why can't you understand that we are not using my code formatter. This project is using code formatter defined in https://github.com/ta4j/ta4j/blob/master/ta4j-core/code-formatter.xml since https://github.com/ta4j/ta4j/pull/438 
You do not need to adapt your code formatter in IDE, because maven plugin is added and code formatting can be performed by maven build.

nimo23:
I understand, but it would be nice if the maven plugin is executed in CI instead..I already said it in https://github.com/ta4j/ta4j/pull/456#issuecomment-531818365 and it was also handled in https://github.com/ta4j/ta4j/issues/320#issuecomment-519981483.

kennethjor:
> Calculations based on close price are incorrect.

@milczarekIT I absolutely agree, but all the analysis classes are doing this currently. I'm working on a PR elsewhere to fix this and unify all the profit calculations. I'm hoping to submit it today, but in the process of doing that, I found a couple of basic problems in the implementations and didn't want to include bug fixes in the same PR as a refactor. Keeping things separated.

I do also recognise that the original implementation here used the prices on the orders. Would you like me to add that back in, but keep the overall fix in-place?

kennethjor:
Conflicts fixed. Also ran the formatter, hence the extra files.

kennethjor:
@victorchung0325 Agreed on point 2, this PR has a number of conflicts I'll need to resolve at some point.

Regarding point 1, the previous implementation was incorrect.

"Return" is the amount of money compared to the original investment which comes back. So on a return of 110%, you make 10% over your original investment and on an 85% return, you lose 15% of your original investment. For both a long and a short trade, your original investment is the entry price. A long trade entering at $10 and exiting at $12 is a 20% profit. A short trade exiting at $8 would also be a 20% profit. On a long trade, your maximum profit is infinite (unlikely, but mathematically, it is) because the price can go as high as it likes and the maximum loss is finite. On a short trade, your maximum loss is infinite and the maximum profit is finite, for the same reasons. This property is represented in the new implementation.

If you used the original formula, a total profit on a short trade with entry at $10 and exit at _x_, the profit would approach infinity as the _x_ approaches $0, before finally resulting in a division by zero:
image
_Entry price is $10, x is exit price, y is return [source](https://www.wolframalpha.com/input/?i=plot+10%2Fx+from+0+to+10)._

Technically, that would mean that if I exit at $0.01, my return would be 1,000% and thus my actual profit should be $990. Yet, I only made $9.99 ($10 - $0.01). Where is my remaining $980.01?

Conversely, in the new implementation, you get a nice flat line:
image

_Entry price is $10, x is exit price, y is return [source](https://www.wolframalpha.com/input/?i=plot+%2810-x%29%2F10%2B1+from+0+to+30)._

Notice how the return actually goes negative, this represents the fact that you can lose more than 100%.

See also: https://www.investopedia.com/ask/answers/05/maxreturnshortsale.asp

kennethjor:
I believe this PR is ready once again.

victorchung0325:
For point 1:

You mentioned the below 2 points:
(A) "Return" is the amount of money compared to the original investment which comes back.
(B) For both a long and a short trade, your original investment is the entry price.

I agree (A) but not agree (B)

For short trade, investor receive money at entryPrice and pay money at exitPrice.
Investment is the money paid by investor, which is the exitPrice
Return is the money received by investor, which is entryPrice
Profit % = (Return - Investment) / Investment * 100% = (entryPrice - exitPrice) / exitPrice

For point 2:
In TotalProfitCriterion.java, should use trade.getProfit() not trade.getGrossProfit()?
Becasuse trade.getProfit() will include tradeCost but trade.getGrossProfit() will not include tradeCost

kennethjor:
@victorchung0325 I think we have a separate definition of "return" here. I don't think it matters where the money comes and goes. We invest money in order to gain wealth, right? Any investment I make where I have $10 before making the investment and $12 after completing the investment, that's 20% profit and 120% return, right? It shouldn't matter if it's a long trade, short trade, gamble at a casino, insurance payout, sports bet, loan, tax refund, savings account, etc. The vehicle is, IMHO, irrelevant.

Re point 2: yeah, should probably use the net instead of the gross.

victorchung0325:
I agreed with kennethjor for point 1, thx for explain.
For point 2, we both agreed to use trade.getProfit() in TotalProfitCriterion.calculate()

PR #537, opened 2020-03-15
Refactored profit and return calculations for all analysis criterions.

kennethjor:
* Implemented `getGrossReturn` for return calculations on trades.
* Moved the fallback on the bars' close price found in `TotalProfitCriterion` to be handled by `Trade` and `Order`.
* A lot of analysis criterions had their own internal calculations for profits and returns. Some of these were taking into account short trades, some weren't. I have replaced all of these with the `getProfit` and `getGrossReturn` on `Trade`.
* This also means all the analysis calculations now use the entry and exit numbers on the trades themselves, not the bars on which they occurred. In live trading, this is rarely the case!
* Implemented unit tests for short trades where missing.
* It also means all these calculations take trading costs into account.
* Updated a number of the javadoc entries in the analysis package to provide better descriptions.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -16,3 +16,4 @@ Contributors
     nimo23
     edlins
     milczarekIT (Bartosz Milczarek)
+    kennethjor (Kenneth Jørgensen)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 ### Fixed
 
 ### Changed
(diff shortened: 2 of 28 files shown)


kennethjor:
As a side-note, I have a few other PRs open pertaining to bugs found while doing this work. This PR doesn't include any of those fixes. Once (if) they're merged, this PR will need to be updated accordingly.

team172011:
Hi @kennethjor thanks! I will merge this and you can continue with your bug fixes.

kennethjor:
@team172011 Thank you very much!

PR #541, opened 2020-03-17
Added: Num.isBetween()

nimo23:
- Added method isBetween()
- correct javadoc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - :tada: **Enhancement** Added TimeRangeRule for trading within time ranges.
 - :tada: **Enhancement** Added floor() and ceil() to Num.class
 - :tada: **Enhancement** Added getters getLow() and getUp() in CrossedDownIndicatorRule
+- :tada: **Enhancement** Added isBetween(start, end) in Num.class
 - :tada: **Enhancement** Added BarSeriesUtils: common helpers and shortcuts for BarSeries methods.
 - :tada: **Enhancement** Improvements for PreviousValueIndicator: more descriptive toString() method, validation of n-th previous bars in
  constructor of PreviousValueIndicator 
--- a/ta4j-core/src/main/java/org/ta4j/core/num/Num.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/num/Num.java
@@ -203,9 +203,25 @@ public interface Num extends Comparable<Num>, Serializable {
      * @param end the end value, not null
      * @return true if this is between start (inclusive) and end (inclusive)
      */
-	default boolean isBetween(Num start, Num end) {
-        return isGreaterThanOrEqual(start) && isLessThanOrEqual(end);
-	}
+    default boolean isBetween(Num start, Num end) {
+        return isBetween(start, end, true, true);
+    }
+	
+    /**
+     * Checks if this value is between start and end.
+     * 
+     * @param start the start value, not null
+     * @param end the end value, not null
+     * @param startInclusive if this value should include start
+     * @param endInclusive if this value should include end
+     * @return true if this is between start and end
+     */
+    default boolean isBetween(Num start, Num end, boolean startInclusive, boolean endInclusive) {
+        boolean starts = startInclusive ? isGreaterThanOrEqual(start) : isGreaterThan(start);
+        boolean ends = endInclusive ? isLowerThanOrEqual(start) : isLowerThan(start);
+        return starts && ends;
+    }
+
 
     /**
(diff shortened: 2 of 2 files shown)


nimo23:
maven formatter plugin is not executed when using github editor. One more reason to shift the formatting to the CI-server..

nimo23:
>I think the implementation and usage of this kind of method should be up to the user.

the methods I use are only provided by Num, nothing else. And isBetween is a very common math operation..but sure, you decide..

nimo23:
>I think the implementation and usage of this kind of method should be up to the user.

we also have` isGreaterThan`, `isLowerThan`..and similiar..it is the right place to put `isBetween()` within `Num`.

team172011:
similar to `minus()` and `plus()` methods `isGreater|Than` and `isLower|Than` are equivalencies for >|>= and <|<=, operators that can only be applied on primitives in java. isBetween() is a higher level math operation that can easily implemented by the user.

nimo23:
>that can easily implemented by the user.
And why not free up the user from doing it. `isBetween` is a basic aritmethic math operation. And is a common functionality needed for different indicator calculations..

nimo23:
As you know, `isBetween()` can be very well applied on primitives in java..

milczarekIT:
What is more... build is not compiling.
Methods  `isLowerThanOrEqual` and `isLowerThan` do not exist.

nimo23:
We handle range conditions already in Num-Implementation: `isGreaterThan|orEqual`, `isLowerThan|orEqual`. `isBetween` is actually missing and I did a right implementation for all kinds of `Num`-implementations because _it is only based on the methods definined in Num_, nothing else. 

Again, **please explain in facts**, why do you think, there is no need nor it is not the right place to put it into `Num`. The method `isBetween` is a logical consequence of the methods we already defined in `Num`. I cannot share your opinion. It's unlogical to have different kinds of mathematical methods within the Num-implementation and you refuse such a little method which only eats its own bottle. But again: @team172011 you decide.

team172011:
> Again, please explain in facts, why do you think, there is no need nor it is not the right place to put it into Num. 

As I said, we want to provide just basic functionality for the `Num` interface so that it can be handled by users that are familiar with javas primitives like `double` or `int`. E.g `isLessThan` is an necessary method to check if a Num `a` has a lover value than Num `b`, without extracting the delegate and use conditional operators such as `>` for primitives or` a.compareTo(b)` for BigDecimal.
You can compare ta4js `Num` to javas **Double** or **Integer** class or to the **BigDecimal** api. They also do not provide an `isBetween()` method.
As far as `Num` implements the `Comparable` interface we could get rid of `isLowerThan`, `isGreaterThan`, .. , methods as well, but we decided to provided (or to keep) them to simplify the `NaN` handling. 

> The method isBetween is a logical consequence of the methods we already defined in Num.

From my point of view this is not a reason for adding this method. There are millions of "logical consequences" of those methods, should they all be implemented? There is not even a single user request for this simplification, I think because ta4js Num api is simple enough to achieve this range check. Adding more methods means also more effort in maintenance. And speaking of maintenance: you did not provide any unit test for your method. What would be the current result and the expected result of `numOf(5).isBetween(NaN, numOf(10), true, true)?` I think there is a missing `NaN` check...

Especially with those `startInclusive` and `endInclusive` parameters the `isBetween` method does not look like something that makes things easier. The user must read the javadoc to really understand this method and in the end you are just replacing the usage of isLowerThan or isLessThanOrEqual with two boolean parameters.

nimo23:
>The user must read the javadoc to really understand this method 

The `inclusive`-properties are already documented in the API doc..but I can improve the javadoc if there is any unclear description. For the test cases: I find no need, because it only uses methods from `Num` which already provide test cases.

>and in the end you are just replacing the usage of isLowerThan or isLessThanOrEqual with two boolean parameters.

`isBetween()` has different use cases than `isGreaterThan|orEqual`, `isLowerThan|orEqual`.

>There are millions of "logical consequences" of those methods, should they all be implemented?

Look at:

```
>
>=
<
<=
isBetween (actually missing)
```

No more possibilities of range queries can be found, because all possibilities are implemented if we provide `isBetween()`. Having `isBetween()`, the `Num` can be seen as complete.

And `isBetween()` is not only a good alternative of a weaker but controllable `equals()`-method when using `DoubleNum` but it is also a shortcut for indicators needed that method. If I have a few indicators making use of this method, I must duplicate this method: code, maintenance and test cases increases only because we do not abstract this method from the user. For **numercial methods**, range queries are very common and `isBetween()` is a very common range query which deserves the same right as `isGreaterThan|orEqual`, `isLowerThan|orEqual`. Num deserves generalized numerical methods and the best place to put those numerical methods is `Num.class`. I will not debate again, if you are against it then close this issue.

PR #548, opened 2020-03-22
Update NaN.java

nimo23:
Add method valueOf()

- added method valueOf() (needed for Builder, be consistent with DoubleNum, DecimalNum)

With valueOf() it is possible to use the Builder, for example:

`BaseBar.builder(NaN::valueOf, NaN.class).timePeriod(..).openPrice(NaN).build();`

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/num/NaN.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/num/NaN.java
@@ -50,6 +50,16 @@ public class NaN implements Num {
 
     private NaN() {
     }
+    
+    /**
+     * Returns a {@code Num} version of the given {@code Num}.
+     *
+     * @param val the number
+     * @return the {@code Num}
+     */
+    public static NaN valueOf(NaN val) {
+        return val;
+    }
 
     @Override
     public int compareTo(Num o) {
--- a/ta4j-core/src/main/java/org/ta4j/core/num/NaN.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/num/NaN.java
@@ -1,7 +1,7 @@
 /**
  * The MIT License (MIT)
  *
- * Copyright (c) 2014-2017 Marc de Verdelhan, 2017-2019 Ta4j Organization & respective
+ * Copyright (c) 2014-2017 Marc de Verdelhan, 2017-2020 Ta4j Organization & respective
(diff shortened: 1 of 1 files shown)


nimo23:
@team172011 something is wrong with the license header..and please look at other classes, for example: DecimalNum.java, there is a duplicate license header..does the formatter plugin adds an additional the license header?

nimo23:
@team172011 On another questions.: In `DoubleNum` there is the following apiNote:

> the delegate should never become a NaN value.

Can you explain the reason for this a little more? 

There is a `Double.NaN` but it has nothing to do with our Num's `NaN`. And why is this apiNote missing in DecimalNum (which could also have BigDecimal.NaN)?

ta4j-core/src/main/java/org/ta4j/core/num/NaN.java
```diff
@@ -50,6 +50,16 @@
...
     private NaN() {
     }
+    
+    /**
+     * Returns a {@code Num} version of the given {@code Num}.
+     *
+     * @param val the number
+     * @return the {@code Num}
+     */
+    public static NaN valueOf(NaN val) {
```
team172011: I would change signature to `valueOf(Number val) -> NaN`

nimo23: Do you mean this:

```
    public static Num valueOf(Number val) {
        return NaN;
    }
```

or that:

```
 public static Num valueOf(Num val) {
        return NaN;
    }
```
?

nimo23: Either we can `valueOf(Num num)` or `valueOf(NaN val)` to be usable for the builder

`BaseBar.builder(org.ta4j.core.num.NaN::valueOf, NaN.class).openPrice(NaN).build();`

I changed the code to use `valueOf(Num num)`.

team172011: I mean 
```
    public static Num valueOf(Number val) {
        return NaN;
    }
```
to be more general

team172011:
>the delegate should never become a NaN value.

It is a very old api note when we were using only one wrapper class. I think it should ensure that we are using our NaN and not working with Double.NaN or BigDecimal.NaN under the hood and getting unexpected behaviour.

PR #553, opened 2020-03-25
Update BarSeriesUtils

nimo23:
- add common helpers


nimo23:
@team172011 maybe we could also put the 

`BarSeries getSubSeries(int startIndex, int endIndex);`

from `BarSeries.class` into `BarSeriesUtils`?

team172011:
> @team172011 maybe we could also put the
BarSeries getSubSeries(int startIndex, int endIndex);
from BarSeries.class into BarSeriesUtils?

Yes I totally agree with that.

PR #554, opened 2020-03-26
unit tests for RecursiveCachedIndicator

jamestomk:
add to units tests for indicators using org.ta4j.core.indicators.RecursiveCachedIndicator
add units test for org.ta4j.core.indicators.RecursiveCachedIndicator


nimo23:
Avoid the use of a stale ta4j version when creating a PR. Always use the current ta4j-develop branch before creating a PR.

jamestomk:
okay i think I understand.  I had to first delete my old clone/fork and re-create from the latest repository, and THEN I could re-fork and pull.

PR #555, opened 2020-03-27
bug fix:  ATR(1) was built as ATR(chopTimeFrame)

jamestomk:
add to units tests for indicators using org.ta4j.core.indicators.RecursiveCachedIndicator
add units test for org.ta4j.core.indicators.RecursiveCachedIndicator


jamestomk:
I think I'm just not getting this branch then pull.   I did it fine once before, but my brain seems to have melted!   It now looks like I cloned an obsolete branch with no 'BarSeries', etc!

PR #573, opened 2020-04-27
Rename Total[Loss/Profit]Criterions to respective [Net|Gross][Profit|Loss]Criterion

nimo23:
- Rename TotalLossCriterion to `NetLossCriterion` 
- Rename TotalProfit2Criterion to `NetProfitCriterion` 
- Rename TotalProfitCriterion to `GrossReturnCriterion` 
- Add `GrossLossCriterion` 
- Add `GrossProfitCriterion` 

- [] added an entry with related ticket number(s) to the unreleased section of `CHANGES.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 
 ### Breaking
 - **Breaking:** **PrecisionNum** renamed to **DecimalNum**
+- **Breaking:** **TotalLossCriterion** renamed to **NetLossCriterion**
+- **Breaking:** **TotalProfit2Criterion** renamed to **NetProfitCriterion**
+- **Breaking:** **TotalProfitCriterion** renamed to **GrossReturnCriterion**
+
+
 
 ### Fixed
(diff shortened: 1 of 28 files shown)


nimo23:
Should we better put all PnL related classes within a package `criteria/pnl` ?

 I think it is better because those classes are very related:

-ProfitLossCriterion.java
-ProfitLossPercentageCriterion.java
- and the 4 classes above

nimo23:
Another alternative: 

Remove all `Net[Loss|Profit]Criterions `and only deal with `Total[Loss|Profit]Criterions` (which is the same as Gross[Loss|Profit]Criterions). I mean if the TradingRecord has commissions then it should be always included in the PnL calculations. Or is there a need to also measure `Net[Loss|Profit]Criterions`? If not, I can remove those `Net`-Criterions from ta4j and rename all `Gross`-Criterions to `Total`-Criterions.

nimo23:
This issue is pending because of https://github.com/ta4j/ta4j/pull/534. 

@team172011 Btw, there are a lot of other PRs I created which I wanted to submit after a next release. Would be nice if you don't forget us:)

PR #579, opened 2020-04-29
BarSeriesUtils: convertBarSeries()

nimo23:
- Add convertBarSeries(): converts barData from one Num to another

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/utils/BarSeriesUtils.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/utils/BarSeriesUtils.java
@@ -124,5 +124,40 @@ public final class BarSeriesUtils {
 
         return missingBars;
     }
+    
+    /**
+	 * Gets a new BarSeries cloned from the provided barSeries with bars converted
+	 * by conversionFunction.
+	 * 
+	 * @param barSeries          the BarSeries
+	 * @param conversionFunction the conversionFunction
+	 * @return new cloned BarSeries with bars converted by conversionFunction
+	 */
+	public static BarSeries convertBarSeries(BarSeries barSeries, Function<Number, Num> conversionFunction) {
+		List<Bar> bars = barSeries.getBarData();
+		if (bars == null || bars.isEmpty() || barSeries.function().equals(conversionFunction)) return barSeries;
+		List<Bar> convertedBars = new ArrayList<>();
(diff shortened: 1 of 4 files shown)


ta4j-core/src/main/java/org/ta4j/core/utils/BarSeriesUtils.java
```diff
@@ -124,5 +128,40 @@ public static Bar replaceBarIfChanged(BarSeries barSeries, Bar newBar) {
...
+					.highPrice(bar.getHighPrice().getDelegate())
+					.lowPrice(bar.getLowPrice().getDelegate())
+					.closePrice(bar.getClosePrice().getDelegate())
+					.volume(bar.getVolume().getDelegate())
+					.amount(bar.getAmount().getDelegate())
+					.trades(bar.getTrades())
+					.build();
+			convertedBars.add(convertedBar);
+		}
+		BarSeries convertedBarSeries = new BaseBarSeries(barSeries.getName(), convertedBars, conversionFunction);
```
team172011, line 163: You are creating a new `BaseBarSeries` with the raw **barData** of the provided barSeries. This may not be an equivalent of the provided **barSeries** (that I would expect when using this function). E.g. the original **barSeries** may be created with another constructor using the **endIndex** and **startIndex** parameter or has been subject to operations that modified the **endIndex** or startIndex.

nimo23: Yes, it's a copy of the orginal barseries..the user can set endIndex and startIndex parameter with the copy. Or what are you suggesting? Should I set endIndex and startIndex in this method?

team172011: yes, you should assing **startIndex** and **endIndex** as well, or just iterate from `barSeries#getStartIndex` to `barSeries#getEndIndex`

nimo23: So you mean instead of 

```
for (int i = 0; i < bars.size(); i++) {
..
}
```


I should use this loop:?

```
for (int i = barSeries.getBeginIndex(); i <= barSeries.getEndIndex(); i++) {
			
}
```

nimo23:
Ok, finished.

team172011 (approved):

PR #580, opened 2020-04-29
DoubleNum.java: provide improved equals()-method

nimo23:
- provide improved equals()

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **ProfitLossPercentageCriterion**: Changed to calculate trade profits using Trade's entry and exit prices.
 - **TotalLossCriterion**: Changed to calculate trade profits using Trade's getProfit().
 - **TotalReturnCriterion** (previously TotalProfitCriterion): Changed to calculate trade profits using Trade's getProfit().
+- **DoubleNum**: Changed the calculation of equals()
 
 ### Removed/Deprecated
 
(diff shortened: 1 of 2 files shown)


nimo23:
Even if we come to the conclusion that this PR is **not** needed, **we should investigate why there are build errors** (maybe we can improve the test cases or find any method which leads to wrong calculations):

```
10:16:10.306 [main] DEBUG org.ta4j.core.num.DecimalNum - 1.928749238479283749238472398472936872364823749823749238749238749283749238472983749238749832749274 does not match

10:16:10.312 [main] DEBUG org.ta4j.core.num.DecimalNum - 1.9287492384792837 within offset 1E-17

10:16:10.315 [main] DEBUG org.ta4j.core.num.DecimalNum - 1.92874923847928375 from 1.928749238479283749238472398472936872364823749823749238749238749283749238472983749238749832749274 does not match

10:16:10.315 [main] DEBUG org.ta4j.core.num.DecimalNum - 1.9287492384792837 from 1.9287492384792837 to precision 18

..

testStringNum[Test Case: 0 (0=DoubleNum, 1=DecimalNum)](org.ta4j.core.TestUtilsTest)  Time elapsed: 0.01 sec  <<< FAILURE!

java.lang.AssertionError: Values should be different. Actual: 1.23456789012346E9

..

testNumNum[Test Case: 0 (0=DoubleNum, 1=DecimalNum)](org.ta4j.core.TestUtilsTest)  Time elapsed: 0.001 sec  <<< FAILURE!

java.lang.AssertionError: Values should be different. Actual: 1.23456789012346E9

..
```

@team172011 Please look at the build log and find out why there are errors.

nimo23:
And which do you think is more precise? ..**5**.. or ..**6**..? This is the question here and the reason why I suggest a more precise double equals method. We should discuss the alternative..

nimo23:
@team172011 And please dont close this issue without discussing. It would not be the first time to find wrong or unprecise test cases.

PR #583, opened 2020-04-30
Rename BacktestExecutor.java to ReportExecutor.java

nimo23:
Rename to ReportExecutor.

The class has nothing to do with backtesting, it is **only** a ReportGenerator (according to the interface ReportGenerator<T>).

- Rename BacktestExecutor.java to ReportExecutor.java


team172011:
The methods of `BacktestExecutor` return  lists of `TradingStatement` containing not only `TradeStatsReport` and the `Strategy` but also a `PerformanceReport`. So execution of `BacktestExecutor` does more than only creating a report

PR #87, opened 2017-11-13
New: "Decimal|BooleanTransformIndicator"

nimo23:
Changes proposed in this pull request:
- DecimalTransformIndicator.java
- BooleanTransformIndicator.java

- [x] added an entry to the unreleased section of `CHANGES.md`

PR #585, opened 2020-04-30
Update DecimalTransformIndicator.java

nimo23:
- Replaced switch-loop for static access.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DecimalTransformIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DecimalTransformIndicator.java
@@ -156,7 +156,7 @@ public class DecimalTransformIndicator extends CachedIndicator<Num> {
 	 * @param coefficient the value for transformation
 	 * @param type        the type of the transformation
 	 */
-	public DecimalTransformIndicator(Indicator<Num> indicator, double coefficient, DecimalTransformType type) {
+	public DecimalTransformIndicator(Indicator<Num> indicator, Number coefficient, DecimalTransformType type) {
 		super(indicator);
 		this.indicator = indicator;
 		this.coefficient = numOf(coefficient);
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DecimalTransformIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DecimalTransformIndicator.java
@@ -37,140 +37,165 @@ import org.ta4j.core.num.Num;
  */
 public class DecimalTransformIndicator extends CachedIndicator<Num> {
 
-    private static final long serialVersionUID = -8017034587193428498L;
+	private static final long serialVersionUID = -8017034587193428498L;
 
-    /**
-     * Select the type for transformation.
-     */
-    public enum DecimalTransformType {
+	/**
+	 * Select the type for transformation.
+	 */
+	public enum DecimalTransformType {
 
-        /**
-         * Transforms the input indicator by indicator.plus(coefficient).
-         */
-        plus,
+		/**
+		 * Transforms the input indicator by indicator.plus(coefficient).
+		 */
+		plus {
+			@Override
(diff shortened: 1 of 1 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DecimalTransformIndicator.java
```diff
@@ -37,140 +37,165 @@
...
+			}
+		};
+
+		abstract Num calculate(Num val);
+	}
+
+	private Indicator<Num> indicator;
+	private Num coefficient;
+	private DecimalTransformType type;
+	private DecimalTransformSimpleType simpleType;
```
team172011, line 150: Not directly related to this PR, but I don't like the design of this indicator. We have two instance variables type and simpleType, but only one can be used per indicator instance and the other one stays `null`. And because of that we always have to do a null check  in calculate method.

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DecimalTransformIndicator.java
```diff
@@ -37,140 +37,165 @@
...
+
+	/**
+	 * Constructor.
+	 * 
+	 * @param indicator   the indicator
+	 * @param coefficient the value for transformation
+	 * @param type        the type of the transformation
+	 */
+	public DecimalTransformIndicator(Indicator<Num> indicator, double coefficient, DecimalTransformType type) {
+		super(indicator);
```
team172011, line 160: for `coefficient` we could use `Number` instead of `double` type

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DecimalTransformIndicator.java
nimo23: How can we else avoid this null check?

team172011: We could split this up into two indicators
- SimpleDecimalTransformIndicator
- DecimalTransformIndicator

Another solution (but imo dirty) could be to merge `DecimalTransformType ` and `DecimalTransformSimpleType` into one enum, and the functions of `DecimalTransformSimpleType` are ignoring the coefficient. And coefficient could get a default value = 0

Btw. MultiplierIndicator.java should also be removed by this indicator, or?

nimo23: Splitting into two classes only to avoid the null-check? I had think about it in the past when I created this indicator and came to the conclusion that it is better to have one indicator dealing with "transforming values by math operations". 

And yes, you can remove `MultiplierIndicator.java` because it is already handled by this indicator.

nimo23: >Another solution (but imo dirty) could be to merge DecimalTransformType and DecimalTransformSimpleType into one enum, and the functions of DecimalTransformSimpleType are ignoring the coefficient. And coefficient could get a default value = 0

I can do this. I think this solution would be better than splitting the indicator into two classes. Should I do this?

nimo23: >Another solution (but imo dirty) could be to merge DecimalTransformType and DecimalTransformSimpleType into one enum, and the functions of DecimalTransformSimpleType are ignoring the coefficient. And coefficient could get a default value = 0

Should I do this?

nimo23: > We could split this up into two indicators

I can do this by another PR.

nimo23:
@team172011 As I said in https://github.com/ta4j/ta4j/pull/87#issuecomment-344244598, we can **remove** `MultiplierIndicator.java` and `AbsoluteIndicator.java`, because they are all handled by this indicator.

PR #595, opened 2020-05-23
add stream support for BarSeries

chautrader:
Fixes #.

- add stream support for BarSeries
- 
- 


nimo23:
This method has no additional use. User can always call list.stream().

team172011:
agree with @nimo23 and this method provides a stream with raw bar data ignoring end or start index of this bar series

chautrader:
Thanks, I missed 
```
    @Override
    public List<Bar> getBarData() {
        return bars;
    }
```

PR #616, opened 2020-07-27
Simplified TransformIndicator, remove redundant indicators

team172011:
- Remove redundant indicators: AbsoluteIndicator and MultiplierIndicator
	-> use DecimalTransformIndicator

- [x] added an entry with related ticket number(s) to the unreleased
section of `CHANGES.md`


milczarekIT:
I cannot see added an entry in `CHANGES.md`

team172011:
I forgot to add the modified CHANGELOG.md, added it now

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **TotalReturnCriterion** (previously TotalProfitCriterion): Changed to calculate trade profits using Trade's getProfit().
 
 ### Removed/Deprecated
+- **MultiplierIndicator**: replaced by DecimalTransformIndicator.
+- **AbsoluteIndicator**: replaced by DecimalTransformIndicator.
```
nimo23: @team172011 Please rename  "replaced by DecimalTransformIndicator." to "replaced by TransformIndicator."

team172011:
In this PR I will also redesing this inidicator. I think something like `new DecimalTransformIndicator.multiply(indicator, 3)` would be much easier than working with two different constructors expecting two different enum types

nimo23:
@team172011 ok then please call it `TransformIndicator` instead of `DecimalTransformIndicator`. And maybe we could provide static access instead of instead of using new-instantiation?

`TransformIndicator.multiply(indicator, 3);`

instead of 

`new DecimalTransformIndicator.multiply(indicator, 3)`

nimo23:
I think it is better to not call it `TransformIndicator`. We should call it `NumTransform`. And the class should be an enum where we can make static calls:

```
NumTransform.multitply(indicator, 3);
NumSimpleTransform.max(indicator, 3, 4);
```

team172011:
Now for instance you can call `TransformIndicator addThree = TransformIndicator.plus(indicator, 3)` or `TransformIndicator absoluteValue = TransformIndicator.abs(indicator)`  or via public constructor   
 ` public TransformIndicator(Indicator<Num> indicator, Function<Num, Num> transformation) `
you can instantiate a TransformIndicator with a custom transform function like:

        Num three = indicator.numOf(3);
        Num ten = indicator.numOf(10);
        TransformIndicator plusThreeDivByTenAbs = new TransformIndicator(indicator, val -> val.plus(ten).dividedBy(three).abs());

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/TransformIndicator.java
```diff
@@ -0,0 +1,139 @@
...
+package org.ta4j.core.indicators.helpers;
+
+import java.util.function.Function;
+
+import org.ta4j.core.Indicator;
+import org.ta4j.core.indicators.CachedIndicator;
+import org.ta4j.core.num.Num;
+
+/**
+ * Decimal transform indicator.
```
nimo23: Use this javadoc:

```
/**
 * Num transform indicator.
 * <p>Transforms the Num of any indicator by using common math operations.
 *
 * @apiNote Minimal deviations in last decimal places possible. During the
 *          calculations this indicator converts {@link Num DecimalNum} to
 *          {@link Double double}
 */
```

PR #632, opened 2020-08-28
Update .travis.yml

nimo23:
- apply formatter

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,4 @@ jdk:
   - openjdk8
 
 script:
-  - mvn -B formatter:validate license:check test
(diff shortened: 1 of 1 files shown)


nimo23:
Another alternative is to use this:

- mvn -B test license:check formatter:format

There is no need to run `formatter:validate` because `formatter:format` will format it according to the formatter rules.

milczarekIT (changes requested):
Please do not turn off validation.

Yes, if you add `formatter:format` then travis during job will format code but do not push it to repo, so it means that validation and formatting is turned off.

@team172011  please do not turn off formatting validation.

PR #636, opened 2020-09-05
Simplified DPOIndicator calculation

team172011:
- Simplify DPOIndicator calculations

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/DPOIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/DPOIndicator.java
@@ -26,6 +26,8 @@ package org.ta4j.core.indicators;
 import org.ta4j.core.Indicator;
 import org.ta4j.core.BarSeries;
 import org.ta4j.core.indicators.helpers.ClosePriceIndicator;
+import org.ta4j.core.indicators.helpers.DifferenceIndicator;
+import org.ta4j.core.indicators.helpers.PreviousValueIndicator;
 import org.ta4j.core.num.Num;
 
 /**
@@ -46,10 +48,10 @@ import org.ta4j.core.num.Num;
  */
 public class DPOIndicator extends CachedIndicator<Num> {
 
-    private final int barCount;
-    private final int timeShift;
-    private final Indicator<Num> price;
-    private final SMAIndicator sma;
+    private static final long serialVersionUID = -3450705409405267767L;
+
+    private final DifferenceIndicator indicatorMinusPreviousSMAIndicator;
+    private final String name;
 
     /**
      * Constructor.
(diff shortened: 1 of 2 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/DPOIndicator.java
```diff
@@ -46,10 +48,10 @@
...
 public class DPOIndicator extends CachedIndicator<Num> {
 
-    private final int barCount;
-    private final int timeShift;
-    private final Indicator<Num> price;
-    private final SMAIndicator sma;
+    private static final long serialVersionUID = -3450705409405267767L;
+
+    private final DifferenceIndicator indicatorMinusPreviousSMAIndicator;
+    private final String name;
```
nimo23, line 54: remove the `name`-property

ta4j-core/src/main/java/org/ta4j/core/indicators/DPOIndicator.java
```diff
@@ -69,19 +71,22 @@ public DPOIndicator(BarSeries series, int barCount) {
...
-        timeShift = barCount / 2 + 1;
-        this.price = price;
-        sma = new SMAIndicator(price, this.barCount);
+        int timeFrame = barCount / 2 + 1;
+        final SMAIndicator simpleMovingAverage = new SMAIndicator(price, barCount);
+        final PreviousValueIndicator previousSimpleMovingAverage = new PreviousValueIndicator(simpleMovingAverage,
+                timeFrame);
+
+        this.indicatorMinusPreviousSMAIndicator = new DifferenceIndicator(price, previousSimpleMovingAverage);
+        this.name = String.format("%s barCount: %s", getClass().getSimpleName(), barCount);
```
nimo23, line 80: remove `this.name = ..`

ta4j-core/src/main/java/org/ta4j/core/indicators/DPOIndicator.java
```diff
@@ -69,19 +71,22 @@ public DPOIndicator(BarSeries series, int barCount) {
...
     @Override
     protected Num calculate(int index) {
-        return price.getValue(index).minus(sma.getValue(index - timeShift));
+        return indicatorMinusPreviousSMAIndicator.getValue(index);
     }
 
     @Override
     public String toString() {
-        return getClass().getSimpleName() + " barCount: " + barCount;
+        return name;
```
nimo23, line 90: change to `return getClass().getSimpleName() + " barCount: " + barCount;`

nimo23 (approved):

ta4j-core/src/main/java/org/ta4j/core/indicators/DPOIndicator.java
team172011: why not caching this value that may be requested several times in loggers or other output data?

ta4j-core/src/main/java/org/ta4j/core/indicators/DPOIndicator.java
team172011: same as line 90

ta4j-core/src/main/java/org/ta4j/core/indicators/DPOIndicator.java
nimo23: 1. we should be consistent: all other indicators, as far as I know, do not cache value only because of the logger
2. why caching value that may be not used at all if one decides not to log
3. logging is always optional, no need to add overhead in constructor

team172011: I agree, but in this case we would have to store the barCount only for this purpose in an instance variable. Therefore I thought why not store the indicator name instead.

nimo23: Ok, then leave it as is.

PR #644, opened 2020-11-19
#643 bug fixed 

### LINKED ISSUE #643: [BUG] - ParabolicSarIndicator calculates the value incorrectly
I wrote a test for my claim, and then check it with the TradingView site, then found out that ParabolicSarIndicator.getValue works in a wrong way.

```
package org.ta4j.core.indicators;

import org.junit.Test;
import org.ta4j.core.Bar;
import org.ta4j.core.Indicator;
import org.ta4j.core.mocks.MockBar;
import org.ta4j.core.mocks.MockBarSeries;
import org.ta4j.core.num.Num;

import java.util.ArrayList;
import java.util.List;

import static org.junit.Assert.assertEquals;
import static org.ta4j.core.TestUtils.assertNumEquals;

public class ParabolicSarIndicatorTest extends AbstractIndicatorTest<Indicator<Num>, Num> {

    public ParabolicSarIndicatorTest(Function<Number, Num> numFunction) {
        super(numFunction);
    }

    @Test
    public void startUpAndDownTrendTest() {
        List<Bar> bars = new ArrayList<Bar>();
        bars.add(new MockBar(4261.48, 4285.08, 4485.39, 4200.74, numFunction)); // the first candle of BTCUSDT in the Binance cryptocurrency exchange.
        bars.add(new MockBar(4285.08, 4108.37, 4371.52, 3938.77, numFunction)); // the second ...

        ParabolicSarIndicator sar = new ParabolicSarIndicator(new MockBarSeries(bars));

        assertEquals("NaN", sar.getValue(0).toString());
        assertNumEquals(4485.39, sar.getValue(1));
    }

}
```

as shown in the image that I attached, it should be 4485.39, but the value that the indicator calculates is 4371.52

ParabolicSarIndicatorBug

you can check the numbers in https://www.tradingview.com/ with Daily candles on BINANCE:BTCUSDT symbol.
I checked 18 Aug 2017 ParabolicSarIndicator value and it is not equal to what you can see in the image above.

mohsenht:
Fixes #643.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/ParabolicSarIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/ParabolicSarIndicator.java
@@ -129,7 +129,7 @@ public class ParabolicSarIndicator extends RecursiveCachedIndicator<Num> {
                 Num lowestPriceOfTwoPreviousBars = new LowestValueIndicator(lowPriceIndicator, Math.min(2, index - startTrendIndex)).getValue(index - 1);
                 if(sar.isGreaterThan(lowestPriceOfTwoPreviousBars))
                     sar = lowestPriceOfTwoPreviousBars;
-                currentExtremePoint = new HighestValueIndicator(highPriceIndicator, index - startTrendIndex)
+                currentExtremePoint = new HighestValueIndicator(highPriceIndicator, index - startTrendIndex + 1)
                         .getValue(index);
                 if (currentExtremePoint.isGreaterThan(minMaxExtremePoint)) {
                     incrementAcceleration();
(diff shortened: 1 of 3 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/ParabolicSarIndicator.java
```diff
@@ -126,7 +126,10 @@ protected Num calculate(int index) {
                 currentExtremePoint = getBarSeries().getBar(index).getLowPrice(); // put point on max
                 minMaxExtremePoint = currentExtremePoint;
             } else { // up trend is going on
-                currentExtremePoint = new HighestValueIndicator(highPriceIndicator, index - startTrendIndex)
+                Num lowestPriceOfTwoPreviousBars = new LowestValueIndicator(lowPriceIndicator, Math.min(2, index - startTrendIndex)).getValue(index - 1);
```
team172011, line 132: What is the reason for this change? 
~~current:  Uptrend or downtrend means we have to calculate the extreme point. 
yours: You are replacing this with calculating the lowest min price of the trend or at least of the last two bars if a up trend has been detected?~~ (I read the diff wrong, you were not replacing this, but you are doing this with SAR value replacing, why?)

mohsenht: You may ask this question from who invented this indicator. I used your library to not overwrite what is written before and tested some of my strategies on it and then found out that the ParabolicSarIndicator calculates wrong, so I searched and then found a Wikipedia page for it and do what it says and then tested it on TradingView values, then it seems that works great. 

check it out in the link below.

_**If the next period's SAR value is inside (or beyond) the current period or the previous period's price range, the SAR must be set to the closest price bound. For example, if in an upward trend, the new SAR value is calculated and if it results to be more than today's or yesterday's lowest price, it must be set equal to that lower boundary.**_

https://en.wikipedia.org/wiki/Parabolic_SAR

Thanks for your consideration.

team172011:
Thanks!

mohsenht:
Thank you for this great library.

PR #672, opened 2021-01-27
Added Num#between() 

nimo23:
- :tada: **Enhancement** added Num#isBetween() to test the Num if it is between a range
- **InPipeRule**: reduced complexity by using Num.isBetween()
- **InSlopeRule**: reduced complexity by using Num.isBetween()
- **InSlopeRule**: more unit tests to test against NaN

Would be nice to add `Num#isBetween()`. Other (future) indicators or (future) rules can use this common method instead of duplication code and tests. The `isBetween()`-method is bullet proof and checks agains all kinds of `Null` or `NaN` (which is currently not the case within the above rules, I have thus added more unit tests for that). 

I will **not** add any other methods within Num in future, as we really should keep it sleek. However, `isBetween()` is the missing link between isGt and isLt. Other indicators and rules can make use of it, if they need such a check. We could also look for current indicators or rules which has such a check and replace the code by using `isBetween()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,8 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **ProfitLossPercentageCriterion**: Changed to calculate trade profits using Trade's entry and exit prices.
 - **TotalLossCriterion**: Changed to calculate trade profits using Trade's getProfit().
 - **TotalReturnCriterion** (previously TotalProfitCriterion): Changed to calculate trade profits using Trade's getProfit().
-- **WMAIndicator**: reduced complexity of WMAIndicator implementation
+- **InPipeRule**: reduced complexity by using Num.isBetween()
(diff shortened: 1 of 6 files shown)


nimo23:
@team172011 Looking the Num implementations, there are a lot of other math methods like log(), matches(), abs(), sqrt(), .. So why are you not willing to put such a trivial method in it? Without the `isBetween()`, we have to **duplicate code** (with is currently not consistent and not bug free as it is the case in InPipeRule, InSlopeRule). **DRY** (dont repeat yourself). My PR also reduces code complexity and adds more unit tests, so why are you closing this without any argues? That's not team-friendly.

I also have a few indicators needing this i`sBetween()`, which I wanted to submit in future. But I must admit, if you do not merge this trivial helper I am not willing to submit those PR's. Because I am not willing to hurt against DRY. 

Would be reasonable to think about it again and it would be nice to merge this. Thanks!

team172011:
I will not discuss the same things again and again. Please see #541 for reasons why I do not want to have this function

PR #673, opened 2021-01-28
Refactored: NumberOf(Min,Max)PositionsCriterion

nimo23:
Fixes #.

﻿- **Breaking:** **`NumberOfTradesCriterion`** renamed to **`NumberOfMaxPositionsCriterion`**
- :tada: **Enhancement** **`NumberOfMinPositionsCriterion`** 

Currently the `betterThan()` with NumberOfPositionsCriterion is vague. Thus, split this criterion to two and let the user decide if one is better than the other. For example, for backtesting:

- NumberOf**Min**PositionsCriterion: the more positions, the more tests were done, the better.
- NumberOf**Max**PositionsCriterion: the less positions, the better.


team172011:
Whats the sense behind splitting this up into Min and Max? This criterion was just about counting Positions in a TradingRecord

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfMaxPositionsCriterion.java
```diff
@@ -29,9 +29,9 @@
 import org.ta4j.core.num.Num;
 
 /**
- * Number of position criterion.
+ * Number of maxium allowed position criterion.
  */
-public class NumberOfPositionsCriterion extends AbstractAnalysisCriterion {
+public class NumberOfMaxPositionsCriterion extends AbstractAnalysisCriterion {
```
team172011, line 34: Why renaming this?

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfMinPositionsCriterion.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Number of minimum required position criterion.
+ */
+public class NumberOfMinPositionsCriterion extends AbstractAnalysisCriterion {
+
+    @Override
+    public Num calculate(BarSeries series, TradingRecord tradingRecord) {
+        return series.numOf(tradingRecord.getPositionCount());
+    }
```
team172011, line 39: Same as in NumberOfPositionsCriterion that has been renamed.

ta4j-core/src/test/java/org/ta4j/core/analysis/criteria/NumberOfMinPositionsCriterionTest.java
```diff
@@ -0,0 +1,80 @@
...
+    public NumberOfMinPositionsCriterionTest(Function<Number, Num> numFunction) {
+        super((params) -> new NumberOfMinPositionsCriterion(), numFunction);
+    }
+
+    @Test
+    public void calculateWithNoPositions() {
+        MockBarSeries series = new MockBarSeries(numFunction, 100, 105, 110, 100, 95, 105);
+
+        AnalysisCriterion buyAndHold = getCriterion();
+        assertNumEquals(0, buyAndHold.calculate(series, new BaseTradingRecord()));
```
team172011, line 52: ?

team172011 (changes requested):

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfMaxPositionsCriterion.java
nimo23: Currently the betterThan() with NumberOfPositionsCriterion is vague. Look at the current impl of `betterThan()`. **Why is having more positions better?** Thus, split this criterion to two and let the user decide if one is better than the other. For example, for backtesting:

    NumberOfMinPositionsCriterion: the more positions, the more tests were done, the better.
    NumberOfMaxPositionsCriterion: the less positions, the better.

That's the idea why I wanted to split it to two criterions. Should I add new criterions instead? `NumberOfMinMaxPositionsCriterion` and leave the current as it is?

nimo23:
>Whats the sense behind splitting this up into Min and Max? This criterion was just about counting Positions in a TradingRecord

@team172011 But criterions are ment to be comparable by `betterThan()`. With the actual implementation, it is supposed that _the more trades the better_. But this is not always the case and depend on the users scenario. Hence, split this indicator:

- `NumberOfMinPositionsCriterion`:  the more trades, the better.
- `NumberOfMaxPositionsCriterion`: the less trades, the better.

team172011:
Okay I unterstand, but than I would prefer to delete this criterions. In both ways it is a heavily overengineered feature for just counting the trades

PR #679, opened 2021-02-09
Update Position.java: simpler getGrossReturn() 

nimo23:
- reduce complexity (no need for intermediate result "profitPerAsset")


nimo23:
I find the name **grossReturn** for the current impl irritating. Normally, grossReturn is something like `(a/b) - 1` (or the same: `(a-b)/b`). But in ta4j, the gross return is something different. 

@team172011 Can you please explain?

PR #682, opened 2021-02-12
Update AbstractIndicator.java

nimo23:
- add transient to Logger


nimo23:
**non static** loggers should always have transient keyword.

PR #688, opened 2021-02-18
BarSeriesUtils: addBars()

nimo23:
- add convenient method to add a bunch of bars to existing BarSeries


team172011 (approved):

ta4j-core/src/main/java/org/ta4j/core/utils/BarSeriesUtils.java
```diff
@@ -188,4 +196,40 @@ public static BarSeries convertBarSeries(BarSeries barSeries, Function<Number, N
...
+     */
+    public static List<Bar> sortBars(List<Bar> bars, boolean forceSorting) {
+        if (!bars.isEmpty()) {
+            if (forceSorting) {
+                Collections.sort(bars, BarSeriesUtils.sortBarsByTime);
+            } else {
+                Bar firstBar = bars.get(0);
+                Bar lastBar = bars.get(bars.size() - 1);
+                if (firstBar.getEndTime().isAfter(lastBar.getEndTime())) {
+                    Collections.reverse(bars);
```
nimo23: @team172011 This `Collections.reverse` only exist because I think reverse a list could be faster than sort, but I am not sure. If you think that this performance is negligible, then I can simplify the code to only use the following:

```
public static List<Bar> sortBars(List<Bar> bars){
    if (!bars.isEmpty()) Collections.sort(bars, BarSeriesUtils.sortBarsByTime);
}
```

team172011: Yes, I also think sort() makes sense here. As well as the current implementation assumes that the bars are already sorted but in wrong order

nimo23: So you think I should remove `Collections.reverse` and only use `Collections.sort`?

team172011 (approved):

PR #694, opened 2021-02-23
ProfitLossPercentagePerTime

nimo23:
Fixes #.

- ﻿﻿Enhancement Added ProfitLossPercentagePerTimeCriterion
- 
- 


nimo23:
I think that this indicator does not add any value and the calculation is also doubtful. 

@team172011 If you see any benefit or how this criterion would become better, you can respond. 

For now, I will close this.

PR #720, opened 2021-04-08
introduce RuleWithCtx interface (#719)

### LINKED ISSUE #719: Support StopGain with precise trade exit price
I need support for a TakeProfitRule
- which supports a fixed pips/points (like 20 pips)
- that sets the price to the market price and not the close price

Example:
Candle opens at 20 closes at 21 but the wick was at high 21.5
My stop gain / take profit is at 21.3 --> currently the 21 would be the trade price.
But it should be 21.3

golesny:
new interface is used for the base rules and for backward compatibility
added new StopGainRule2 that supports percentage and fixed gain and
price is used as exit price

Fixes #719 .

- **Enhancement** Added new interface RuleWithCtx to define the trade price
- **Enhancement** AndRule, OrRule, XorRule, NotRule implementing RuleWithCtx
- **Enhancement** new StopGainRule2 that supports percentage and fixed gain and price is used as exit price

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,9 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **Enhancement** Added BarSeriesUtils: common helpers and shortcuts for BarSeries methods.
 - **Enhancement** Improvements for PreviousValueIndicator: more descriptive toString() method, validation of n-th previous bars in
 - **Enhancement** Added Percentage Volume Oscillator Indicator, PVOIndicator.
+- **Enhancement** Added new interface RuleWithCtx to define the trade price
+- **Enhancement** AndRule, OrRule, XorRule, NotRule implementing RuleWithCtx
+- **Enhancement** new StopGainRule2 that supports percentage and fixed gain and price is used as exit price
  constructor of PreviousValueIndicator 
 - :tada: **Enhancement** added getGrossProfit() and getGrossProfit(BarSeries) on Trade.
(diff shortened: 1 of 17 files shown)


nimo23:
Hi, thanks for your commit. 

But I really think that your commit introduces a lot of influences to stable classes (e.g., bad and vague names, ctx for context (which is also vague), shouldDo (interferes with shouldEnter(), which leads to confusion)  and the other stuff which really adds too much complexitiy, brings too much confusion to stable API. I think, for the one issue you try to solve the whole baggage you provided can be simplified a lot by using only one class..no need to pollute so much classes for such an issue..

nimo23:
Unfortunatly, I don't have time to explain every little step.

At first, you should consider to create only a new Rule or Indicator for your issue... If this is not possible for you, then leave it as you think and use this _special_ solution in **your** fork instead of trying to put this to this repo..this repo is considered for general use cases..your solution introduces too much complexity (pollutes too much stable classes) for a very easy task..

nimo23:
Try to solve your solution by using a Rule or Indicator, don't change all those classes only for your task..if this is not possible, then leave this solution as it is in **your** fork, but I really don't like to see this _special_ solution in this repo..

golesny:
Mmmh, not sure how to solve that in an easy way. TakeProfit/StopLoss braket with 20 pips is not a special case in my eyes.
I'll try to re-work that ... but I guess thats not possible as the BarSeriesManager defines the exit price of the trade and do not ask the strategy nor the rules (separation of concern pattern). I have to check what is possible

PR #724, opened 2021-04-15
Exposed both EMAs in MACD indicator to be usable in strategy

petrmac:
Fixes #.

-  Exposed both EMAs in MACD indicator to be usable in strategy

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/MACDIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/MACDIndicator.java
@@ -36,6 +36,8 @@ import org.ta4j.core.num.Num;
  */
 public class MACDIndicator extends CachedIndicator<Num> {
 
+    private static final long serialVersionUID = -6899062131135971403L;
+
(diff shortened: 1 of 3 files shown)


Olbix (approved):

ta4j-core/src/main/java/org/ta4j/core/indicators/MACDIndicator.java
```diff
@@ -36,6 +36,8 @@
  */
 public class MACDIndicator extends CachedIndicator<Num> {
 
+    private static final long serialVersionUID = -6899062131135971403L;
```
nimo23: remove this line. we don't use serialVersionUID.

nimo23 (approved):

ta4j-core/src/main/java/org/ta4j/core/indicators/MACDIndicator.java
petrmac: True, fixed

team172011 (approved):

PR #725, opened 2021-04-16
Added IndicatorUtils with streamOf to create a stream from indicator values

petrmac:
Would this addition be useful? Sometimes, I need to process indicator values in a stream. This helps to convert to stream. 
Perhaps other location is better... no strong opinion.

- Added streamOf utility function in _ IndicatorUtils_

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/utils/IndicatorUtils.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/utils/IndicatorUtils.java
@@ -25,6 +25,7 @@ package org.ta4j.core.utils;
 
 import org.ta4j.core.Indicator;
 
+import java.util.function.Function;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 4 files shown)


petrmac:
Now added another function to allow mapping within indicator while converting to stream. 

Example:
`final Function<IndicatorUtils.IndicatorContext<Num>, ValueWithTimestamp> mapFunction = (
                IndicatorUtils.IndicatorContext<Num> ctx) -> new ValueWithTimestamp(indicator.getValue(ctx.getIndex()),
                        ctx.getIndicator().getBarSeries().getBar(ctx.getIndex()).getEndTime().toInstant());

Stream<ValueWithTimestamp> stream = IndicatorUtils.streamOf(indicator, mapFunction);`

This extracts stream while also joining the Num with Close time. 

Please have a look...

team172011:
I will have a look. In general it makes sense to have an IndicatorUtlis class. We could also move the _toDouble(Indicator<Num> ref, int index, int barCount)_ method from the `Indicator` Interface to this class

petrmac:
@team172011 ,

Please have a look. I have updated as per your suggestion:

> I will have a look. In general it makes sense to have an IndicatorUtlis class. We could also move the _toDouble(Indicator ref, int index, int barCount)_ method from the `Indicator` Interface to this class

Thnx

ta4j-core/src/main/java/org/ta4j/core/utils/IndicatorUtils.java
```diff
@@ -0,0 +1,103 @@
...
+     *                    argument; that contains the indicator and integer index
+     * @param <T>         type of indicator value
+     * @param <R>         result type
+     * @return the resulting value
+     */
+    public static <T, R> Stream<R> streamOf(final Indicator<T> indicator,
+            final Function<IndicatorContext<T>, R> mapFunction) {
+        return IntStream.range(0, indicator.getBarSeries().getBarCount())
+                .mapToObj((int i) -> mapFunction.apply(new IndicatorContext<>(indicator, i)));
+    }
```
team172011, line 68: I am not sure if this method provides any additional value if you already have the _IndicatorUtils#streamOf_ method. I also have reservations regarding the IndicatorContext.class. For each index there will be created a new Object and the 'indicator' parameter is obsolet because it is always the same.

The user can just use the streamOf method and manually apply functions or mappings to the stream.

team172011 (changes requested):
Please add entries to the CANGELOG.md

team172011:
Looking at the changes again, I am also unsure about the location for this functionality. Having an `IndicatorUtils.class` would be ok, but It would also be nice in the `Indicator.class`. Indicator.streamOf like Indicator.toDouble would be cooler with less classes and parameters (like syntactic sugar). What do you think?

ta4j-core/src/main/java/org/ta4j/core/utils/IndicatorUtils.java
nimo23: I have the same opinion as @team172011. Such logic adds unnecessary complexitiy to this lib. The `IndicatorContext` seems to be a very special use case which should not belong to this lib.  Maybe @petrmac can explain use cases for `IndicatorContext` and convince us that this is usable for a variety of things.

team172011:
#741

PR #726, opened 2021-04-16
Added an example Gson serialization to store or load a given Series t…

MarcDahlem:
…o json

Fixes #.

- added an example Gson de-/serialization for BarSeries

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 ### Added
 - :tada: **Enhancement** Added possibility to use CostModels when backtesting with the BacktestExecutor
 - :tada: **Enhancement** added Num#zero, Num#one, Num#hundred
-- **Example** added a json serialization and deserialization example of BarSeries
+- **Example** added a json serialization and deserialization example of BarSeries using google-gson library
 
 ## 0.14 (released April 25, 2021)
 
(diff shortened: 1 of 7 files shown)


MarcDahlem:
Oh well, that failed test run is interesting. It randomly picked one index, which does not need the bigInteger precession.
As I also worked with the Interface "Number" in the Gson-Wrapper-Classes, Gson treats the number as double value instead of BigDecimals.

How should I proceed here?
a) Is the (Base)BarSeries able to contain some bars as BigDecimal and others as double or even integer? (This would mean I could just fix the tests).
b) Or should I force Gson to always produce BigDecimals while adding a new bar to the created bar series?

MarcDahlem:
Okay, nevermind.
The test alone runs fine on all elements; only if executed with all other tests together it didnt work.

Someone thought that changing the defaultfunction in the BarSeriesTest (main-method in ta4jexamples/barSeries/BuildBarSeries.java ) was a good idea...

I made this BuildBarSeries reset the defaultFunction to ensure correct behaviour for all tests running afterwards

team172011 (changes requested):
Thanks for the PR. But this example is mostly based on a third party library that I do not want to add as a dependency to this lib

MarcDahlem:
Its just in the examples. Dont you consider this as better than having no code at all available how to store a data series in a useful format?
Which json library would you prefer over gson? Moshi? Jackson?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 
 ### Added
 - **Enhancement** Added possibility to use CostModels when backtesting with the BacktestExecutor
+- **Example** added a json serialization and deserialization example of BarSeries
```
team172011: Please add using google-gson library

team172011 (approved):

team172011:
Okay, please resolve conflicts and I will merge this example. I think it adds value to the example project in the same way like plotting data using the jFreeCharts library

CHANGELOG.md
MarcDahlem: done

team172011 (approved):

PR #704, opened 2021-03-21
Added criterions

nimo23:
Changes proposed in this pull request:
- :tada: **Enhancement** added **`AverageProfitCriterion.class`**.
- :tada: **Enhancement** added **`AverageLossCriterion.class`**.
- :tada: **Enhancement** added **`ProfitLossRatioCriterion.class`**.
- :tada: **Enhancement** added **`ExpectancyCriterion.class`**.
- :tada: **Enhancement** added **`NumberOfConsecutiveWinningPositionsCriterion.class`**.
- :tada: **Enhancement** added **`LosingPositionsRatioCriterion.class`**.
- :tada: **Enhancement** added **`Position#hasProfit()`**.
- :tada: **Enhancement** added **`Position#hasLoss()`**.

- [x] added an entry with related ticket number(s) to the unreleased section of `CHANGES.md` 

- The `AverageLossCriterion`, `AverageProfitCriterion`, `ProfitLossRatioCriterion` can be used standalone, but is required to calculate `ExpectancyCriterion`.

- The `NumberOfConsecutiveWinningPositionsCriterion` calculates the maximum number of consecutive winnings within a trading record. (It's a long name, maybe rename to ConsecutiveWinningPositionsCriterion?)

PR #727, opened 2021-04-18
Adding system fitness metrics

petrmac:
Adding some fitness metrics (for backtest purposes) plus some utility functions.
Please have a look. 
Help with Std deviation of PNL welcome (for SQN).

- added AverageLossCriterion
- added AverageProfitLossCriterion
- added AverageWinCriterion
- added ExpectancyCriterion
- added LosingPositionsRatioCriterion
- added RewardToRiskCriterion


nimo23:
Hi, thanks for your commit.

There is some overlap with your commit and https://github.com/ta4j/ta4j/pull/704. I would suggest to look at first https://github.com/ta4j/ta4j/pull/704 so we can decide to choose the better implementation.

Btw:
- Please rename `AverageWinCriterion` to `AverageProfitCriterion`.
- And maybe put those NumUtilities (`NumAverager.class`, `NumCollector.class`) to `utils/NumUtils.class` (if possible).

petrmac:
Thanks for answer, 
I am currently working on SQN, hence this is still a draft.
I'll move classes around and rename as requested... I haven strong opinion about location...

petrmac:
@nimo23, pls have a look again. Renamed what was requested.

I have some issues with formatter plugin. Especially with stream pipe, it makes the code messy. I prefer to have each stream operation on single line - it is more readable. 
Also I have itching, that license does not start with `/**` but only with `/*` - Intelli is reporting many errors with dangling Javadoc comment.

petrmac:
Maybe one question: 

shall we have ExpectancyCriterion as well as AverageProfitLossCriterion? They have almost same meaning, with only exception -> how the breakevens are handled (Expectancy filters them out).

Any ideas?

nimo23:
>Especially with stream pipe, it makes the code messy. I prefer to have each stream operation on single line - it is more readable.

We should also format each stream operation on a single line. I improved our code-formatter (see https://github.com/ta4j/ta4j/pull/728).

>Also I have itching, that license does not start with /** but only with /* - Intelli is reporting many errors with dangling Javadoc comment.

I don't know about this. Maybe @team172011 uses IntelliJ..

nimo23:
>shall we have ExpectancyCriterion as well as AverageProfitLossCriterion

I would prefer `ExpectancyCriterion`. But I don't mind to put `AvgProfitLossCriterion` also. Note that I also added this indicator in https://github.com/ta4j/ta4j/pull/704 which is implemented a little different. Please look at those PR. We should choose the better one..

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/AverageLossCriterion.java
```diff
@@ -0,0 +1,61 @@
...
+import org.ta4j.core.BarSeries;
+import org.ta4j.core.Position;
+import org.ta4j.core.TradingRecord;
+import org.ta4j.core.num.Num;
+import org.ta4j.core.utils.NumUtils;
+
+/**
+ * This is simple average loss of all losing positions
+ */
+public class AverageLossCriterion extends AbstractAnalysisCriterion {
```
nimo23, line 37: There exists another version: https://github.com/ta4j/ta4j/pull/704/files. We should choose the better one..

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/AverageLossCriterion.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private boolean isLosingPosition(BarSeries series, Position position) {
```
nimo23: For that, we have `NumberOfLosingPositionsCriterion`. No need to repeat code..

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/AverageProfitCriterion.java
```diff
@@ -0,0 +1,60 @@
...
+import org.ta4j.core.BarSeries;
+import org.ta4j.core.Position;
+import org.ta4j.core.TradingRecord;
+import org.ta4j.core.num.Num;
+import org.ta4j.core.utils.NumUtils;
+
+/**
+ * This is simple average profit of all winning positions
+ */
+public class AverageProfitCriterion extends AbstractAnalysisCriterion {
```
nimo23, line 37: There exists another version in https://github.com/ta4j/ta4j/pull/704/files. We should choose the better one..

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/AverageProfitCriterion.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private boolean isWinningPosition(BarSeries series, Position position) {
```
nimo23: Use `NumberOfWinningPositionsCriterion` instead of duplicating code.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/ExpectancyCriterion.java
```diff
@@ -0,0 +1,63 @@
...
+import org.ta4j.core.num.Num;
+
+/**
+ * Expectancy is the expected return of the strategy. The formula is average win
+ * * win % - average loss * loss %
+ *
+ */
+public class ExpectancyCriterion extends AbstractAnalysisCriterion {
+
+    private final AverageProfitCriterion averageWin = new AverageProfitCriterion();
```
nimo23, line 39: rename variables to `avgProfit` and `avgLoss`

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/LosingPositionsRatioCriterion.java
```diff
@@ -0,0 +1,61 @@
...
+import org.ta4j.core.Position;
+import org.ta4j.core.TradingRecord;
+import org.ta4j.core.num.Num;
+
+/**
+ * Calculates the percentage of positions which are profitable.
+ *
+ * Defined as <code># of winning positions / total # of positions</code>.
+ */
+public class LosingPositionsRatioCriterion extends AbstractAnalysisCriterion {
```
nimo23, line 36: There is already `LosingPositionsRatioCriterion` in https://github.com/ta4j/ta4j/pull/695. No need to duplicate. We should choose the better one..

nimo23 (commented):
I did not look this whole PR. At first, we should try to eliminate duplicate code between those different PRs..after that, we can have a depper look..

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/AverageProfitCriterion.java
petrmac: Not sure I understand what you mean. I checked the '_NumberOfWinningPositionsCriterion_' and it counts the number of positions that are profitable. The '_AverageProfitCriterion_' is counting **Average** of the profitable positions: 
**SUM of all profits / number of profitable positions**.
It does it in one go in the stream processing. 
If I included the other criterion, and used it to calculate the count, it would have been slower (the count is processed extra and then the average). 
But we can make a micro benchmark to check... although I strongly believe, my code would perform better than adding criterion call:) But I could be also mistaken... 

I also find the stream processing more readable... but this is matter of taste.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/AverageLossCriterion.java
petrmac: same comment as below... BUT I think we might actually add this method to the Position -> then tehre would be no duplication needed.

petrmac: something like `position.isProfitable` that return boolean. How about that?

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/LosingPositionsRatioCriterion.java
petrmac: @nimo23, I have updated the code a bit to demonstrate what I meant in my comments. Pls have a look. I think this is now better...

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/AverageLossCriterion.java
petrmac: Added the code in _Position_

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/AverageProfitCriterion.java
petrmac: We are both obviously biased:) @team172011 , could you please also look? Thx

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/AverageLossCriterion.java
nimo23: I included `Position#isWinning()` and `Position#isLoosing()` in https://github.com/ta4j/ta4j/pull/704 which can be used without the need for instantiating new criterions..

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/AverageProfitCriterion.java
nimo23: Yes, I included Position#isWinning() and Position#isLoosing() in https://github.com/ta4j/ta4j/pull/704 which can be used here..

nimo23:
@team172011 Long story short: I think, we should at first merge https://github.com/ta4j/ta4j/pull/704 and https://github.com/ta4j/ta4j/pull/695 and after that, I recommend that @petrmac creates a new PR with the improvements (if any) from this draft. This PR duplicates lots of logic which is already done in PR 704 and 695. This draft also contains some improvements over the other PR's but it's far easier to have discuss those improvements after the other's are merged.

petrmac:
Agree, Initially I didn't know about overlap. 
We should join forces and focus on what is missing - SQN. The goal is to have the Criterions merged...

team172011:
merged #704 next one would be #728

nimo23:
I would suggest to create a whole new PR (`Add Criterion: SQN`) and close this PR.

nimo23:
@team172011 Please close this issue in favour of https://github.com/ta4j/ta4j/pull/768.

Issue #731, opened 2021-04-21
numOf(constant) inside calculate()


johnthetireless:
I could be wrong, but I really don't think calling numOf(constant) inside the calculate() method is a good idea.

Here is what I think happens something like numOf(0) is used inside the calculate(index) loop.
0 is an int, so it is boxed into a Number 
AbstractIndicator<T> calls barSeries.numOf()  						
Either PrecisionNum::valueOf or DoubleNum::valueOf is called.  I think PrecisionNum is the default
After after some redirection through PrecisionNum and BigDecimal constructors, this very ugly looking constructor is called:

public BigDecimal(char[] in, int offset, int len, MathContext mc)

I am sure this is well optimized and all. but its ~200 lines of code.

DoubleNum::valueOf eventually calls an equally scary-looking static method:

FloatingDecimal::parseDouble(String s) 

So I suggest this kind of thing is better.  Some of the indicator classes do this already, but a lot don't.

private final Num zero = numOf(0);   /// autobox and convert to Num once; use in calculate loop

Here is a partial list of classes with what I deem an abusive call to numOf...

AroonDownIndicator  				numOf(barCount)   // numBarCount can be fixed in constructor 
AroonUpIndicator   				numOf(barCount)
CCIIndicator						numOf(0)
CMOIndicator						numOf(100)
DoubleEMAIndicator				numOf(2)
FisherIndicator					numOf(1)
MassIndexIndicator					numOf(0)
PPOIndicator						numOf(100)
RAVIIndicator						numOf(100)
ROCIndicator						numOf(100)
RSIIndicator						several 0, 1, 100
StochasticOscillatorKIndicator		numOf(100)
TripleEMAIndicator					numOf(3)
UlcerIndexIndicator					numOf(100)
ZLEMAIndicator					several 1, 2

MinusDIIndicator					numOf(100)
PlusDIIndicator					numOf(100)
DXIndicator						numOf(100)    -- unrelated to current topic, but this class causes a package dependency cycle; suggest move to adx package

GainIndicator						numOf(0)
LossIndicator						numOf(0)
MedianPriceIndicator				numOf(2)
MinusDMIndicator					numOf(0)
PlusDMIndicator					numOf(0)
TypicalPriceIndicator				numOf(3)

AbstractIchimokuLineIndicator		numOf(2)	-- unrelated, but please fix the broken ChikouSpan; as reported elsewhere it projects N periods into the future rather than the past
[...]

PR #734, opened 2021-04-23
Add common static Num's

nimo23:
- provide common static Num's for `DecimalNum` and `DoubleNum`
- provide access to those Num's in `Num`-Interface


nimo23:
If this PR is taken, then I will add those **static** Num's to all other those needed classes, e.g. the classes described in https://github.com/ta4j/ta4j/issues/731.

The static access may be 
a) faster (we save the time for initializiation)
b) more memory friendly (because only one instance of a static Num is created)

nimo23:
There are 3 alternatives  to the above implementation:

1) move all those static Nums (from `DoubleNum.class`, `DecimalNum.class`, `Num.class`) to enum `NumOf.class` (so that `DoubleNum.class`, `DecimalNum.class`, `Num.class` are unchanged). All static Nums are solely accessed via `NumOf.class`.

2) move the methods from `NumOf.class` to `Num.class` and **remove** `NumOf.class`. All static Nums are accessed via `Num.class` (and `DoubleNum.class`and `DecimalNum.class`).

3) use `NumOf.class` and provide static shortcut methods within `Num.class` (this is the current implemetation as defined in this initial PR). 

Which way should I choose?

nimo23:
Another alternative is to move all methods from `NumOf.class` to `Num.class` and delete `NumOf.class` so that all access to the static Num's can be done through `Num.class` (for example, `Num.HUNDRED(otherNum);`) 

Would it be better to remove `NumOf.class` and move the static methods into `Num.class`?

team172011 (changes requested):
This hardcoded `NumOf` methods depend on the currently available `Num` implementations. If the user would try to use his own/another implementation of `Num` everything calculation would break because `NumOf`... would always return `NaN`.

team172011:
> Would it be better to remove NumOf.class and move the static methods into Num.class?

Yes, because if we want to cache common values we have to do it in a way that it is not bound to the current Num implementations but can be enhanced when someone tries to add his own implementation of Num

johnthetireless:
I am thinking of making all the changes identified in the list of "numOf() offenders" I made for #731   This should use whatever you come up with the end.

Let me know when you have finalized a mechanism the indicators should use and I will fix the ones I listed.

nimo23:
Ok, now it's perfect. I added 

-`Num#zero`
-`Num#one`
-`Num#hundred`

which obtains its value by a static field.

@team172011 any other suggestions? If not, I will go further and add those access to all dedicated classes. Actually, I did the improvement only for `CCIIndicator`.

CHANGELOG.md
```diff
@@ -89,6 +89,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - :tada: **Enhancement** added Position#hasProfit.
 - :tada: **Enhancement** added Position#hasLoss.
 - :tada: **Enhancement** exposed both EMAs in MACD indicator
+- :tada: **Enhancement** added Num#zero, Num#one, Num#hundred.
```
team172011: please update

ta4j-core/src/main/java/org/ta4j/core/num/Num.java
```diff
@@ -41,6 +41,21 @@
...
+
+    /**
+     * @return the Num of 1
+     */
+    Num one();
+
+    /**
+     * @return the Num of 100
+     */
+    public Num hundred();
```
team172011: Please remove `public` modifier from interface.


we could add a `default` implementation to those methods like

```java
default Num hundred(){
    valueOf(100);
}
```

so once is not forced to override this methods and the implementation would still work. Using this defaults we could remove the implementation from `NaN` class

ta4j-core/src/main/java/org/ta4j/core/num/Num.java
```diff
@@ -41,6 +41,21 @@
  */
 public interface Num extends Comparable<Num>, Serializable {
 
+    /**
+     * @return the Num of 0
+     */
+    Num zero();
```
team172011: we could enhance javadoc to explain that this is used for caching common used constants

nimo23: The javadoc for interface should not contain implementation details. But if you want you can add a hint.

nimo23:
Ok, I am done. 

After this PR is merged, another PR can adapt all sources (indicators, etc) where those static Num's can be used.  See the classes defined in https://github.com/ta4j/ta4j/issues/731#issue-863393765.

I will not add those changes in this PR. I recently have not much time. 

@team172011 or  @johnthetireless  Maybe you  can create a new PR and add those static nums to all those needing classes.

team172011 (approved):

Issue #608, opened 2020-06-22
CachedIndicator should be a wrapper indicator not in inheritance structure


GxtPerfectoM:
Currently it looks like the CachedIndicator is causing problems in multi-thread mode.
In general you should not use the CachedIndicator in inheritance structure but instead offer it as a wrapper to other indicators (Delegator) and this will give the way to avoid using it if needed.

nimo23:
@GxtPerfectoM would be nice if you can make a PR with your suggestion.

johnthetireless:
Here is a naive "caching decorator" around Indicator<Num>.
You can adapt for <T> and change the caching policy as you see fit.

public interface NumericIndicator extends Indicator<Num> {
}

public class CachedNumericIndicator implements NumericIndicator {  

	private final List<Num> cache = new ArrayList<>();
	
	public CachedNumericIndicator(Indicator<Num> delegate) {
		super(delegate);
	}

	/*
	 * Simplistic implementation will probably suffice for most use cases.
	 * Replace or override if desired
	 */
	@Override
	public Num getValue(int index) {
		if ( index < cache.size() ) {
			return cache.get(index);
		}
		else {
			Num result = delegate.getValue(index);  
			cache.add(result);
			return result;
		}
	}
	
	@Override
	public Num numOf(Number number) {
		return delegate.numOf(number);
	}
	
	@Override
	public BarSeries getBarSeries() {
		return delegate.getBarSeries();
	}

}

johnthetireless:
Generic types didn't copy/paste correctly.

johnthetireless:
Caching 

There are 6-7 "helper" indicators like MedianPrice whose calculation is based solely of the price/volume data of a single bar. Caching the result of Open(i) + Close(i) / 2 seems questionable.

A larger set of higher-level indicators do the same, but they do arithmetic on values obtained from "child" indicators.
PlusDI(n) = MMA)n) / ATR(n) * 100, for example. I would be inclined to let this calculate each time without being cached.

I would really like to see caching removed from the Difference, Sum and other "arithmetic" indicators. They don't seem to be used much, but series arithmetic is a powerful concept. A(i) = B(i) + C(i) is very easy to code as an operation of 2 series.

nimo23:
See https://github.com/ta4j/ta4j/issues/522#issuecomment-601354192

nimo23:
[...]

PR #735, opened 2021-04-24
ADX Constructors

johnthetireless:
Fixes:

Fixed numOf() inside calculate() for ADX related indicators using existing final instance field pattern.

Added constructors so ADX and related objects can be created and used with substantially fewer "child" indicators.  Reduces the number of array lists (27 to 12, by my count) if ADX and +/-DI are used together in a strategy.  Might improve processing a little.  These constructors will create 1 instance of each object, rather than separate objects that end up calculating and caching the same numbers.

Moved DX to adx package.  The ADX constructor is the only place DX is referenced.  

+DM and -DM were also moved to adx package, only being referenced by +/-DI respectively.  There is a small chance these 3 moves will break import statements in 0.13 user code, but it seems unlikely that people would instantiate these objects independently.

Added getBarCount() and toString to ATR.


nimo23:
>Did not move +/- DM to adx package yet, although that is where they belong. Left them alone to ensure tests don't break.

We should add those +/- DM to adx package.

johnthetireless:
A little diagram of the ADX subsystem.  I have a nice modeling tool.  I reverse-engineered the develop branch and created this diagram with some drag-and-drop.  Comes out very nicely, IMO.

[ADX.pdf](https://github.com/ta4j/ta4j/files/6370455/ADX.pdf)

johnthetireless:
Take a look at the way ADX and ATR delegate calculation to their respective MMA objects.  If I have untangled the caching mechanism correctly, ADX and ATR (both cached) are going to maintain array lists with exact copies of what their MMAs have calculated.  The MMAs are cached; recursively so, in fact...  :)   The ATR and ADX do not need to be cached.

To keep my PRs organized and consumable, I will not include additional code changes for this perceived useless caching.  If/when I do submit a PR with such changes, there are 2 design choices I can see:

1) Make ATR and ADX subclasses of MMA.  The names and definitions of these guys "average true range" and "average DX" strongly implies inheritance to me.  ATR(14) = MMA(TR,14)

2) Make ATR and ADX subclasses of AbstractIndicator, OR have them simply implement the interface and delegate as they do now.

Either way, we can eliminate a few more array lists from the execution model.  I have no preference and can easily implement non-cached versions of these guys with either approach.  Your call, guys.

There are a few other classes that do the "delegate and cache anyway" thing.  I will change them all once we decide on a subclass vs delegate approach.    Perhaps the semantics and definitions of some of these do not support inheritance the way ATR and ADX do; we can use a mix of subclassing and  whatever other non-caching approach is deemed best.

BB Middle ("is the SMA of price")
Keltner middle ("is the EMA of price")
MVWAP (sounds like an average to me)
Coppock
HMA
Stoch %D

nimo23:
Btw, the caching layer may be pluggable in the future as described in https://github.com/ta4j/ta4j/issues/608 (I can imagine that a child inheriting a caching indicator inherits also the caching layer - but this is not always wanted, therefore childs should be able to opt out from the caching layer in such cases).

Keeping inheritance structure as simple as possible, I would tend to choose:

> 2. Make ATR and ADX subclasses of AbstractIndicator, OR have them simply implement the interface and delegate as they do now.

johnthetireless:
OK  When I make a PR to modify the "delegate and subclass anyway" indicators I will subclass AbstractIndicator.  I will take care to leave the same constructor interfaces in place.  I will also include a simple proposal for a caching wrapper.  The group will need to look at that.

On the matter of caching... why are the 5 price and volume indicators cached?  These guys are as close to the bar series in terms of computation as it gets.  The extra array lists are a waste.  (Yes, I am obsessed with the array lists.  Much of my work was for embedded systems.  Easily fixable resource waste simply bothers me.)  The path through the caching mechanism can only slow things down.  99% of the time, the close price indicator will read a value from the bar series and copy it to an array list.  Checking the "cache" later is going to be slower than simply applying the getClosePrice selector.

Caching the price indicators is an issues because a lot of others create them as children.  If I create the full Ichimoku cloud, for example (it is very cool looking), the current constructors create 2 (maybe 3) instances of both high and low price indicators.  This bothers me.  I assume thread safety and support for smaller devices is something you guys have thought about.  Some rationalization of the existing indicators is required before I would even think about that.   

I will uncache these price/volume too for the next PR, unless you can give a reason why its a bad idea.

nimo23:
>I will uncache these price/volume too for the next PR, unless you can give a reason why its a bad idea.

Yes, good point. Do a PR so we can look at it in detail. 

In general, if any indicator (e.g. `ClosePriceIndicator`) only retrieves values from its `BarSeries`, then it does not make sense to cache this indicator. But be aware, that the `BaseBarSeries` can remove exceeded bars (if `barSeries.maximumBarCount` is set). Maybe, for example, the cachable `ClosePriceIndicator` can have more close price values in the cache as there exists in its truncated BarSeries. If this is not the case then we should remove the caching layer from those (price and volume indicators) indicators ..

johnthetireless:
Understood.  I need to look at the bar removal code more closely.

Perhaps it will be useful if we distinguish a couple of concepts.  To most people (me) "caching" means storing a value so it doesn't need to calculated or fetched again.  Some indicators (OBV, accum/dist) must store their running totals or we can never do getValue(index-1) sensibly.  I will invent the term "calculation cache" for this.  Some others (moving averages) arguably expend enough effort in calculation that we want to store the results rather than recalculating them.  This matches my basic view of "caching".  Not absolutely necessary to avoid recursion, but probably a good idea.
The idea that an indicator should be smart enough to handle a bar series of changing length seems largely orthogonal to the 2 types of caching I think of.  These are "managed" indicators or something else.  I expect all indicators should have these smarts.  But they don't all need array lists.

team172011:
Yes, I also agree that the caching of price/volume indicators is useless, good point. And I also agree and would prefer the second design approach

>  Make ATR and ADX subclasses of AbstractIndicator, OR have them simply implement the interface and delegate as they do now.

keeping things as simple as possible.

team172011:
> Maybe, for example, the cachable ClosePriceIndicator can have more close price values in the cache as there exists in its truncated BarSeries.

The current caching logic respects this and will remove exceeding results from the array cache

team172011:
Looking at the `PriceIndicator` I think it is over-engineered. Just for calculating the close price you have an additional inheritance and an additional instance field holding the `priceFunction`.

ta4j-core/src/main/java/org/ta4j/core/indicators/adx/ADXIndicator.java
```diff
@@ -37,26 +37,65 @@
...
+        this.adxBarCount = adxBarCount;
+        this.dx = new DXIndicator(atr);
+        this.averageDXIndicator = new MMAIndicator(dx, adxBarCount);    	
+    }
 
+    /*
+     * This is obviously pure delegation.  
+     * ADX is caching the exact same values needlessly, I believe.
+     * ATR and 5-6 others do this too.
+     */
```
team172011: Yes, this should be discussed and/or changed. But this comment should not be part of this lib, please remove.

team172011 (changes requested):
Please add entries to the CHANGELOG.md

ta4j-core/src/main/java/org/ta4j/core/indicators/adx/ADXIndicator.java
johnthetireless: Yes..  those non-javadoc comment were intended for reviewers.  I will remove them.

I am not sure what you mean by adding to the change log.

johnthetireless: > Looking at the `PriceIndicator` I think it is over-engineered. Just for _calculating_ the close price you have an additional inheritance and an additional instance field holding the `priceFunction`.

Agree.  The price and volume classes could have been done as static fields or enums or something.  There are a few others too.  Typical price, median price... maybe as many as 10 classes could have been represented a simple lambdas like  b -> b.getClose() + b.getOpen() / 2... etc.  Coding Nums in lamdas is messy, but not cause for separate classes, IMO.  More classes to maintain...  more files...  oh well

johnthetireless:
I will add one more comment/suggestion about caching...

As implied above, there are 6-7 "helper" indicators like MedianPrice whose calculation is based solely of the price/volume data of a single bar.  Caching the result of Open(i) + Close(i) / 2 seems questionable.

A larger set of higher-level indicators do the same, but they do arithmetic on values obtained from "child" indicators. 
PlusDI(n) = MMA)n) / ATR(n) * 100, for example.  I would be inclined to let this calculate each time without being cached.

I would really like to see caching removed from the Difference, Sum and other "arithmetic" indicators.  They don't seem to be used much, but series arithmetic is a powerful concept.  A(i) = B(i) + C(i) is very easy to code as an operation of 2 series.  

I have a wrapper class called NumericIndicator, for example.  I also have simple proxies to replace DifferenceIndicator, Sum, etc.

I create these things like pure function objects.  I could code stuff like this:

plusDI = mma.dividedBy(atr).mulipliedBy(oneHundred)  // indicatorarithemtic;  mma and atr are Indicator<Num>
// plusDI is a "pseudo" indicator (generator function/recurrence relation; I call them "Series")

plusDI.getValue(index);    /// the arithmetic operations on values at index do not happen until this point

It would be nice if I didn't need uncached proxies for difference and sum.

nimo23:
> Would you consider a super type of Indicator?

Give it a try. Please provide a draft PR so we can look at it in detail..it must not be perfect, it should only give us a first impression what you mean in detail.

>I will add one more comment/suggestion about caching...

Would be nice to forward the communication of all those caching ideas to https://github.com/ta4j/ta4j/issues/608..

johnthetireless:
> Give it a try. Please provide a draft PR so we can look at it in detail..it must not be perfect, it should only give us a first impression what you mean in detail.
> 
I am going to post some of the NumericIndicator stuff I mentioned to the examples project so you can get a better idea before we change Indicator<T>.  You can just have a look and play with it a bit without merging.  If you like any or all of it, we can include it in the examples.  I'm old.  I don't care.  :)
> 
> Would be nice to forward the communication of all those caching ideas to #608..
The relevant parts have been copied there.

Can I ask you about the changelog?  Am I supposed to prepare or edit a file for PRs?

nimo23:
> Can I ask you about the changelog? Am I supposed to prepare or edit a file for PRs?

Normally, any PR should include a _short summary_ in https://github.com/ta4j/ta4j/blob/develop/CHANGELOG.md. You only need to edit that file (with a entry of your change) when making a PR. So for example: with this PR, you can include a sentence in the changelog-section **Changed** or **Breaking**.

johnthetireless:
Rats..  I added a bunch of code for the examples project, thinking it would create a separate PR.  If you can merge just the core changes that would be great.  The examples code is meant for review and discussion.  Sorry about this.

johnthetireless:
One more newbie question...  how can I get my johnthe...  repo to stay in sync with ta4j?  For now I just want to edit the latest changelog.  Going forward I will want approved changes to upstream synced automatically.    

I have used corporate github-like things, but this is different.

nimo23:
I find the functional approach nice and this is definiitly a way to consider if or how we can include it. I will look at it deeper in the next time. 

In general, please add a **new PR** (or a new draft PR) each time you publish a new distinct idea. Currently, the `functional approach commits` are polluting this `ADX Constructors-commits` hindering/slowing down the merge of this PR.


>One more newbie question... how can I get my johnthe... repo to stay in sync with ta4j

You need to `Fetch from origin` into your ta4j-branch to stay in sync. Maybe @team172011 can provide more help with that.

johnthetireless:
> and remove these commits from _this_ PR.

I moved the files locally and committed again.  They appear as deleted now.

ta4j-core/src/main/java/org/ta4j/core/indicators/adx/ADXIndicator.java
johnthetireless: @team172011   

When you have a moment, I could use a little help,  I made a couple of rookie mistakes with commits yesterday and I would like to avoid doing it again.  I am new to this environment and the set of tools I am using may not be appropriate.
Here is what I have done:

- fork ta4j/ta4j into john../ta4j (6-7 days ago)
- clone to my local machine; prompted to install and use Github desktop, which is fine
- switch branches locally; my file system matches what I see on the server
- open eclipse; nice and seamless
- change code
- GitHub desktop sees changes and its easy to commit locally and push
- create pull request using one of these web pages

So far OK?  
 Questions
- how do I fetch from main to my repo;  my changelog file is out of date and I want to edit it
- how do I  automate this, so approved changes that go into main are pushed to my repo
- how do I resolve conflicts?  I must need a separate tool, as the web pages don't seem to give me a change to do anything but look; I will try to find a merge utility
- - maybe I can and should do all this through Eclipse; if you have tips for that I would appreciate it

thanks

nimo23: >maybe I can and should do all this through Eclipse; if you have tips for that I would appreciate it

Yes, definitly. You don't need to use Github Desktop. It's far easier to handle all of that within Eclipse. All your questions can be answered simply by using Git within Eclipse IDE. I recommend to read the official documentation: https://wiki.eclipse.org/EGit/User_Guide. The workflow is simple: 

1. Fetch Repo within Eclipse by using the View "Git Repository"
2. Import fetched Repo into Eclipse "Package Explorer"
3. make changes
4. Push or fetch changes from Eclipse (right click on "Package Explorer">"Team"-Menu)

Read 
- https://dzone.com/articles/tutorial-git-with-eclipse#
- https://www.baeldung.com/egit 
- or any guide you find by google with the keyword "_Eclipse EGit Tutorial_".

johnthetireless: Thank you very much.  I should have read some of this before asking questions.  I have made some progress.  I think the file set for this PR is what it should be.  2 files appear as conflicts.  I think they need to be approved and then merged.

nimo23: > I think they need to be approved and then merged.

No, conflicts means that **you** must resolve the conflict by yourself. Please edit

- CHANGELOG.md
- ta4j-examples/pom.xml 

so that the conflicts are resolved.

johnthetireless: I am afraid I am missing something.  I fetched the latest version of the changelog; I edited and committed it.  My changes were strictive additive.  I don't understand how this is causing a conflict.

The examples/pom.xml was changed after the 0.14 release, it seems.  I mistaken checked in my local pom with old version numbers.  I have fixed this and committed again.  

I'm afraid I do not know what git wants me to do

nimo23: Try to replace the whole `pom.xml` of your PR with our current version of the **`develop`**-branch.

ta4j-examples/pom.xml
```diff
@@ -6,7 +6,7 @@
         <version>0.15-SNAPSHOT</version>
     </parent>
     <artifactId>ta4j-examples</artifactId>
-
+
```
johnthetireless: black line with spaces?

ta4j-examples/pom.xml
```diff
@@ -1,48 +1,210 @@
...
-    <name>Ta4j Examples</name>
-    <description>Provides examples that demonstrate how to use ta4j</description>
-
-    <dependencies>
-        
-        <!-- Test dependencies -->
-        <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.13.1</version>
```
johnthetireless: copy

ta4j-core/src/main/java/org/ta4j/core/indicators/adx/ADXIndicator.java
johnthetireless: I have done that twice now.  Something is slightly different that I can't see.  I will look at it again tomorrow

johnthetireless:
> I find the functional approach nice and this is definiitly a way to consider if or how we can include it. 

I just did a little scan using reflection.  The 0.14 release has 250+ concrete indicator classes; a couple might be anonymous inners.  0.13 had about 120, I haven't looked the new classes, but I suspect the implementation of a at least a few of them would have been simpler and clearer (or unnecessary)  using some functional approaches..  I don't see a clean fit for the kind of code I showed at this point - most of it is deliberately coded as a layer above ta4j,, but it could be adapted and might be something to consider for the future.  250 classes is a lot to maintain and track.

Couple of quick questions:  to produce a PR for what you have aptly called "functional approaches" I should probably create a feature branch.  I would do that in master first?  Is that the way you have been doing things?  Is a PR from a feature branch what you call a draft?

nimo23:
> to produce a PR for what you have aptly called "functional approaches" I should probably create a feature branch.

You must **ALWAYS** use the `develop`-branch. You should never use the `master`-branch. 
The develop-branch is the branch from which you create a PR (includes draft/feature branch).

> Is a PR from a feature branch what you call a draft?

Yes. You should create a future branch (or any other kind of PR) always from the `develop`-branch.

ta4j-core/src/main/java/org/ta4j/core/indicators/adx/ADXIndicator.java
nimo23: >Something is slightly different that I can't see.

Another alternative quick solution would be: 
1. close this PR
2. create a new branch "ADX Constructors" from the `develop`-branch. 
3. Adapt your changes (don't forget to include the changes in `changelog.md`). 
4. Submit PR.

Then it should work.

ta4j-examples/pom.xml
nimo23: Please remove the blank line.

ta4j-examples/pom.xml
nimo23: btw, normally you do not need to edit `pom.xml`.

johnthetireless: I did not edit the examples/pom; at least not deliberately.  It and the changelog were modified for the 0.14 release.  

BTW, the new examples POM doesn't work as it is - at least not for me.  It cannot find core 0.15-SNAPSHOT; I think there was an issue with the new junit version too.  Perhaps your build is inserting the SNAPSHOT version for development and the released one is fine.  But as it is, I will need to edit the POM to make it work and be careful not to commit it.  I can adapt.

johnthetireless:
I may resubmit this.  I may not   Good luck

PR #739, opened 2021-04-26
Num: remove serializable

nimo23:
- remove `Serializable` in `Num.class`

It does not make any sense to have `Serializable` in Num (and its dedicated implementations). It's the same as if I would add Serializable to BigDecimal (which does also not make any sense). Thus, remove it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **PrecisionNum**: improve performance for methods isZero/isPositive/isPositiveOrZero/isNegative/isNegativeOrZero.
 - **BaseTimeSeriesBuilder** moved from inner class to own class
 - **TrailingStopLossRule** added ability to look back the last x bars for calculating the trailing stop loss
+- **Num** removed Serializable
 
 ### Added
 - **Enhancement** Added getters for AroonDownIndicator and AroonUpIndicator in AroonOscillatorIndicator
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -129,6 +129,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **PrecisionNum**: improve performance for methods isZero/isPositive/isPositiveOrZero/isNegative/isNegativeOrZero.
 - **BaseTimeSeriesBuilder** moved from inner class to own class
 - **TrailingStopLossRule** added ability to look back the last x bars for calculating the trailing stop loss
+- **Num** removed Serializable
```
team172011: this is the wrong release

ta4j-core/src/main/java/org/ta4j/core/indicators/DistanceFromMAIndicator.java
```diff
@@ -23,15 +23,14 @@
...
-
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
 
+import org.ta4j.core.Bar;
+import org.ta4j.core.BarSeries;
+import org.ta4j.core.num.Num;
+
```
team172011, line 33: is all this swapping of imports done by your IDE or by the formatter plugin?

nimo23: I use Eclipse and by using `Source->Organize Imports` you can improve the ordering of the import statements. This is done automatically and it's recommended. I don't know how to do that with formatter plugin.

team172011 (approved):

dazre:
It was hurried decision. In my trading platform build on ta4j I uses message queue driver to reduce code coupling and prepare system to split onto independent modules. After upgrade ta4j to 0.15 version i have got  an error on the BaseBar serialization so I had to downgrade back to v.0.14.

PR #743, opened 2021-05-02
Feature/combine indicator

MarcDahlem:
Fixes #.

- Introduced a CombineIndicator, similar to the TransformIndicator, which allows to combine two Indicator values with a given combine-function
- Abstracted the DifferenceIndicator to the more general CombineIndicator
- Marked DifferenceIndicator as deprecated


ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DifferenceIndicator.java
```diff
@@ -34,24 +34,23 @@
  */
 public class DifferenceIndicator extends CachedIndicator<Num> {
```
team172011: This indicator would get redundant and can be removed/replaced

team172011 (commented):
Thanks, this indicator adds missing functionality for operations between two indicators and can replace the `DifferenceIndicator`

MarcDahlem:
Thx for reviewing.
Did I get it right, that I should remove the DifferenceIndicator completely instead of marking it as deprecated?

Two more general questions:

1.  The implementation of the TransformIndicator(indicator, coefficient) could be replaced with CombineIndicator(indicator, new ConstantIndicator(coefficient)).
  a) this would increase CodeReuse.
  b) But I'm not sure if the caching behavior of the ConstantIndicator harms the speed
2. I guess some in-place instantiations of temporary indicators or math operations could be replaced as well, like https://github.com/ta4j/ta4j/blob/develop/ta4j-core/src/main/java/org/ta4j/core/indicators/DoubleEMAIndicator.java#L58, https://github.com/ta4j/ta4j/blob/develop/ta4j-core/src/main/java/org/ta4j/core/rules/TrailingStopLossRule.java#L124 or https://github.com/ta4j/ta4j/blob/master/ta4j-core/src/main/java/org/ta4j/core/indicators/pivotpoints/DeMarkPivotPointIndicator.java#L101
  a) this could speedup some calculations as well

What do you mean? Should we do it?

nimo23:
>Did I get it right, that I should remove the DifferenceIndicator completely instead of marking it as deprecated

yes, remove it and add this change in changelog (remove-part)

>The implementation of the TransformIndicator(indicator, coefficient) could be replaced with CombineIndicator(indicator, new ConstantIndicator(coefficient)).

No, leave it as it is. The TransformIndicator is not the same as the CombineIndicator (even the name does not fit). Btw, everything you can do with the CombineIndicatorr can be done with the TransformIndicator, but not vice versa. 

> I guess some in-place instantiations of temporary indicators or math operations could be replaced as well,

Please do not put other issues into an PR. Open another dedicated issue or PR to discuss those other topics.

nimo23:
>The implementation of the TransformIndicator(indicator, coefficient) could be replaced with CombineIndicator(indicator, new ConstantIndicator(coefficient)).

The `TransformIndicator` is not the same as the `CombineIndicator` (even the name does not fit). We could replace it but I am afraid of wrapping each `Num` into an Indicator only to do simple math operations..(and the name "Transform" fits good in what this class does..)

If we are to merge both indicators into one, then we should also name this indicator `**MathIndicator.class**` (because what it does is doing math operations on two properties ("indicator<->indicator", or "indicator<->num"). The `CombineIndicator` is a little irritating..

MarcDahlem:
> We could replace it but I am afraid of wrapping each `Num` into an Indicator only to do simple math operations..
I had the same feeling, thats why I did not replace it so far.

> 
> If we are to merge both indicators into one, then we should also name this indicator **`MathIndicator.class`** (because what it does is doing math operations on two properties ("indicator<->indicator", or "indicator<->num"). The `CombineIndicator` is a little irritating..

Thats definitely true if we would put everything together in one class.
But if we stay seperated in two classes, do you also here prefer another name? So far, I could not come up with a better idea...

nimo23:
I think, we should do the following:

- rename this PR to _Feature/MathIndicator_
- rename `CombineIndicator.class` to `MathIndicator.class`
- remove `extends CachedIndicator<Num>` from `MathIndicator.class`
- use `extends AbstractIndicator<T>` in `MathIndicator.class`
- put all those static methods from `TransformIndicator.class` to `MathIndicator.class` (put all those test cases of `TransformIndicatorTest.class` into `MathIndicatorTest.class`

Those (static) methods from the `MathIndicator.class` are not cacheable at all.

**The only question is:** 
Does it hurt, to wrap a `Number` into a `Num` which is then wrapped again into a (non cachable) indicator (= ConstantIndicator). I guess, performance will not hurt in a way that is measurable..but we have to look..

@team172011  what do you think?

MarcDahlem:
Why would you remove the caching?

l = indicatorLeft.value(i) is cached, yes
r = indicatorRight.value(i) is also cached.

But the function application f(l,r) will then not be cached anymore.
This means, especially for expensive functions f, like f=mult or f=div, we would perform the same operation multiple times...

nimo23:
I thought that the `MathIndicator` is only used within another **cachable** indicator (which needs some MathOperations provided by MathIndicator), but yes: if we use it in isolation, then caching should also be possible.

team172011:
I think we should stay with having two different indicators, one for the (num value) transformation and one for operations between two (or even more) indicator values. Otherwise it will get too complicated and confusing for the user. 
Caching would make sense for both indicators because the user could provide a custom function of `BinaryOperator<Num>` or `UnaryOperator<Num>` that could be expensive.

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/DifferenceIndicator.java
MarcDahlem: I removed it completely, instead of marking it deprecated.

MarcDahlem:
Rework of the PR finished:
* I removed the DifferenceIndicator completely
* I moved the already available DifferenceIndicator-Tests to the CombineIndicator-Test file

Should now be ready to merge

team172011 (approved):

MarcDahlem:
Is this PR waiting for something? It got approved but is not merged yet.

Issue #742, opened 2021-05-02
Position-based indicators


MarcDahlem:
**I have checked existing issues and wiki**
- [ ] I could not find similar [issues](https://github.com/ta4j/ta4j/issues?utf8=%E2%9C%93&q=)
- [ ] I could not find a solution in the [wiki](https://ta4j.github.io/ta4j-wiki/) or [faq section](https://ta4j.github.io/ta4j-wiki/FAQ.html)

If possible provide a **[minimal working example](https://stackoverflow.com/help/mcve)**!

Hi,

I created a break-even indicator, which contains the value that must be reached to close a position with a positive net profit (regarding transaction costs).

![grafik](https://user-images.githubusercontent.com/2813850/116809869-bea9fc80-ab40-11eb-987b-d76620c2ae72.png)

So far, I constructed the indicator manually by executing a strategy and storing all buy and sell timepoints/index positions.
Is there a way to calculate such an position-based indicator live ?

i.e. can I get the already executed trades/ opened and closed positions inside the "protected Num calculate(int index) { " - Method ?

Here the code I have so far:
```

        BigDecimal buyFeeFactor = BigDecimal.ONE.add(buyFee);
        BigDecimal sellFeeFactor = BigDecimal.ONE.subtract(sellFee);
        TransformIndicator breakEvenCalculator = TransformIndicator.divide(TransformIndicator.multiply(new ClosePriceIndicator(series), buyFeeFactor), sellFeeFactor);
        
        
    @Override
    protected Num calculate(int index) {
        Integer lastBuyIndex = **sortedBuyIndeces**.floor(index);
        Integer lastSellIndex = **sortedSellIndeces**.floor(index);
        if (lastBuyIndex == null) {
            return NaN;
        }

        if (lastSellIndex == null || lastSellIndex <= lastBuyIndex || lastSellIndex == index) {
                return indicator.getValue(lastBuyIndex);
        }
        return NaN;
    }
```

The problem with this code is, that I have to collect all buy and sell points manually... Is there a better way that I didn't see so far?

nimo23:
In general, you can access all your trades solely by TradingRecords. 

I would suggest to create a new PR with your `PositionIndicator.class` (if this could be usable for others, too). So we can look at it in detail for any improvements..

MarcDahlem:
Solved with https://github.com/ta4j/ta4j/pull/757

Thx @nimo23  for your suggestion using Position or Trades in the indicator. That worked and the result is very powerful!

[...]

PR #747, opened 2021-05-04
Add: CriterionIndicator

nimo23:
- `CriterionIndicator.java`

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CriterionIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CriterionIndicator.java
@@ -1,0 +1,92 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2014-2017 Marc de Verdelhan, 2017-2021 Ta4j Organization & respective
+ * authors (see AUTHORS)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package org.ta4j.core.indicators.helpers;
+
+import org.ta4j.core.AnalysisCriterion;
+import org.ta4j.core.BarSeries;
+import org.ta4j.core.Position;
+import org.ta4j.core.TradingRecord;
+import org.ta4j.core.indicators.CachedIndicator;
+import org.ta4j.core.num.Num;
+
+/**
+ * Criterion indicator.
+ * 
+ * <p>
+ * Transforms any AnalysisCriterion into an Indicator. Returns <code>true</code>
+ * if the calculated criterion value on a bar index is better than the given
+ * {@link #criterionValue}, otherwise returns <code>false</code>.
+ */
+public final class CriterionIndicator extends CachedIndicator<Boolean> {
+
+    private final AnalysisCriterion criterion;
+    private final Num requiredCriterionValue;
+    private final TradingRecord tradingRecord;
+    private final Position position;
+
+    /**
+     * Constructor.
+     * 
+     * @param series                 the bar series
+     * @param tradingRecord          the trading record
+     * @param criterion              the criterion to get the calculated criterion
+     *                               value on a bar index
+     * @param requiredCriterionValue the required criterion value to test against
+     *                               the calculated criterion value
+     */
+    public CriterionIndicator(BarSeries series, TradingRecord tradingRecord, AnalysisCriterion criterion,
+            Num requiredCriterionValue) {
+        super(series);
+        this.tradingRecord = tradingRecord;
+        this.criterion = criterion;
+        this.requiredCriterionValue = requiredCriterionValue;
+        this.position = null;
+    }
+
+    /**
+     * Constructor.
+     * 
+     * @param series                 the bar series
+     * @param position               the position
+     * @param criterion              the criterion to get the calculated criterion
+     *                               value on a bar index
+     * @param requiredCriterionValue the required criterion value to test against
+     *                               the calculated criterion value
+     */
+    public CriterionIndicator(BarSeries series, Position position, AnalysisCriterion criterion,
+            Num requiredCriterionValue) {
+        super(series);
+        this.tradingRecord = null;
+        this.criterion = criterion;
+        this.requiredCriterionValue = requiredCriterionValue;
+        this.position = position;
+    }
+
+    @Override
+    protected Boolean calculate(int index) {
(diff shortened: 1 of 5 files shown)


nimo23:
@team172011 Can you please share your ideas about this CriterionIndicator.

- The `CriterionIndicator` should transform any `AnalysisCriterion` into an Indicator (which can then be used for rules). 
- The `CriterionIndicator` returns `true` if the calculated criterion value _on a bar index*_ is better than the `requiredCriterionValue`, otherwise it returns `false`.

**The current issue I have in mind is:**
The `CriterionIndicator`calculates its _criterion value_ either on the whole barSeries or on the single position (any `AnalysisCriterion` cannot calculate its value dependent to a specific bar index). 
For example: The CriterionIndicator returns true as long as the whole tradingRecord fullfills the `requiredCriterionValue`. I think, it is not wrong to ignore the _bar index*_ when calculating the criterion value within the `CriterionIndicator` - or what do you think?

ta4j-examples/src/test/java/ta4jexamples/loaders/JsonBarsSerializerTest.java
```diff
@@ -23,18 +23,20 @@
...
 import java.io.IOException;
 import java.math.BigDecimal;
 import java.util.concurrent.ThreadLocalRandom;
 
-import static org.junit.Assert.*;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.ta4j.core.Bar;
+import org.ta4j.core.BarSeries;
```
team172011: @nimo23 you do a lot of imports swapping in different files that are not relevant in the specific PR

nimo23: Are you suggesting that I do such import swappings in a new PR in the future (I think such import swappings is ike formatting a file and does not deserve a PR).

nimo23: In future, I will release a PR when doing big reformats (or import swappings). Sorry for this inconvenience.

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CriterionIndicator.java
```diff
@@ -0,0 +1,123 @@
...
+                // consider only those positions made till index
+                if (pos.getEntry().getIndex() <= index) {
+                    Trade entry = pos.getEntry();
+                    tradingRecordTillIndex.enter(entry.getIndex(), entry.getNetPrice(), entry.getAmount());
+                }
+                if (pos.getExit() != null && pos.getExit().getIndex() <= index) {
+                    Trade exit = pos.getExit();
+                    tradingRecordTillIndex.exit(exit.getIndex(), exit.getNetPrice(), exit.getAmount());
+                }
+            }
```
MarcDahlem, line 104: Instead of completely copying a trading record, wouldnt it be better to use something completely different here?

two ideas:
1. The tradingRecord stores all entry- and exit-indeces, e.g. in a sorted set. Then one can get the last entry index according to a given index with "sortedEntryIndeces.floor(index);" or last exit entry regarding a given index i with "sortedExitIndeces.floor(index)"
2. Or add a method criterion.calculate(series, originalTradingRecord, fromIndex, toIndex);

MarcDahlem:
(How) can this indicator also be used to solve https://github.com/ta4j/ta4j/issues/742 ?
In my opinion, this indicator is exactly what is needed to solve the position-based calculations.
But: How can one use criterion-based indicators for trading? Because we need the trading record BEFORE we initialize the indicator...

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CriterionIndicator.java
nimo23: The criterion needs a tradingRecord..your suggestion is not acceptable.

MarcDahlem: Of course it needs a trading record. But I would suggest not to copy a complete trading history for every calculation step.
We can use other approaches instead (like creating a new method  criterion.calculate(BarSeries series, TradingRecord record, int fromIndex, int toIndex); Then the copy tradingRecordTillIndex  is not needed at all

nimo23: Your suggestion involves changes in `AnalysisCriterion` and all its dedicated classes only to deserve the need for this Indicator.  I will not do that. But if you want you can create a PR after this was merged and suggest such improvements.

MarcDahlem: Yeah thats true.
I just wanted to indicate, that you can easily run into speed and memory problems with your implementation.
If you have a series with lets say 8000 bars and 100 positions for that series, you create a bit less than 8000*100 = 800.000 copies of positions for executing your strategy once (with just 1 such CriterionIndicator, for more indicators its getting worse)

nimo23: >that you can easily run into speed and memory problems with your implementation.

Thanks for your suggestions. Well, there are plenty more things in ta4j which can also run into speed and memory problems which thousands of bars.. If you want you can create a PR after this was merged and suggest such improvements.

MarcDahlem: Fine with me. I already got an idea ;-) Will be happy if you review it as soon as I have finished the work ;-)

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CriterionIndicator.java
```diff
@@ -0,0 +1,124 @@
...
+        this.requiredCriterionValue = requiredCriterionValue;
+        this.position = position;
+    }
+
+    @Override
+    protected Boolean calculate(int index) {
+        if (tradingRecord != null && !tradingRecord.getPositions().isEmpty()) {
+            TradingRecord tradingRecordTillIndex = new BaseTradingRecord(tradingRecord.getStartingType(),
+                    tradingRecord.getTransactionCostModel(), tradingRecord.getHoldingCostModel());
+            for (Position pos : tradingRecord.getPositions()) {
```
nimo23, line 94: Another alternative to the current implementation would be to have a method within `TradingRecord` to get a shallow copy (objects within the trades-list are only copied by reference):

```
    @Override
    public TradingRecord getTradingRecord(int startIndex, int endIndex) {
        if (startIndex < 0) {
            throw new IllegalArgumentException(String.format("the startIndex: %s must not be negative", startIndex));
        }
        if (startIndex >= endIndex) {
            throw new IllegalArgumentException(
                    String.format("the endIndex: %s must be greater than startIndex: %s", endIndex, startIndex));
        }
        TradingRecord shallowClone = new BaseTradingRecord(startingType, transactionCostModel, holdingCostModel);
        if (!trades.isEmpty()) {
            shallowClone.getTrades().addAll(trades.subList(startIndex, endIndex));
        }
        return shallowClone;
    }
```
@MarcDahlem @team172011 what do you think? Is it worth the effort?

nimo23:
Currently, the part of the tradingRecord till an index will be **deep copied** on every calculation:

To ways to improve this criterion:
1. add `criterion.calculate(BarSeries series, TradingRecord record, int fromIndex, int toIndex);` in `AnalysisCriterion` and all its dedicated classes only to deserve the need for this Indicator.
2. provide method within TradingRecord to **shallow copy** a tradingRecord (`TradingRecord#getSubTradingRecord(int startIndex, int endIndex)`)
3. Any other (better) solutions welcome.

I find solution 2 more attractive as it involves only one class change and avoid the need to implement `criterion.calculate(BarSeries series, TradingRecord record, int fromIndex, int toIndex)` for every Criterion now and in the future. 

I will NOT improve that in this PR. After this PR was merged, the next PR can improve the code.

MarcDahlem:
For me, idea 2 seems also for me to be the best alternative so far.

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CriterionIndicator.java
```diff
@@ -0,0 +1,124 @@
...
+            boolean entryTradeWithinIndex = position.getEntry().getIndex() <= index;
+            boolean exitTradeWithinIndex = position.getExit() == null
+                    || position.getExit() != null && position.getExit().getIndex() <= index;
+            if (entryTradeWithinIndex && exitTradeWithinIndex) {
+                Num calculatedCriterionValue = criterion.calculate(getBarSeries(), position);
+                return criterion.betterThan(calculatedCriterionValue, requiredCriterionValue)
+                        || calculatedCriterionValue.isEqual(requiredCriterionValue);
+            }
+        }
+        // empty tradingRecords or no position must return undefined
```
nimo23, line 120: The questions is: Should I return `null` instead of `false` ? The `null` (for Boolean) is similar to `NaN` for numbers. As long as no trades are recorded, neither `true` nor `false` can be determined, thus `null`. @team172011 what do you think?

PR #753, opened 2021-05-06
Improve BarSeries with numType

nimo23:

- Currently, all BarSeries (with its builder) needs `Function<Number, Num> function`. Instead of providing such function anywhere (where it is needed), we can replace it by providing an example of a `Num` (the property is called numType which is a common Num). Having the direct glue with Num instead of its Function (when constructing the BarSeries), provides the following benefits:
- we use `numType#function()` to get its function: no need to instantiate a new Function for every BarSeries-constructors or BarSeriesBuilders.
- we use static defined Num's (for example, `DecimalNum.ZERO`): no need to instantiate a new Num for every BarSeries-constructors or BarSeriesBuilders
- with `Function<Number, Num> function`, it's not possible to determine its (runtime) Num-Type. With numType, it is easy to determine its runtime type.
- using `numType` instead of `DecimalNum::valueOf`, we can provide both the function (by numType.function()) and also determine its type - vice versa, it's not possible.
- we can use `series.zero()`, `series.one()`, `series().hundred()` (etc.) which get its value by its static defined properties in the dedicated Num.class (this avoids the need to instantiate, for example,  `series.valueOf(1)`).
- we can use those static `series.zero()` on any other place (for example, indicators). Actually, we need at least one number to derive its constants (0,1,100, etc). Having `numType` instead of `numFunction` as a first class property in any barSeries solves this issue and we can access those static nums without the need of an instantiated num. So we can call easily `series.zero()` instead of `numberOfTen.zero()` (which is a little irritating).

Please review this PR carefully (in case I did forget something). 

By the way:
- if this PR is taken, then we can easily replace all those `valueOf(0)`, `valueOf(1)`, `valueOf(100)` by `series.zero()`, `series.one()`,  `series.hundred()` in the next step. Actually, we need at least one instantiated Num to access its constants. (A BarSeries can be empty, thus indicators cannot actually acccess the Num constants. Having numType solves this issue.).
- with `numType`, there is no more need to instantiate a `Num` and/or a `Num-function` on every construction (because both, `numType` and `function` can be easily accessed by constants defined in the Num-classes.)

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/BarSeries.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BarSeries.java
@@ -357,4 +357,30 @@ public interface BarSeries extends Serializable {
      */
     Function<Number, Num> function();
 
+    /**
+     * @return any num to determine its Num type
+     */
+    Num numType();
+
+    /**
+     * @return the Num of 0
+     */
+    default Num zero() {
+        return numType().zero();
+    }
+
+    /**
+     * @return the Num of 1
+     */
+    default Num one() {
+        return numType().one();
+    }
+
+    /**
+     * @return the Num of 100
+     */
+    default Num hundred() {
+        return numType().hundred();
+    }
+
 }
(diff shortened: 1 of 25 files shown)


nimo23:
I did not continue this PR (for example, in BaseBar: replace function with numType, adapt test cases, ..). 

So far, I only want to know if this PR is worth the effort before continuing.

team172011:
Interesting approach with some benefits. But we need to change a lot of code

nimo23:
Maybe it would be sufficient to change the `BarSeries`-constructors  (and do **not** change the `Bar`-constructors with all its testing classes..). The Bar-Constructors can have the Num Type from its BarSeries. We should look to have only overall **minimal** changes..

PR #757, opened 2021-05-07
Feature: trade-dependent indicators

### LINKED ISSUE #742: Position-based indicators

MarcDahlem:
Fixes #742.

- Allow trade based indicators, which can easily calculate their values by using the last executed trade
- Added a break even indicator based on the last executed entry position


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - :tada: **Enhancement** Added possibility to use CostModels when backtesting with the BacktestExecutor
 - :tada: **Enhancement** added Num#zero, Num#one, Num#hundred
 - **Example** added a json serialization and deserialization example of BarSeries using google-gson library
+- :tada: **Enhancement** Implemented an (abstract) TradeBasedIndicator, which can perform calculations based on the last trade
+- :tada: **Enhancement** Implemented a BreakEvenIndicator, which returns the needed break even regarding the last trade according to given transaction fees
```
team172011: Please change _Implemented ..._ to _Added ..._

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
```diff
@@ -47,7 +49,7 @@
     /**
      * The recorded trades
      */
-    private List<Trade> trades = new ArrayList<>();
+    private TreeMap<Integer, Trade> sortedTrades = new TreeMap<>();
```
nimo23: I find that this collection type seems more appropriate at the first glance. However, we should show (by testing through JMH) that this change does not decrease the overall performance (=memory and time) of ta4j (those `trades` are needed all over the place and especially the access for `ArrayList` is considerable faster (O(1)) than with `TreeSet`. Those `trades` are also analyzed by criteria which is used for backtests (and doing backtests should always be as fast as possible).

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
```diff
@@ -77,7 +79,7 @@
     /**
      * The recorded positions
      */
-    private List<Position> positions = new ArrayList<>();
+    private TreeMap<Integer, Position> sortedPositions = new TreeMap<>();
```
nimo23: Rename `sortedPositions` to `positions` (or at least to `allPositions`) (a TreeMap is always sorted, so the word `sortedPositions` is redundant). And change the javadoc to 

`/** The recorded positions (sorted in ascending order). */`

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
```diff
@@ -226,13 +228,31 @@ public boolean exit(int index, Num price, Num amount) {
 
     @Override
     public List<Position> getPositions() {
-        return positions;
+        return new LinkedList<>(sortedPositions.values());
```
nimo23: In general, `LinkedList` performance is not better than `ArrayList` (see https://stackoverflow.com/questions/322715/when-to-use-linkedlist-over-arraylist-in-java). We should show (by testing through JMH) that changing our core collection type does not decrease overall performance (memory and time) in a considerable manner.

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
```diff
@@ -226,13 +228,31 @@ public boolean exit(int index, Num price, Num amount) {
...
+    }
+
+    @Override
+    public Position getLastPositionRegardingIndex(int index) {
```
nimo23: rename to 

`public Position getLastPosition(int index){..}`

The "regarding index" is already implicitly told by its needing parameter list `(int index)`

ta4j-core/src/main/java/org/ta4j/core/TradingRecord.java
```diff
@@ -147,6 +148,11 @@ default Position getLastPosition() {
         return null;
     }
 
+    /**
+     * @return the last position recorded at the given time of 'index'
+     */
+    Position getLastPositionRegardingIndex(int index);
```
nimo23: rename to

`Position getLastPosition(int index){..}`

The "regarding index" is already implicitly told by its needing parameter list `(int index)`

ta4j-core/src/main/java/org/ta4j/core/TradingRecord.java
```diff
@@ -158,6 +164,11 @@ default Position getLastPosition() {
      */
     Trade getLastTrade(TradeType tradeType);
 
+    /**
+     * @return the last trade recorded at the given time of 'index'
+     */
+    Trade getLastTradeRegardingIndex(int index);
```
nimo23: rename to

`Position getTrade(int index){..}`

The "regarding index" is already implicitly told by its needing parameter list `(int index)`

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
MarcDahlem: As the trades and positions of a trading record are never accessed by position (there is only getLastTrade() and getLastPosition() on the class) , the slower access to elements should not be an issue here.

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
MarcDahlem: I can change it, no problem. (Although most criteria should iterate with for(Trade trade: trades) which is same speed for ArrayList and LinkedList). Personally, I would like to get rid of List's completely and would like to replace them with Collection's as the concrete type does not matter. Then we could just return sortedPositions.values() as it is

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
nimo23: Yes, right. Then there is no need to test for performance.

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
nimo23: Yes, change it to `return new ArrayList<>(sortedPositions.values());`. Don't return `Collection` because it does not explicitly express the kind of collection (which is important to _know_ for the calling methods, for example, `calculate(..)`); Returning a `List` is self-explanatory.

ta4j-core/src/main/java/org/ta4j/core/BaseTradingRecord.java
nimo23: Did not see that `getPositions(`) returns a list. Hence, it can remain

`TreeMap<Integer, Position> sortedPositions`

team172011 (changes requested):
I have reservation against this indicator type. Similar to #761 why should we mix up trades and indicators? Wouldn't a `Rule` be the place for such a decisions? In a Rule you have access to the `TradingRecord` of the current run. Having access to a `TradingRecord` in an `Indicator` means that you have to create another strategy with rules and indicators and so on just to apply this TradingRecord to this indicator. There may be special cases for this, but I am not sure if ta4j should support this by default and within the current architecture of this framework

CHANGELOG.md
MarcDahlem: > 
> 
> I have reservation against this indicator type. Similar to #761 why should we mix up `Trades` and `indicators`? Wouldn't a `Rule` be the place for such a decisions? In a `Rule` you have access to the `TradingRecord` of the current run. Having access to a `TradingRecord` in an `Indicator` means that you have to create another `Strategy` with `Rules` and `Indicators` and so on just to apply this `TradingRecord` to this indicator. There may be special cases for this, but I am not sure if ta4j should support this by default and within the current architecture of this framework

Mhm, not sure if that can be handled in Rules. If you have any ideas how to do it, I'm open to every suggestion.

I use the trade-based indicator quite much in my trading bot.
Examples:
* Break even indicator
    * --> if break even is reached, the exit criterion of the strategy is changed (needs only true and false)
    * --> place an order 2% above the break even (needs the concrete value/Num of the break even for every trade. Thats why it is an "calculator"(Indicator) and not a "condition"(Rule))
* Lowest/highest since last sell
* Restart of rules/conditions after a sell
    * --> example: Buy if EMA(20)>EMA(200). Sell on 2% gain. Without a "reset", the bot would directly buy again if the EMA(20) is still over EMA(200).
* BuyPrice-Indicator --> Indicator which holds the buy price during the lifetime of a trade
* IsBuyPhase indicator ..> true if in BuyPhase and false if not
* ...

If it is an indicator, other indicators and rules can easily depend on it.
For example TransformIndicator.muplitply(breakEvenIndicator, 1.04) directly contains the value which is 4% above the break even in EVERY Buy-Phase. new OverIndicatorRule(closePrice, breakEven) is true if the break even is reached...No need to handle multiple indicators, swich rules, swich strategies, adapt indicators... The only thing that must be handled is storing TradingRecords in my bot on every real executed trade on the market while it is running.

And those indicators can easily be used also in backtesting. They are really powerful!

nimo23:
>Wouldn't a Rule be the place for such a decisions? In a Rule you have access to the TradingRecord of the current run.

Yes, I also think you should create a new Rule instead of an Indicator. Because Indicator is linked with Bars and nothing else. I will look to create a new PR `CriterionRule` which is an improvement over https://github.com/ta4j/ta4j/pull/761. Maybe this CriterionRule is also sufficent for your use case. If not, please create a new Rule and call it `BreakEvenRule`.

nimo23:
After rethinking, I guess, such trade based decisions would not be solely possible with a Rule because we need the already executed TradingRecord (till the observed index) to look for. We need to add Indicators within Rules to determine if the Rule is satisfied. That's why we could make use of the `CriterionIndicator` (from https://github.com/ta4j/ta4j/pull/761) **within** a Rule (but a Rule cannot replace the CriterionIndicator).

team172011:
I think we have to distinguish between different use cases: **live trading** (e.g. implementing a bot) and **backtesting**. So far the only use case for `AnalysisCriterions` in ta4j has been backtesting. 

- After creating one or several `Strategies` you can execute them and get the results in form of `TradingReports`. After that you can analyze results by using `AnalysisCriterions`. It makes no sense here to have a backtesting run with Indicators that depend on an `AnalysisCriterion` and this criterion depends on a `TradingRecord` (that needs to be the result of another run?).

- For live trading once could imagine use cases where you need access to the `TradingRecord` and the (current) value of an `AnalysisCriterion` and where you want to implement logic that modifies the `TradingRecord` and so on. If you are using it in a trading bot for live trading this is also fine, but there is no general concept for using it together within an `Indicator` and/or `AnalysisCriterions` during a session. 

We cannot add _trade based indicators_ or _criterion based indicators_ to this lib just because you are using it during live trading in a specific situation. It will get confusing for the user if we mix things up this way.
I am open for improvement or extension for more live trading support in ta4j. But first we need a general and separated concept for live trading that can be discussed.

MarcDahlem:
For the moment, I dont need Crtierion.

I personally need Indicators, which can calculate their current value according the the current trade state.
Especially LowestSinceLastExit and BreakEven (only contains values in BuyPhases).
I need them especially during backtesting!

For live-trading I could workaround that by replacing my complete strategy with new indicators and rules as soon as a position changed.

But for Backtesting, I need calculators, which change if the last position is opened, closed or there is no last position. (And here I cannot change the complete Strategy and indicators, because Backtesting would not work). There are simple features like wait at least 200 ticks after an exit. Or get the current buy price in running open positions to calculate borders for gain and loss accordingly.

If you are not interested in such very useful indicators, I can close the PR and just include it in my project.

nimo23:
>We cannot add trade based indicators or criterion based indicators to this lib just because you are using it during live trading in a specific situation. It will get confusing for the user if we mix things up this way.

@team172011 Is this PR in any kind wrong? Does it return wrong values? Is this PR not usable? You say, the only reason is because it confuses users. Really? By the way, it's only one indicator (or criterion), nothing else. The CriterionIndicator or the TradeBasedIndicator is a bridge with minimal changes needed to provide such use cases. If you are able to provide a better solution, then please make a suggestion of how to analyze a tradingRecord the way like we need it. If both solutions are not included or discussed now, then it is very likely that such solutions will never be put into this lib.

PR #758, opened 2021-05-07
Feature: Simplified indicator cache

MarcDahlem:
Fixes #.

- Reduce complexity by replacing complicated cache index computations with simple TreeMap-Konstruct
- The new TA4JCache sorts automatically by indeces, and removes the least results regarding the index, if the capacity of the corresponding series is exceeded- 

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/CachedIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/CachedIndicator.java
@@ -23,12 +23,9 @@
  */
 package org.ta4j.core.indicators;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-
 import org.ta4j.core.BarSeries;
 import org.ta4j.core.Indicator;
+import org.ta4j.core.Ta4jCache;
 
 /**
  * Cached {@link Indicator indicator}.
@@ -39,15 +36,9 @@ import org.ta4j.core.Indicator;
 public abstract class CachedIndicator<T> extends AbstractIndicator<T> {
 
     /**
-     * List of cached results
+     * cached results
      */
-    private final List<T> results;
-
-    /**
-     * Should always be the index of the last result in the results list. I.E. the
-     * last calculated result.
-     */
-    protected int highestResultIndex = -1;
+    private final Ta4jCache<T> cache;
 
     /**
      * Constructor.
@@ -56,8 +47,7 @@ public abstract class CachedIndicator<T> extends AbstractIndicator<T> {
      */
     protected CachedIndicator(BarSeries series) {
         super(series);
-        int limit = series.getMaximumBarCount();
-        results = limit == Integer.MAX_VALUE ? new ArrayList<>() : new ArrayList<>(limit);
+        this.cache = new Ta4jCache<>(series);
     }
 
     /**
@@ -77,6 +67,9 @@ public abstract class CachedIndicator<T> extends AbstractIndicator<T> {
 
     @Override
     public T getValue(int index) {
+        if (cache.contains(index)) {
+            return cache.get(index);
+        }
         BarSeries series = getBarSeries();
         if (series == null) {
             // Series is null; the indicator doesn't need cache.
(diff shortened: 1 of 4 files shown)


nimo23:
@MarcDahlem I find your solution very nice. It reduces complexity very much. One point before going further: Can you compare both the actual `ArrayList`-Cache and this TA4JCache (with `TreeMap`) regarding Performance (**time and space** of create-read-update-delete entries) by running a simple benchmark through JMH?

MarcDahlem:
Mhm, never did that before. (here are very nice comparisons https://www.baeldung.com/java-microbenchmark-harness )

But TreeMap has O(log n) for all operations, whereas Arraylist.remove and Arraylist.add(index, entry) is O(n) (which were both used in the oild code, especially the shrink function should be very slow as it copies arrays, removes from the front and so on).

But Arraylist.get(index) has O(1), whereas TreeMap has also O(log n). Still, we dont need all those overhead calculations to calculate the index. As the old Caching-Behaviour should be very heavy in computation time, I think we dont loose so much if we performe a lot of get()s.

Lets see if I can benchmark it somehow

ta4j-core/src/main/java/org/ta4j/core/indicators/CachedIndicator.java
```diff
@@ -77,6 +67,9 @@ protected CachedIndicator(Indicator<?> indicator) {
 
     @Override
     public T getValue(int index) {
+        if (cache.contains(index)) {
+            return cache.get(index);
+        }
```
team172011, line 72: How does this work for bar series with removed bars (checked in line 96)?

For instance you have a `BarSeries` 
- with Bars at index 0, 1 and 2 (bar0, bar1, bar2)
- **maximumBarCount** = 3 

and you do 

- getValue(0) -> cache miss (calculate result0 and add 0 -> result0)
- getValue(1) -> cache miss (calculate result1 and add 1 -> result1)
- getValue(2) -> cache miss (calculate result2 and add 2 -> result2) // will not be cached because  2 == endIndex

after this the map in the cache looks like this:
0 -> result0
1 -> result1

now you add a further `Bar` to the `BarSeries`. That means the **first bar will be removed** (maximumBarCount = 3 and _bar1_ and _bar2_ will be shifted to index 0 and index 1 and the new _bar3_ will be added to index 2). 

- Then you call **getValue(0)** again -> cache hit, but with wrong result0

The wrong value will be returned and the corresponding logic for this case ignored

team172011 (commented):
Does this cache work correct with removed bars for BarSeries with maximumBarCount?

ta4j-core/src/main/java/org/ta4j/core/indicators/CachedIndicator.java
MarcDahlem: You are right, the cache only removes values if the cache contains more elements than the corrsponding series allows.

This does NOT mean, that every already calculated indicator value is removed from the cache if the corresponding bar is removed from the series.
But isnt that better and more accurate than the old cache, which returned always "calculate(0)" when the index was referencing a deleted bar?

The main reason to have a limited cache size, is to save memory for large running systems. And not to reduce accurracy, or am I wrong?
With the given new implementation it is assured, that the cache.size() is always smaller or equal to series.maxBarCount(). And for old values you may either get the correctly computed value if it is in the cache, or the first available result if it is not in the cache

team172011: > But isnt that better and more accurate than the old cache, which returned always "calculate(0)" when the index was referencing a deleted bar?

How can it be better to return a wrong value? The old caching logic returned the correct value because  "calculate(0)" is returning the value based of the first not removed Bar.

MarcDahlem: I think i get what you mean.
Expected: 
* indicator.get(series.getBeginIndex()) == X... 
* indicator.get(series.getBeginIndex()+1) == Y
... wait for first bar to be removed
* indicator.get(series.getBeginIndex()) == Y ---> but **at the moment you would still get X, as series.getBeginIndex() always returns 0**

I see two options:
1. I could fix this PR locally, by 'correcting' the index locally in the CachedBarIndicator
2. I would strongly suggest to correct the BarSeries. getBeginIndex() should return initialBeginIndex + removedBarsCount and getEndIndex should always return the overall added bars.

(Why I prefer 2.: Trades and all other classes store their index, which doesnt make sense when the index shifts afterwards.
For example trade.getIndex() is only valid at creation time of the trade, but if the series changes, the trade.getIndex() refers to a completely other bar than series.getBar(trade.getIndex()) )

team172011: > I would strongly suggest to correct the BarSeries. getBeginIndex() should return initialBeginIndex + removedBarsCount and getEndIndex should always return the overall added bars.

I think it is not just about the getBeginIndex, but about every time the index gets invalid for the cache because the bar series has reached its maximum bars limit and the first bars get deleted  for 'reusing' the inner `ArrayList`. Thats the reason why the current CachedIndicator is so complex. It needs to deal with all the different cases and may end up in a very expensive calculations for `BarSeries` with rolling indices.

> Trades and all other classes store their index, which doesnt make sense when the index shifts afterwards.

Good point. At the moment the _maximumBarCount_ feature only makes sense for live trading, where you are processing a lot of bars and just listening for signals and indicator values. It is not suitable for recording and applying criterions on it.

Maybe it would make sense to **not use the index** for caching but **the start or end date** of a bar. So you do not have to check if the cached values got invalid (only not caching the last bar surely). We only need to ensure that the cache will not become too large.

MarcDahlem: Thx for reviewing and pointing out some important current behaviours and wrong-calculations of my PR.
I have some ideas to optimize this PR by still simplifying the caches but also treating thereby all those current and important behaviours in a simplified way.

Until this is done, I mark this PR as draft.

MarcDahlem:
no time to work on this --> closed

PR #749, opened 2021-05-04
Add: NumberOfConsecutivePositionsCriterion

nimo23:
Changes proposed in this pull request:
- :tada: **Enhancement** added **`NumberOfConsecutiveLosingPositionsCriterion`**

- [x] added an entry with related ticket number(s) to the unreleased section of `CHANGES.md`

PR #759, opened 2021-05-08
Simplify criterions

nimo23:
- **NumberOfWinningPositionsCriterion** removed in favour of **NumberOfPositionsCriterion**
- **NumberOfLosingPositionsCriterion** removed in favour of **NumberOfPositionsCriterion**
- **WinningPositionsRatioCriterion** removed in favour of **PositionsRatioCriterion**
- **LosingPositionsRatioCriterion** removed in favour of **PositionsRatioCriterion**
- **NetProfitCriterion** removed in favour of **NetCriterion**
- **NetLossCriterion** removed in favour of **NetCriterion**
- **GrossProfitCriterion** removed in favour of **GrossCriterion**
- **GrossLossCriterion** removed in favour of **GrossCriterion**
- **AverageProfitCriterion** removed in favour of **AverageCriterion**
- **AverageLossCriterion** removed in favour of **AverageCriterion**


nimo23:
I am unopiniated about this PR. I don't fully know which solution is better in this case: the former version (with more classes, more redundant code, more cluttering, but easier creation) or the version of this PR. 

I added https://github.com/ta4j/ta4j/pull/749 making use of `PositionFilter` (which was a good solution for that case), but is it also a good solution for these classes listed in https://github.com/ta4j/ta4j/pull/759#issue-633819378?

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfPositionsCriterion.java
```diff
@@ -33,18 +35,60 @@
...
     }
 
     @Override
     public boolean betterThan(Num criterionValue1, Num criterionValue2) {
+        if (positionFilter == PositionFilter.PROFIT) {
+            return criterionValue1.isGreaterThan(criterionValue2);
+        }
+        if (positionFilter == PositionFilter.LOSS) {
+            return criterionValue1.isLessThan(criterionValue2);
+        }
```
team172011: `betterThan(c1, c2) `returns `true` if c1 is better than c2. So that means the comparison has nothing to do with `this.positionFilter` nor? We have to consider positionFilter of c1 and c2 and maybe throw an exception if positionFilters are not equal

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfPositionsCriterion.java
```diff
@@ -33,18 +35,60 @@
  */
 public class NumberOfPositionsCriterion extends AbstractAnalysisCriterion {
 
+    private final PositionFilter positionFilter;
```
team172011: This needs a Getter

ta4j-core/src/main/java/org/ta4j/core/AnalysisCriterion.java
```diff
@@ -39,6 +39,14 @@
  */
 public interface AnalysisCriterion {
 
+    /** Filter to differentiate between winning and losing positions. */
+    public enum PositionFilter {
```
team172011: This enum should be moved to its own file. It is used by some criterias

team172011: Can we find a better name? (Position)ReturnType?

team172011 (approved):

ta4j-core/src/main/java/org/ta4j/core/AnalysisCriterion.java
nimo23: - PositionType (not good, this it irritating with Buy and Sell)
- PositionSection
- PositionArea
- PositionItem
- PositionPart (good; means: either the profit part or the loss part)
- PositionReturnType (good, but just a little long name when using in so many criterions)

@team172011 You can choose.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfPositionsCriterion.java
nimo23: > So that means the comparison has nothing to do with this.positionFilter nor

@team172011 I don't understand what you mean. In general, the comparison depends on `this.positionFilter`. 

- Line 86-88 origin is https://github.com/ta4j/ta4j/blob/b6bad621a49260364a8450bd02539dd61373963a/ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfWinningPositionsCriterion.java#L49
- Line 89-91 origin is https://github.com/ta4j/ta4j/blob/b6bad621a49260364a8450bd02539dd61373963a/ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfLosingPositionsCriterion.java#L49
- Line 92 origin is https://github.com/ta4j/ta4j/blob/b6bad621a49260364a8450bd02539dd61373963a/ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfPositionsCriterion.java#L48

Or did you mean something else?

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfPositionsCriterion.java
nimo23: @team172011 I will do that after the best name for `PositionFilter` is selected. One general question about this PR (because I am a _little_ unsure): do you find this PR is an improvement in compare to the current version?

nimo23:
@team172011 and others: One general question about this PR (because I am a _little_ unsure): do you find this PR is an improvement in compare to the current version?

ta4j-core/src/main/java/org/ta4j/core/AnalysisCriterion.java
nimo23: I selected `PositionType`, it fits neatly to our existing naming scheme (for example, the enum `TradeType` is placed within `Trade.class`, so the `PositionType` is placed within `Position.class`).

`@param PositionType the PositionType to select either profit or loss positions`

The benefit is that the name is not only short, but it can be used also for other use cases not only limited to a `returnType`.

ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/NumberOfPositionsCriterion.java
team172011: ah yes, you're right this are `Num`

nimo23:
I added `Position#getType()` to determine its `PositionType`. With this, even more code simplifications can be made in dedicated Criteria (for example, a future criteria which uses `positionType == PositionType.PROFIT ? profit.isPositive() : profit.isNegative();` can be replaced by `positionType == position.getType()`. 

@team172011 I am done with this PR. Please review and suggest any improvements. Thanks.

team172011:
> I am unopiniated about this PR. I don't fully know which solution is better in this case: the former version (with more classes, more redundant code, more cluttering, but easier creation) or the version of this PR.
> 
> I added #749 making use of `PositionFilter` (which was a good solution for that case), but is it also a good solution for these classes listed in [#759 (comment)](https://github.com/ta4j/ta4j/pull/759#issue-633819378)?

I am also not sure. The current solution has a lot of redundant code, but your solution adds more complexity with an additional enum and in some cases this enum check needs to be done in each calculate-iteration (or we do more refactoring).

I think some Criteria are not really useful. For instance you do not need a `NumberOfLosingPositionsCriteria`, because it is the same as all positions minus number of winning positions. Same for the ratio (`WinningPositionsRatioCriterion` and `LosingPositionsRatioCriterion`). They only became relevant if you have zero profit trades and this could be configured with a parameter (boolean includeZeroProfit) or so.

nimo23:
>I think some Criteria are not really useful. For instance you do not need a NumberOfLosingPositionsCriteria, because it is the same as all positions minus number of winning positions

I don't think so. A `NumberOfLosingPositionsCriteria` (or NumberOfPositionsCriteria(PositionType.LOSS)) is useful if the user wants to add the criterion to a backtest. Otherwise it is not possible. So those criterions have their right to exist.

>in some cases this enum check needs to be done in each calculate-iteration (or we do more refactoring).

Good point! How can we improve that? I think "performance" is important in this case because criteria are used for backtesting purposes..

Without `enum PositionType`, we have a lot of redundant code (which must be maintained) and more classes (which should belong together). But I also don't want to add unnessecary complexity. 

@team172011 I also don't know which solution is better..if you want, you can think about this PR for a while and after decision is made give feedback or close this issue. Maybe other users can also provide feedback about this PR.

nimo23:
I will do a full rework of this PR in the next week. 

Now, it's clear that classes which are heavily related and use redundant (or very similar) logic should be merged together - the "disciminators" should be "parameters" (within a constructor or method body) and not "classes".

PR #761, opened 2021-05-08
New: CriterionIndicator

nimo23:
- `CriterionIndicator.java`

Replaces https://github.com/ta4j/ta4j/pull/747.

Currently, in `CriterionIndicator#calculate(int index)` the part of the tradingRecord till an index will be **deep copied** on every calculation:

Ways to improve this criterion:
1. add `criterion.calculate(BarSeries series, TradingRecord record, int fromIndex, int toIndex);` in `AnalysisCriterion` and all its dedicated classes only to deserve the need for this Indicator.
2. provide method within TradingRecord to **shallow copy** a tradingRecord (`TradingRecord#getSubTradingRecord(int startIndex, int endIndex)`)
3. Any other (better) solutions welcome.

I find solution 2 more attractive as it involves only one class change and avoid the need to implement `criterion.calculate(BarSeries series, TradingRecord record, int fromIndex, int toIndex)` for every Criterion now and in the future. 

I will NOT improve that in this PR. After this PR was merged, the next PR can improve the code.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CriterionIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CriterionIndicator.java
@@ -46,7 +46,6 @@ public final class CriterionIndicator extends CachedIndicator<Boolean> {
     private final AnalysisCriterion criterion;
     private final Num requiredCriterionValue;
     private final TradingRecord tradingRecord;
-    private final Position position;
 
     /**
      * Constructor.
@@ -64,7 +63,6 @@ public final class CriterionIndicator extends CachedIndicator<Boolean> {
         this.tradingRecord = tradingRecord;
         this.criterion = criterion;
         this.requiredCriterionValue = requiredCriterionValue;
-        this.position = null;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CriterionIndicator.java
```diff
@@ -0,0 +1,124 @@
...
+            boolean exitTradeWithinIndex = position.getExit() == null
+                    || position.getExit() != null && position.getExit().getIndex() <= index;
+            if (entryTradeWithinIndex && exitTradeWithinIndex) {
+                Num calculatedCriterionValue = criterion.calculate(getBarSeries(), position);
+                return criterion.betterThan(calculatedCriterionValue, requiredCriterionValue)
+                        || calculatedCriterionValue.isEqual(requiredCriterionValue);
+            }
+        }
+        // empty tradingRecords or no position must return undefined
+        return false;
```
nimo23: The questions is: 
Should I return `null` instead of `false` ? The null (for Boolean) is similar to NaN for numbers. As long as **no** trades are recorded, neither true nor false can be determined, thus null. 

@team172011 what do you think?

team172011: I would return false, I do not like the possibility of trivalent expressions for boolean values

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/CriterionIndicator.java
```diff
@@ -0,0 +1,124 @@
...
+                    Trade exit = pos.getExit();
+                    tradingRecordTillIndex.exit(exit.getIndex(), exit.getNetPrice(), exit.getAmount());
+                }
+            }
+
+            Num calculatedCriterionValue = criterion.calculate(getBarSeries(), tradingRecordTillIndex);
+            return criterion.betterThan(calculatedCriterionValue, requiredCriterionValue)
+                    || calculatedCriterionValue.isEqual(requiredCriterionValue);
+
+        } else if (position != null) {
```
team172011: I would not support to provide a `Position` or a `TradingRecord` in the construction. This check needs to be done in every calculation, but is always the same then

nimo23: >I would not support to provide a Position or a TradingRecord in the construction...but is always the same then

But the `calculate()` can start with an empty tradingRecord..which is filled while running the indicator. That's the reason for the check on every calculate()-method. Or do you mean something else?

team172011: where does the TradingRecord gets filled? If you use the second constructor the TradinRecord is always `null`. I am also unsure if there is a real use case for this kind of indicators

nimo23: The point of this `CriterionIndicator` is: 
To have the ability to measure the `TradingRecord` while trading, the same as we measure the `BarSeries` with Indicators while trading. It's similar to https://github.com/ta4j/ta4j/pull/757, but more general - because you can put any kind of criterion within this `CriterionIndicator`. For example, you measure the PnL of the TradingRecord with this indicator (by `CriterionIndicator(new ProfitLossCriterion()`) to get a signal when to exit. Currently, we have no ability to use all the criterions as if there where indicators. 

While the typical Indicators only use `BarSeries` for its decision, with the CriterionIndicator it can use any `TradingRecord` for its decision.

I simplified the CriterionIndicator and removed the Position from the Constructor (it does not make sense to measure a Position, it would be the same as I would have a BarSeries with only one Bar).

team172011:
So far this kind of indicator has not been discussed anywhere and so far I do not think that it makes sense to add this to the library. 
BarSeries -> Indicator -> Rule -> Strategy -> TradingRecord -> Criterion
A `TradingRecord` is the result of running a `Strategy` and a `Criterion` can be applied on a `TradingRecord`. I think everything you could archive with this Indicator you could also archive by creating or using corresponding `Rules`.


> I will NOT improve that in this PR. After this PR was merged, the next PR can improve the code.

I do not agree with this procedure. A PR is for reviewing and discussion changes. Once should only merge if the feature or fix is ready and approved.

nimo23:
>I do not agree with this procedure. A PR is for reviewing and discussion changes. Once should only merge if the feature or fix is ready and approved.

**It is ready and approved!** I works. But there is room for improvements, the same as the whole lib has room for improvements. 


>So far this kind of indicator has not been discussed anywhere and so far I do not think that it makes sense to add this to the library.

No problem. I solve this problem and leave this limitation in this lib by closing this issue and using it in my fork.

PR #765, opened 2021-05-09
fix: "Returns.java" calculate endIndex within loop

nimo23:
- inlcude endIndex within loop of calculate()

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/analysis/Returns.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/analysis/Returns.java
@@ -173,7 +173,7 @@ public class Returns implements Indicator<Num> {
         // returns are per period (iterative). Base price needs to be updated
         // accordingly
         Num lastPrice = position.getEntry().getNetPrice();
-        for (int i = startingIndex; i < endIndex; i++) {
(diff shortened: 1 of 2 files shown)


ta4j-core/src/main/java/org/ta4j/core/analysis/Returns.java
```diff
@@ -173,7 +173,7 @@ public void calculate(Position position, int finalIndex) {
         // returns are per period (iterative). Base price needs to be updated
         // accordingly
         Num lastPrice = position.getEntry().getNetPrice();
-        for (int i = startingIndex; i < endIndex; i++) {
+        for (int i = startingIndex; i <= endIndex; i++) {
```
nimo23, line 176: @team172011 I am not sure, if my fix is valid. But the `endIndex` is calculated by 

`int endIndex = CashFlow.determineEndIndex(position, finalIndex, barSeries.getEndIndex());`


and `CashFlow.determineEndIndex()` has the following javadoc:

`Determines the the valid final index to be considered.`

This means to me, that the endIndex should be considered, which is currently not the case when using `<endIndex ` instead of `<=endIndex`. Am I right?

PR #770, opened 2021-05-12
added NumUtils

nimo23:

-:tada: **NumUtils.java** to provide common "financial/statistical/math"-formulas with `Num`.

In future, we can move common calculations (needed for indicators or criteria, calculations currently duplicated on multiple places)  to `NumUtils.java`. 

If you encounter duplicate formulas used within any indicator/cirterion/etc., then it may be worth to consider to include a general formula within NumUtils (DRY = "dont repeat yourself").

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/utils/NumUtils.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/utils/NumUtils.java
@@ -28,7 +28,7 @@ import org.ta4j.core.num.Num;
 /**
  * Common formulas for Math and Finance.
  */
-public final class MathUtils {
+public final class NumUtils {
 
     /**
      * Calculates the compound interest by the following formula:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 
 ### Breaking
 - **DifferencePercentage** renamed to **`DifferencePercentageIndicator`**
+- **DXIndicator** moved to adx-package
+- **PlusDMIndicator** moved to adx-package
+- **MinusDMIndicator** moved to adx-package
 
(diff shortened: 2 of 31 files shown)


ta4j-core/src/main/java/org/ta4j/core/utils/NumUtils.java
```diff
@@ -0,0 +1,51 @@
...
+     * <pre>
+     * (1 + percentGrowth/100)pow(days) * initialCapital
+     * </pre>
+     * 
+     * @param initialCapital the initial capital
+     * @param percentGrowth  the percent growth in percentage format (e.g. 5%)
+     * @param time           the time (e.g. 1 year)
+     * @return the compound interest
+     */
+    public static Num compoundInterest(Num initialCapital, Num percentGrowth, int time) {
```
team172011, line 45: This is nothing that I would expect to be in a NumUtils class

nimo23: The `NumUtils` class should provide common "**financial/statistical/math**"-formulas..this also includes typcial formulas which can be used in indicators and criterions (for example, typical math formulas, like formulas for average, variance, regression formulas, etc. ). And the `NumUtils` should avoid the need to include something like apache-math or the like. The `compoundInterest` is an example of a  typcial financial formula (https://www.investopedia.com/terms/c/compoundinterest.asp). The NumUtils should also avoid duplicated formulas in indicators, criteria, etc. 

>This is nothing that I would expect to be in a NumUtils class

What would you expect to have in the NumUtils?

team172011:
as far as not needed in indicators or criterions providing common "financial/statistical/math"-formulas is out of scope for ta4j. @nimo23 if you not agree please open a ticket for discussion

nimo23:
Ok, no problem. I also think that this it out of scope.

PR #772, opened 2021-05-14
Added a delay indicator

MarcDahlem:
Fixes #.

- Added an indicator which can delay its computation.
- This DelayIndicator can be used by different indicator implementations, which rely on delayed information, such as the Ishimoku-Cloud computation
- 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,6 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 ### Added
 - :tada: **Enhancement** added possibility to use CostModels when backtesting with the BacktestExecutor
 - :tada: **Enhancement** added Num#zero, Num#one, Num#hundred
-- :tada: **Enhancement** added possibility to use CostModels when backtesting with the BacktestExecutor
 - :tada: **Enhancement** added Indicator#stream() method
 - :tada: **Enhancement** added a new CombineIndicator, which can combine the values of two Num Indicators with a given combine-function
 - **Example** added a json serialization and deserialization example of BarSeries using google-gson library
@@ -35,6 +34,8 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - :tada: **Enhancement** added **`StandardErrorCriterion`**
 - :tada: **Enhancement** added **`VarianceCriterion`**
 - :tada: **Enhancement** added **`AverageCriterion`**
+- :tada: **Enhancement** added **`DelayIndicator`**
+
 
 
 ## 0.14 (released April 25, 2021)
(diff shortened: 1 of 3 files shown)


team172011 (changes requested):
I think this is the same as `PreviousValueIndicator`

MarcDahlem:
Oh well, I didnt find this Indicator.
Still there are some differences:
* They DelayIndicator returns NaN if the delayed index is out of bounce. The PreviousValueIndicator returns the value of the first bar
* The DelayIndicator works with negative delays (printing to future). The PreviousValueIndicator doesnt and throws an exception
* The DelayIndicator uses getBeginIndex() the PreviousValueIndicator uses '0' to get the first bars index.

All in all I would like to merge both indicators together into one indicator which has the behaviour of the DelayIndicator.
Would that be okay for you?
How should the new Indicator be named? (PreviousValueIndicator doesnt show that future indeces are allowed. But its already there...)

team172011:
> The DelayIndicator works with negative delays (printing to future). The PreviousValueIndicator doesnt and throws an exception

How and why should this be possible? You cannot look into the feature and there is and should not be any strategy or indicator that is based on values from the future

> The DelayIndicator uses getBeginIndex() the PreviousValueIndicator uses '0' to get the first bars index.

this can be fixed in `PreviousValueIndicator` by using beginIndex instead of 0

> They DelayIndicator returns NaN if the delayed index is out of bounce. The PreviousValueIndicator returns the value of the first bar

This is just a matter of taste. The current implementation follows the approach to avoid returning NaN because it can cause problems for recursive indicators or rather indicator that depend on previous values

MarcDahlem:
> 
> 
> > The DelayIndicator works with negative delays (printing to future). The PreviousValueIndicator doesnt and throws an exception
> 
> How and why should this be possible? You cannot look into the feature and there is and should not be any strategy or indicator that is based on values from the future

Just do new PreviousValueIndicator(new PreviousValueIndicator(indicator, 5), -3). This doesnt look really into the future, but it gives future values according to the first previous value indicator. It's not really needed, though.  But it can enhance readability e.g. the "Future Cloud" of Ishimoku Indicator and similar indicators.

> 
> > The DelayIndicator uses getBeginIndex() the PreviousValueIndicator uses '0' to get the first bars index.
> 
I will fix this.

> this can be fixed in `PreviousValueIndicator` by using beginIndex instead of 0
> 
> > They DelayIndicator returns NaN if the delayed index is out of bounce. The PreviousValueIndicator returns the value of the first bar
> 
> This is just a matter of taste. The current implementation follows the approach to avoid returning NaN because it can cause problems for recursive indicators or rather indicator that depend on previous values

But e.g. the EMA(n) and SMA(n) values are wrong for the first n time steps. You can write new UnstableIndicator(ema, n). But if it directly returns NaN this is not needed anymore.

MarcDahlem:
no time to work on this --> closed

PR #773, opened 2021-05-14
Feature/before rule

MarcDahlem:
Fixes #.

- Added a before rule, which allows to easily create conditions that must not be valid at the same time.
- E.g. a cross up is needed before some other conditions hold like in the Ichimoku indicator
- 


team172011:
IMO this rule is too specific to be part of the ta4j `Rule` base. @MarcDahlem if you disagree, please open a ticket for discussion

MarcDahlem:
No its fine. I use all the PRs in my project or fork and dont need them in ta4j project

PR #779, opened 2021-05-18
Keltner construction/uncaching

johnthetireless:
Uncached Keltner middle; it completely delegates calculation to its EMA

Uncached upper and lower; the calculations are simple and do not need to be cached, IMO; I can undo this if you want.  If you agree with this uncaching for Keltner upper/lower, there are quite a few others I can uncache the same way

Added constructors so upper/lower can share an ATR instance; possible small performance gain

Added getBarCount() and toString() for consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,8 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 
 ### Changed
 - **KeltnerChannelMiddleIndicator** changed superclass to AbstractIndicator; add GetBarCount() and toString()
-- **KeltnerChannelUpperIndicator** changed superclass to AbstractIndicator; add constructor to accept pre-constructed ATR; add GetBarCount() and toString()
-- **KeltnerChannelLowerIndicator** changed superclass to AbstractIndicator; add constructor to accept pre-constructed ATR; add GetBarCount() and toString()
+- **KeltnerChannelUpperIndicator** add constructor to accept pre-constructed ATR; add GetBarCount() and toString()
+- **KeltnerChannelLowerIndicator** add constructor to accept pre-constructed ATR; add GetBarCount() and toString()
 - **BarSeriesManager** removed empty args constructor
 - **Open|High|Low|Close** do not cache price values anymore
 - **DifferenceIndicator(i1,i2)** replaced by the more flexible CombineIndicator.minus(i1,i2)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - :tada: **Fixed** **`ChaikinOscillatorIndicatorTest`**
 
 ### Changed
+- **KeltnerChannelMiddleIndicator** changed superclass to AbstractIndicator; add GetBarCount() and toString()
+- **KeltnerChannelUpperIndicator** changed superclass to AbstractIndicator; add constructor to accept pre-constructed ATR; add GetBarCount() and toString()
+- **KeltnerChannelLowerIndicator** changed superclass to AbstractIndicator; add constructor to accept pre-constructed ATR; add GetBarCount() and toString()
 - **BarSeriesManager** removed empty args constructor
 - **Open|High|Low|Close** do not cache price values anymore
 - **DifferenceIndicator(i1,i2)** replaced by the more flexible CombineIndicator.minus(i1,i2)
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/keltner/KeltnerChannelLowerIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/keltner/KeltnerChannelLowerIndicator.java
@@ -24,7 +24,7 @@
 package org.ta4j.core.indicators.keltner;
 
 import org.ta4j.core.indicators.ATRIndicator;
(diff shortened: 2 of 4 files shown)


team172011:
IMO not caching middle Keltner Indicator makes sense. But I am unsure about the lower and upper middle indicator. You think not having the caching logic with array list outweighs the redundant num operations there?

johnthetireless:
>>You think not having the caching logic with array list outweighs the redundant num operations there?

I mostly look at this as a space vs time issue,  If an indicator like Keltner upper is cached, we consume more RAM and we add a little bit of overhead to the most common case, where getValue() is called for the first (and maybe only) time as we iterate over the bar series (backtesting) or add new bars (live trading).  If we don't cache, we save the RAM and the first pass getValue() is slightly faster.   

Caching can only have a performance benefit if we revisit indexes in the getValue(0 call.  I don't believe the calculation EMA(i) + ATR(i) * 2 will take more than a few microseconds.  (I have some experience with high volume trading systems for banks; I have seen much more complex BigDecimal math than this with no thought of caching.)

If people are worried about running out of RAM, they set a maxBarCount (I don't use the lib this way, but it seems obvious).  This will keep the bar series array list at a reasonable length.    Keeping the number of "parallel" array lists used for indicator caching is an equally good way to guard against out of memory conditions.  

BTW,  the CachedIndicator constructor is a bit scary.  It will try to create really big array lists if the user sets barSeries.maxBarCount to a big number.  This is a stability problem.  I doubt security is a priority for this lib at this point in time, but this would almost certainly raise alarms in a security review.

I will restore the caching if that's what you want.  Just let me know one way or the other.

MarcDahlem:
I was running into memory problems very fast while executing benchmarks on ~300 different series with a lot of params.
The trick to avoid this was 2 things

1. Use Queues instead of Lists, Sets or Maps to hold a strategy and all corresponding indicators and rules. THis allows to iterate over the created Strategies with  for(..) but still allow the garbage collector to remove already benchmarked configurations and indicators.
    * https://github.com/MarcDahlem/ta4j/blob/test/test-my-strategies/ta4j-examples/src/test/java/ta4jexamples/strategies/IntelligentTa4jOhlcBenchmarks.java#L417
2.  Destroy a strategy as soon as the benchmark results are available. 
    * https://github.com/MarcDahlem/ta4j/blob/test/test-my-strategies/ta4j-examples/src/test/java/ta4jexamples/strategies/IntelligentTa4jOhlcBenchmarks.java#L607
    * https://github.com/MarcDahlem/ta4j/blob/test/test-my-strategies/ta4j-core/src/main/java/org/ta4j/core/BaseStrategy.java#L157

Maybe this helps with your problem

johnthetireless:
> I was running into memory problems very fast while executing benchmarks on ~300 different series with a lot of params.

I wonder if your benchmarks would notice any difference before and after uncaching indicators as I suggest.  I will buy you a coffee if you can prove the uncached versions are more than - 5%, perhaps? - slower than the cached versions.  :)

MarcDahlem:
Without your commit:

StartTime Keltner Benchmark 00:02:23.399
EndTime Keltner Benchmark: 00:10:27.183
Time ellapsed: 8 minutes and 3.784 s

With your commit:
StartTime Keltner Benchmark 00:20:02.184
EndTime Keltner Benchmark 00:27:45.912
Time ellapsed: 7 minutes 43.728s

Both setups used all available 12GB ram, but did not crash. I have to perform a larger benchmark or set my jvm limits to force the out of memory crashes, what I didnt want to do for now ;-)

johnthetireless:
> Without your commit:
> > Time ellapsed: 8 minutes and 3.784 s
> 
> With your commit:
> > Time ellapsed: 7 minutes 43.728s
> 
Thank you, Marc.  The elapsed times look very much the same.  The small difference might be due to the cache check.  The uncached versions should be a fraction faster in the usual backtesting case.

I assume you simply run over the series once, from 0 to barCount?  This doesn't really test the caching; not your concern, of course.  I should probably create a test that gets values from the same index several times, using cached and uncached indicators.  I am guessing the uncached versions will be slower on the second and subsequent attempts to getValue(i), but only slightly.

johnthetireless:
Quick test using a small bar series....  simply calls getValue for the 3 Keltner indicators...
Bar count: 252
Iterate over series N times, creating new cached indicators each time
Total for 1000 runs: 1934 millis
Iterate over series N times; create cached indicators once
Total for 1000 runs: 26 millis
Iterate over series N times, creating new uncached indicators each time
Total for 1000 runs: 1259 millis
Iterate over series N times; create uncached indicators once
Total for 1000 runs: 206 millis
Iterate over series N times; create uncached indicators once; shared ATR
Total for 1000 runs: 194 millis

Uncached versions are a little faster for one time 0..barCount iteration.
Uncached versions are a bit slower if we iterate several times; the calculations take slightly longer than the cache retrieval;
- note that the 26 vs 206 millis difference is for 1000 iterations; the actual difference per access is measured in microseconds
- the tests that create new indicators for each iteration probably measure object creation time more than anything else
- sharing the ATR doesn't seem to improve things; the underlying MMA value is cached, so I am a little surprised by this

I did not test cases with maxBarCount but I think we can be sure cached indicators will start to perform worse when the limit is reached.

johnthetireless:
@team172011 

In the interest of moving forward, I have  reverted the caching for Keltner upper/lower.   I think we have agreed on all the small changes effected by this PR.  When this one is merged I will move on to other indicators, keeping the same narrow focus.

team172011 (approved):

PR #783, opened 2021-05-21
Feature/useable ichimoku indicators

### LINKED ISSUE #705: Ichimoku Chikou Span Indicator Calculation Incorrect?
Hi.

Just evaluating Ichimoku Cloud and i dont know, but might be calculation from Chikou Span is not correct.
If i check your underlying link https://school.stockcharts.com/doku.php?id=technical_indicators:ichimoku_cloud i see the calculation must be 26 Days in the past. So, 26 bars back. In IchimokuChikouSpanIndicator Class i see, calculation is done by "index + timeDelay". So this mean index (now) plus 26 in Front. Is this right?

core/indicators/ichimoku/IchimokuChikouSpanIndicator.java
Line: 74

`

  int spanIndex = index + timeDelay;
  if (spanIndex <= getBarSeries().getEndIndex()) {
    return closePriceIndicator.getValue(spanIndex);
  } else {
    return NaN.NaN;
   }
}
`

MarcDahlem:
Fixes #705.

- Cloud-Span indicators calculate their value according to the current index (only printing is in future)
- Lagging-Span is just a synonym for the current close price (only printing in graphs is in the past.)
- Added an example strategy which shows how to use those ichimoku indicators and how to get the current cloud, past cloud and how to use the lagging span correctly

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/ichimoku/IchimokuChikouSpanIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/ichimoku/IchimokuChikouSpanIndicator.java
@@ -44,39 +44,25 @@ public class IchimokuChikouSpanIndicator extends CachedIndicator<Num> {
     private final ClosePriceIndicator closePriceIndicator;
 
     /**
-     * The time delay
-     */
-    private final int timeDelay;
-
-    /**
-     * Constructor.
+     * Constructor. The ichimoku chikou span returns for an index i always the
+     * current close price for i. Only its usage and comparison against other prices
+     * takes into account the past. E.g. new OverIndicatorRule(chikouSpanIndicator,
+     * new PreviousValueIndicator(closePriceIndicator, chikouSpanDelay)) This rule
+     * is satisfied, if the current value of the span is over the old close price.
+     *
+     * The chikou span calculation is always based on the current values, but only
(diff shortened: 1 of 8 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/ichimoku/IchimokuChikouSpanIndicator.java
```diff
@@ -44,39 +44,25 @@
...
     @Override
-    protected Num calculate(int index) {
-        int spanIndex = index + timeDelay;
-        if (spanIndex <= getBarSeries().getEndIndex()) {
-            return closePriceIndicator.getValue(spanIndex);
-        } else {
-            return NaN.NaN;
-        }
+    protected Num calculate(int i) {
+        return closePriceIndicator.getValue(i);
```
team172011, line 65: I do not agree with this change. The Ichimoku chikoun span is not equal to close price value. https://www.investopedia.com/terms/i/ichimoku-cloud.asp

team172011 (changes requested):
I do not agree with your interpretation and calculations of the different lines of the ichimoku cloud. compared to the existing implementation that seems to be based on https://www.investopedia.com/terms/i/ichimoku-cloud.asp the changes are not correct
or at least misleading for the different indicators

PR #788, opened 2021-06-16
added javadoc for all rules

nimo23:
Currently it's not clear which rule makes use of the `tradingRecord` and which not (without looking into the source).

- added javadoc within all rules to make clear which rule uses the tradingRecord and which not


team172011 (changes requested):
this javadoc additions are not compliant with existing javadoc. Please use the existing format 


```
    /**
     * 
     * @param index
     * @param tradingRecord
     * @return
     */
```

nimo23:
>this javadoc additions are not compliant with existing javadoc.

the javadoc additions should inherit 

```
/**
     * 
     * @param index
     * @param tradingRecord
     * @return
     */
```

**from its parent (interface)**. I  only added a comment (without param) so that it's only  added to the existing javadoc. If I use the full style as you suggested, then the javadoc of the interface is overwritten which I wanted to avoid.

team172011 (approved):

PR #796, opened 2021-08-05
New class org.ta4j.core.utils.Analysis

Fenrur:
- Adding new class org.ta4j.core.utils.Analysis: for reduce code, it contains only static elements
- Modify all indicators with org.ta4j.core.utils.Analysis

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/ATRIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/ATRIndicator.java
@@ -27,6 +27,9 @@ import org.ta4j.core.BarSeries;
 import org.ta4j.core.indicators.helpers.TRIndicator;
 import org.ta4j.core.num.Num;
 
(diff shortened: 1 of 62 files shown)


nimo23:
Hi,

the `Analysis.class`:
- contains only redundant code from other classes
- behaves like a god class
- uses not always self descriptive names (for example, tr, nz, etc.)
- has no javadoc at all (and if you put a javadoc, then it would be only redundant to its calling method which shows more that it only contains duplicates code)
- all indicators have a strong dependency to this `Analysis.class`, so the big `Analysis.class` must be always loaded even if an indicator only uses one static method
- adds unnecessary complexity
- adds indirect code which has not any benefit
- acts only as a nonsense wrapper (instead of simply calling the contructor directly, you always call a static method which then calls the constructor and nothing more) -> this is nonsense

I really don't like to see this kind of `Analysis.class` in this lib.

team172011:
@Fenrur I agree with all points mentioned by @nimo23 I do not think this is an improvement

PR #797, opened 2021-08-07
Prevent superfluous `Object[]` allocation for varargs `log.trace` and `log.debug` calls

Petersoj:
- Wraps all vararg (infinite argument) `log.trace` and `log.debug` SLF4j Logger calls in `if`-statement to prevent superfluous `Object[]` creation which decreases CPU performance and increases memory usage. 

The biggest performance hit in my case comes from the `CachedIndicator.getValue()` call which is called thousands of times per second when running my trading algo backtest. I used the Java VisualVM profiler to profile memory when running my backtests and it revealed the varargs `log.trace()` call is hurting memory/CPU performance quite a bit.
This first screenshot shows a memory profile of the top allocators of `java.lang.Object[]`. When the vararg call of any Logger method is used, a `java.lang.Object[]` is created before invoking the log method, as the SLF4j documentation states [here](http://www.slf4j.org/apidocs/org/slf4j/Logger.html#trace(java.lang.String,java.lang.Object...)):
> However, this variant incurs the hidden (and relatively small) cost of creating an Object[] before invoking the method, even if this logger is disabled for TRACE. The variants taking one and two arguments exist solely in order to avoid this hidden cost.

old_profiler

As you can see, 94.8% of `Object[]` live bytes are allocated by `CachedIndicator.getValue()`.
Here is the profiler screenshot with my changes in this PR:

new_profiler

`CachedIndicator.getValue()` doesn't even show up and the number of `Object[]` live bytes is cut by around 60%. I should note that these memory profiling reports were done under the exact same server environments both times. 
Additionally, I took a holistic memory sample during my backtests using the VisualVM profiler. The number of `Object[]` live objects was cut by around 95%. Note that these memory samples were taken a few moments after a garbage collection was invoked. 

`0.15-SNAPSHOT` develop branch:
old_sampler

With my changes in this PR:
new_sampler

🙂

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - :tada: **Enhancement** added **`VarianceCriterion`**
 - :tada: **Enhancement** added **`AverageCriterion`**
 - :tada: **Enhancement** added javadoc for all rules to make clear which rule makes use of a TradingRecord
+- **Enhancement** prevent Object[] allocation for varargs log.trace and log.debug calls by wrapping them in `if` blocks
 
 ## 0.14 (released April 25, 2021)
 
--- a/ta4j-core/src/main/java/org/ta4j/core/BarSeriesManager.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BarSeriesManager.java
@@ -160,8 +160,10 @@ public class BarSeriesManager {
         int runBeginIndex = Math.max(startIndex, barSeries.getBeginIndex());
         int runEndIndex = Math.min(finishIndex, barSeries.getEndIndex());
 
-        log.trace("Running strategy (indexes: {} -> {}): {} (starting with {})", runBeginIndex, runEndIndex, strategy,
-                tradeType);
+        if (log.isTraceEnabled()) {
+            log.trace("Running strategy (indexes: {} -> {}): {} (starting with {})", runBeginIndex, runEndIndex,
+                    strategy, tradeType);
+        }
         TradingRecord tradingRecord = new BaseTradingRecord(tradeType, transactionCostModel, holdingCostModel);
         for (int i = runBeginIndex; i <= runEndIndex; i++) {
             // For each bar between both indexes...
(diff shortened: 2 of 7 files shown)


nimo23:
Another alternative would be to use **lambda expression**.

For example:

Instead of 

```
if (log.isTraceEnabled()) {
     log.trace("{}({}): {}", this, index, result);
 }
```

lambda expression can be used:

`log.trace("{}({}): {}", ()->this, ()->index, ()->result);`

See more:
- https://jira.qos.ch/browse/SLF4J-371
- https://ionutbalosin.com/2019/11/logging-patterns-and-their-performance-impact/

Petersoj:
@nimo23 using a lambda expression in your example is redundant. If there was some preprocessing to be done on arguments passed into the `log.trace(String, Object)` call, then it would help, but most of the log calls in `ta4j-core` are simply passing references. Creating an in-line lambda expression is actually slightly more processor/memory intensive since an anonymous `Function` object is created in the heap before calling the `log.trace()` method in your example. Even if you were using lamdas, it still creates an `Object[]` when using the varargs calls which is where the performance impact is greatest. So the `if`-block is necessary, and cleaner IMO.

team172011:
@Petersoj nice founding! I also think this makes sense because object allocation is a bottle neck of ta4j at the moment. 
It would also be nice to know the performance impact of this change - we save the `Object[]...` allocation if the logging level has not been reached vs. we have the double check for the logging level if it has been reached. According to http://www.slf4j.org/faq.html#logging_performance evaluating a longer takes around 1% of the whole logging procedure. However I like this solution because also in this case the performance impact only occurs if the logging level has been reached and saves memory when the logging level has not been reached.

team172011 (approved):

Petersoj:
@team172011 processor performance has most definitely increased. I don't have the profiling screenshots, but when I was CPU profiling my backtest, I could see that nearly a quarter of the CPU time for `CachedIndicator.getValue()` was for the logger call (Logback was my SLF4j implementation in this case). To address your comment on the double-checking of the log level, that shouldn't be a concern in my opinion because odds are that if you have the `TRACE` logger level enabled, you aren't concerned about performance and are more concerned about debugging. Also, since a lot of objects in TA4j are immutable (e.g. `BaseBar` and `DoubleNum`), the garbage collector has to spend a lot more time cleaning things up during large backtests where objects are frequently allocated, used once, and then discarded. So the changes in the PR free up the GC to perform garbage collection on those immutable objects rather than wasting the GC threads' CPU time on GC'ing those redundant `Object[]`. So the memory improvement is obvious, but there is also a CPU performance improvement as well. You might even consider removing the `log.trace()` calls in production-level builds (e.g. the artifacts that are released on Maven Central). I've removed all logging calls in my local build of TA4j and the CPU time that the GC thread's use (which I monitor with `htop`) has decreased overall. The small things add up when you have a backtest that takes 16 hours to run haha.

PR #798, opened 2021-08-08
SMAIndicator use the previously cached sma value to calculate

gowsp:
Fixes #.

- The existing SMAIndicator is calculated by traversing the data. In fact, we can use the previous cached sma value to calculate and improve performance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **BarSeriesManager** removed empty args constructor
 - **Open|High|Low|Close** do not cache price values anymore
 - **DifferenceIndicator(i1,i2)** replaced by the more flexible CombineIndicator.minus(i1,i2)
+- **SMAIndicator** use the previously cached sma value to calculate
 
 ### Removed/Deprecated
 - **Num** removed Serializable
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/SMAIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/SMAIndicator.java
@@ -46,13 +46,18 @@ public class SMAIndicator extends CachedIndicator<Num> {
 
     @Override
     protected Num calculate(int index) {
-        Num sum = numOf(0);
-        for (int i = Math.max(0, index - barCount + 1); i <= index; i++) {
-            sum = sum.plus(indicator.getValue(i));
+        if (index < barCount) {
+            Num sum = numOf(0);
+            for (int i = 0; i <= index; i++) {
+                sum = sum.plus(indicator.getValue(i));
+            }
+            return sum.dividedBy(numOf(index + 1));
         }
-
-        final int realBarCount = Math.min(barCount, index + 1);
-        return sum.dividedBy(numOf(realBarCount));
+        Num count = numOf(barCount);
+        Num currentValue = indicator.getValue(index).dividedBy(count);
+        Num nPeriodsAgoValue = indicator.getValue(index - barCount).dividedBy(count);
+        Num prevValue = getValue(index - 1);
+        return prevValue.minus(nPeriodsAgoValue).plus(currentValue);
     }
 
     @Override


team172011:
Hi @gowsp I think there is one problem with indicators that depend on previous values if we do in that way: What if one value becomes NaN? If your code depends always on previous calculated values that depend on previous values again, once we get NaN the indicator will always return NaN, or not?

team172011:
One fix would be always checking for the result being NaN. And if that check is `true` we have to calculate it "the old way" traversing back to check if a value is `NaN` in the range [index - bar count, index] (returning NaN would be ok) or not (returning a result based only on data of that range).

team172011 (commented):
Hi @gowsp I think there is one problem with indicators that depend on previous values if we do in that way: What if one value becomes NaN? If your code depends always on previous calculated values that depend on previous values again, once we get NaN the indicator will always return NaN, or not?

johnthetireless:
Without getting into the NaN issue...  repeated division will accumulate rounding errors... small, but maybe someone cares that the 18th decimal place of their 200-period SMA is off by the time they reach index 21098...  a mathematically better way might be to compute a recursive summation and then divide it by N.   basically the same idea as this PR 

sketch code...

SummationIndicator(Indicator_Num_ arg, int periods)   // sum of last N values of arg

calculate(index)
if index == 0, 
   return arg.getValue(0);
else 
  previousSum = this.getValue(index - 1);
  newSum = previousSum + arg.getValue(index);  // add new value
  if ( index >= periods )   
     newSum = newSum - arg,getValue(index - periods);  // subtract oldest value
 return newSum;

This could be used lots of places beside SMA.  Chaikin Money Flow uses the summation of "money flow volume" divided by the summation of volume, for example.  The math works (at least as well this PR proposes) and its really fast.  At most one addition and one subtraction for each summation value we need to calculate, vs N additions.  (I use something like this in another code base.)  

How do things like the ta4j ChaikinMoneyFlowIndicator handle NaN?   The CLV could do a divide by zero at any index if high = low, and we up trying to do math on a NaN.   A quick look suggests CMF will return a range of NaN until the bad CLV value passes.  This is a horrible problem in general.  Getting it right in one place (like my summation sketch probably could, with lots of testing) might be a very good idea.

johnthetireless:
When does an indicator legitimately return NaN?  I have looked at hundreds of TA calculations; I have implemented many of them.  I don't think I have never seen one that is undefined for a given input, except at the beginning of a series.  (Many people would say a 10-period EMA is undefined until there are 10 values to average. ta4j does things differently, and I am not saying that's wrong.)

Maybe ta4j has some calculations I have not studied.  Can someone provide a legitimate example of when an indicator should return NaN?  If the input has no NaN - and the OHLCV does not in real life - when does some calculation introduce NaN into the mix?  This is all basic math and stats..  I am not seeing how we ever get NaN.

team172011:
> When does an indicator legitimately return NaN?

I also thought about this, maybe NaN creates to much edge cases so that you have to keep it in mind for every calculation. As far as I can see it could replaced by zero in `Analysis` or `Criterion` classes by zero.

But there is also one more key role of `NaN` in `Indicators` and `Strategies`. It represents the absence of valid data for calculations. For instance the `UnstableIndicator` returns NaN for the unstable period or some other Indicators return NaN if the "barCount" lower than the current index. If you apply your sma with previous cached calculations on this kind of indicators you would get into trouble.

nimo23:
>As far as I can see it could replaced by zero in Analysis or Criterion classes by zero.

`NaN` cannot be replaced by `0` because it's mathematical not identical. There is a difference when doing `1/0` and `0/1. =>  1/0=NaN != 0.

NaN has its "right to exist".

johnthetireless:
I will suggest that all indicators are defined to handle stuff like division by zero.  Indicators were largely made to be plotted, after all.  It might be hard to find a reference that goes into all the edge cases, though.  In those cases we as implementers can make a reasonable choice; zero being the most obvious.  As team suggests, letting NaN into the mix opens a whole can of worms.    We simply should never try to do 1 / 0 when calculating indicator values.  Doing so is a programming error, IMO.

BTW, I don't consider the Unstableindicator a real TA calculation.  I am sure it has its uses, but it shouldn't affect "real" calculations, I don't think.  Also, I don't recall seeing any indicators that return NaN when index < barCount.  Even the PreviousValue returns something at index 0.  Again, I have not studied them all.  If at all possible, I would strongly suggest having all indicators always return a real value.  (If previous(close,0) == close(0) is not contrived, I don't know what is...  a "made up" result at index 0..N-1 is computationally reasonable, if previous(0) is to be allowed,  IMO)  SMA won't work as proposed if there are NaNs in the middle.  I would be surprised if the test suite covers all the edge cases where indicator X must handle NaN from indicator Y at index whatever...  Kudos if it does.,

I so not dispute the need for something like NaN for some financial calculations.  Many mathematical operations are undefined for certain inputs  SQRT(-1) won't work with real numbers, for example.  I am much less with concerned with gain/loss and calculations like that.  NaN is actually a nice little implementation; likely better than using nulls or Optionals, IMO.  But Num itself is just too damn hard to use (where is the bar series so I can make a numOf(x)?) and expensive to create.  I digress...

PR #809, opened 2021-10-15
Additional context object to org.ta4j.core.Position

easyappsstudio:
Added an additional context object to `org.ta4j.core.Position` that can be used or extended by developers to create dynamic rules.

Two additional rules that make use of the new `Position`´s object have been added:
- `org.ta4j.core.rules.FixedPriceStopLossRule`
- `org.ta4j.core.rules.FixedPriceStopGainRule`

Additional test classes for the two new rules have been added as well as new tests in `org.ta4j.core.PositionTest`.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/CustomPositionData.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/CustomPositionData.java
@@ -1,0 +1,69 @@
+package org.ta4j.core;
+
+import org.ta4j.core.num.Num;
+
+/**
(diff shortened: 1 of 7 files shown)


easyappsstudio:
Hello @nimo23 
The `CustomPositionData` object acts as a simple extension point since TA4J has not been engineered for extensibility: extending `Position` means also reimplementing a completely new `TradingRecord`. Things get ugly pretty quickly.

Also, the two additional rules implement a behavior which is currently not possibile without the `CustomDataObject`, prividing added value and a compelling example for developers on how to use the extension point.

PR #812, opened 2021-11-28
A "fluent decorator" for indicators and some facades.

johnthetireless:
Fixes: nothing

- NumericIndicator: a fluent decorator
- facades for Aroon, Bollinger bands and some others
- an experimental FIFO cache for indicators and/or bar series
- this is all for review; I don't intend to have this merged as it is; we can discuss packages and such...

Look at the demos in the indicators.numeric.facades package first.  The DMSDemo is a good place to start.


team172011:
I like this design, writing things that way looks very nice and clear! 😍
```
       Aroon aroon = new Aroon(bs, 25);

         // inspired by
         // https://school.stockcharts.com/doku.php?id=technical_indicators:aroon

         Rule previousAroonConsolidation = aroon.up()
                 .previous(5)
                 .isLessThan(20)
                 .and(aroon.down().previous(5).isLessThan(20));

         Rule aroonUpBreak = aroon.up()
                 .isGreaterThan(50)
                 .and(aroon.down().isLessThan(50))
                 .and(previousAroonConsolidation);

         Rule aroonDownBreak = aroon.down()
                 .isGreaterThan(50)
                 .and(aroon.up().isLessThan(50))
                 .and(previousAroonConsolidation);
```

Do you think things get mixed up to much because with this PR there would be a lot of different ways to create the same indicator or rule? Also some Indicators may become redundant like the `SumIndicator` or not?

ta4j-examples/src/main/java/ta4jexamples/indicators/numeric/NumericIndicator.java
```diff
@@ -0,0 +1,232 @@
...
+ * <p>
+ * Methods like sma() and ema() simply create the corresponding indicator
+ * objects, (SMAIndicator or EMAIndicator, for example) with "this" as the first
+ * argument. These methods usually instantiate cached objects.
+ * <p>
+ * Another set of methods, like crossedOver() and isGreaterThan() create Rule
+ * objects. These are also overloaded to accept both Indicator<Num> and Number
+ * arguments.
+ */
+public class NumericIndicator implements Indicator<Num> {
```
nimo23, line 62: please rename it to `NumIndicator`.

johnthetireless: no...  :)   my name is nicer, if a little bit longer to type...  we should use user-facing (public API) names that reflect domain concepts or design intent, not implementation details...  I am old, my friend... I have struggled with choosing good names most of my career...  read it aloud....  a listener is going to stop you and ask what "num" means, even if they know what an "indicator" is

nimo23: @johnthetireless Even if you _feel_ it's nicer we should match and comply to this ta4j's naming scheme and ta4j is using `Num`. The name is already spread and explained over all this lib. It's irritating when we introduce again a new name for number/numeric. And looking at what you are implementing `implements Indicator<Num>`, you do nothing more than using the `Num` as an **indicator**. So would be nice to name the thing what it is: a `NumIndicator`.

johnthetireless: I am not going to change it.  If this PR is accepted you are certainly free to change it then.

nimo23: @johnthetireless I see that it resides within the  `numeric`-package and looking again I must confess, that `NumericIndicator` is indeed the right name for that. I also like it now:)

nimo23 (approved):

johnthetireless:
> I like this design, writing things that way looks very nice and clear! 😍
> Thanks.  I use this approach extensively in another code base.  It has a really cool "expression language" if I may be less than humble; very close to a DSL for technical analysis.  I will attach a snippet if I can.
> 
> Do you think things get mixed up to much because with this PR there would be a lot of different ways to create the same indicator or rule? Also some Indicators may become redundant like the `SumIndicator` or not?

Long term, I think we should merge my ideas into a future major release.  My design is very different than ta4j as it stands...

Short term, we can easily extract the NumericIndicator class itself and decide which "fluency" methods we want to keep.
All the rule creation methods (crossedOver, greaterThan, etc) should stay; they make rule creation nicer.

Methods like plus() and minus() create functional objects that do the same things as SumIndicator, etc with no caching.
I think these are OK as long as users know the difference.  Keep the CombineIndicator or whatever it called and offer my light weight functional methods for cases when you want to do a little math and there is no existing indicator.  For example...

Volatility = StandardDeviation / SMA

As for making existing indicators redundant....

Many (most) existing indicators could be rewritten.  My approach leads to much more readable code.  I am not suggesting you do this.  Authors of new indicators can use my style if they wish.

Immediate suggestion....

I believe the best parts of this should go into ta4j-core.  With your assent I will create another PR to add 8 classes into 2 new packages.

org.ta4j.core.indicators.numeric 
- NumericIndicator
- PredicateOperation (only has isZero now, but likely useful in general; we can hide it if you want)
- BinaryOperation (package-private)
- UnaryOperation (package-private)

org.ta4j.core.indicators.numeric.facades
- Aroon
- BollingerBands
- KeltnerChannels
- MACD

These could all go into the utils package if you prefer.  They are in no way related to the BarSeriesUtils class that exists there now, so I am against this.

I can add a couple of demos to the examples project too.

At some point I may make a draft PR for caching.  I use the FIFO design extensively, with great success.
I know I can make a FIFO bar series and FIFO indicator, better suited for a potentially infinite series of bars/values.
We only need to keep the last N (1024 works well for me).

nimo23:
>I believe the best parts of this should go into ta4j-core. With your assent I will create another PR to add 8 classes into 2 new packages.

Yes, I also like the NumericIndicator into the core. Please create a new PR to submit the all these classes into the `indicator.numeric`-package

>These could all go into the utils package if you prefer. 

Please keep them all into the `numeric.facades` package.

johnthetireless:
OK  I am going to remove the PredicateOperation too.  It doesn't provide enough value to warrant its API footprint.
We can add it again if there is more than the "isZero" use case.

PR #814, opened 2021-12-02
Implemented isNaN in DoubleNum and DecimalNum.

jfarjona:
Fixes #.

- Implement isNaN() on DecimalNum by checking if the delegate is a NaN using Double.NaN().
- Implement isNaN() on DoubleNum by checking if the delegate is NaN using Double.NaN().

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **Fixed** **`ChaikinOscillatorIndicatorTest`**
 - **DecimalNum#remainder()** adds NaN-check 
 - **Fixed** **ParabolicSarIndicatorTest** fixed openPrice always 0 and highPrice lower than lowPrice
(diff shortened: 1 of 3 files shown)


team172011:
@jfarjona I think this change is not correct. The `Num#isNaN()` check should only return true if the class is an instance of `org.ta4j.core.num.NaN`. Not if the delegate value is any kind of NaN implementation

jfarjona:
If that is the case, the documentation should be very explicit on that; and when numOf(Double.NaN) is called it should return a NaN instance.
Also when operations return a NaN , the instance should be converted to NaN…

Juan

team172011:
> Also when operations return a NaN , the instance should be converted to NaN…

this is already the case, you can only return NaN if you return an NaN instance

> and when numOf(Double.NaN) is called it should return a NaN instance.

I think once cannot make an assumption that the specific NaN representation of the delegate means the same as the NaN representation of ta4j. If the delegate becomes NaN it may behave in another way as the ta4j NaN.

PR #817, opened 2021-12-19
Feature/numeric indicator core

johnthetireless:
- NumericIndicator added to ta4j-core
- facades for Aroon, Bollinger, and Keltner 
- minor numOf fixes


johnthetireless:
The build is failing because new demo classes are not resolving references to new core classes.
AroonDemo fails to find NumericIndicator, for example.
This builds OK locally; all tests pass fine.
I suspect the newly built demo classes are not getting the just built core classes in the auto build.
If there is something I can do to make this auto build, please let me know. 
Otherwise I will delete all the examples.

johnthetireless:
So...

I have stripped this down to the NumericIndicator itself and 3-4 simple facades.  
The NumericIndicator uses a couple of other package private classes.

I removed my demos.  They would not build here.  
I would like to provide something so people know this exists and what it can do.
After this stuff is merged I will do that,  I hate git.  

Here is the code base where I have developed a lot of my ideas.  If you liked the NumericIndicator you will probably like this stuff too.  We can merge some of this, probably...  long term 

johnthetireless/ta-expressions

I am not sure where I will take this code base; its mostly just a dump from my local file system right now.
But you are free to look at how I have solved a lot of the problems (I see) with ta4j.   

Here is the FIFO mechanism I use.   ta4j needs something like this, IMO. 

https://github.com/johnthetireless/ta-expressions/blob/main/ta-expressions-core/src/main/java/ta/expressions/app/FIFOSequence.java

This simple little thing works like a charm to keep the last N objects in memory.
I obviously haven't tried, but I am pretty sure this could consume input forever,
until your JVM died for some other reason.  Live trade bitcoin for months, probably.
It won't "thrash" the way a list-backed implementation will, deleting oldest to add new items.
This guy just replaces the oldest in an array.  I don't see how anything could be faster (or simpler).

ta4j-core/src/main/java/org/ta4j/core/indicators/AroonOscillatorIndicator.java
```diff
@@ -35,15 +35,20 @@
  */
 public class AroonOscillatorIndicator extends CachedIndicator<Num> {
 
+    public static final String KEYWORD = "AroonOscillator";
```
nimo23: Such things will increase the memory consumption on every instance creation beforehand. I am against this. If the user needs a string representation, then it should be within `toString()`. Btw, those public `public static final String KEYWORD`-things has nothing to do with this main PR and should not be included here.

johnthetireless: I will remove anything not directly related to this PR.

I don't think there is any difference between putting the string in a static field or hard-coding it in the toString().  The string "AroonOscillator" will exist once in the JVM one way or the other; I suspect all JVMs intern strings.   Caching the value to be returned by toString() is a common approach when performance matters.  It likely doesn't in this case.

As for increasing memory consumption...  consider this... the indicators use each other as children so often, duplicating array lists and calculations, that very little I will do will be important...  I will remove it anyway.  cheers

nimo23: @johnthetireless Thanks. Btw, you can use **JOL** (`org.openjdk.jol`)  to measure object sizes or create a heap dump and analyze by "eclipse memory analyzer"(https://www.baeldung.com/jvm-measuring-object-sizes).

johnthetireless: Thanks.  I have looked at the memory analyzer briefly but I haven't used it for an actual dump yet.
My little "representation" field would have added the size of a reference/pointer to the indicator object: 8 bytes, probably.  The string itself would be interned somewhere by the JVM it seems...

		public static void main(String[] args) {
			String a = "AroonOscillator";
			String b = "AroonOscillator";
			System.out.println(a.equals(b));  	// true, of course
			System.out.println(a == b);		// also true; my JVM interns strings			
		}

team172011 (approved):
I would approve this PR. I am just unsure about the location of the "facades", maybe it would better to rename them to ***Facade and put them into the corresponding package like:

- BollingerBandsFacade.java in org/ta4j/core/indicators/bollinger
- ArronFacade.java in (new) org/ta4j/core/indicators/aroon package with other Aroon indicators
- KeltnerChannelFacade.java in org/ta4j/core/indicators/keltner
- MACDFacade.java (new) inorg/ta4j/core/indicators/macd with MACDIndicator

ta4j-core/src/main/java/org/ta4j/core/indicators/numeric/UnaryOperation.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.function.UnaryOperator;
+
+import org.ta4j.core.BarSeries;
+import org.ta4j.core.Indicator;
+import org.ta4j.core.num.Num;
+
+/**
+ * Objects of this class defer the evaluation of a unary operator, like sqrt().
+ * 
+ * There may be other unary operations on NUm that could be added here.
```
nimo23: write `Num` instead of `NUm`..

ta4j-core/src/main/java/org/ta4j/core/indicators/numeric/BinaryOperation.java
```diff
@@ -0,0 +1,91 @@
...
+
+import org.ta4j.core.BarSeries;
+import org.ta4j.core.Indicator;
+import org.ta4j.core.num.Num;
+
+/**
+ * Objects of this class defer evaluation of an arithmetic operation.
+ * 
+ * This is a lightweight version of the CombineIndicator; it doesn't cache.
+ *
```
nimo23: remove last (empty) line in comments.

nimo23:
>I am just unsure about the location of the "facades", maybe it would better to rename them to ...Facade.java and put them into the corresponding package

Yes, good idea. But consider, that if someone writes a facade for an indicator which is not placed within its own package, then a new package must be created to put this indicator and its facade into the new package (this would lead to breaking changes). However, I don't think, that this is a problem..

johnthetireless:
> I would approve this PR. I am just unsure about the location of the "facades", maybe it would better to rename them to ...Facade.java and put them into the corresponding package like:
> 
> * `BollingerBandFacade.java` in **org/ta4j/core/indicators/bollinger**
> * `ArronFacade.java in` (**new**)**org/ta4j/core/indicators/aroon** package with other Aroon indicators
> * `KeltnerChannelFacade.java` in **org/ta4j/core/indicators/keltner**
> * `MACDFacade.java` (**new**) in **org/ta4j/core/indicators/macd** with MACDIndicator

Moving the existing aroon and macd indicators is a breaking change, as nimo suggests.  I question whether these simple "facades" are worth it.

Perhaps we should remove the facades and put them in the demo project.  I will use them to write a couple of nice demos. 
Restructuring your packages for these things might not be a good idea at this point.  Your call.

team172011:
> Moving the existing aroon and macd indicators is a breaking change, as nimo suggests. I question whether these simple "facades" are worth it.
Perhaps we should remove the facades and put them in the demo project. 

True, moving them to the examples project would be another option. But especially for indicators that are almost always used together (like Bollinger bands or keltner channels) I like this easy way of providing all needed lines with help of only one object.

> Restructuring your packages for these things might not be a good idea at this point. Your call.

I thought a lot about breaking the API here, on the one hand it will of curse always cause trouble for existing projects that want to update, but on the other hand we already have a lot of breaking changes in [this release](https://github.com/ta4j/ta4j/blob/develop/CHANGELOG.md) and ta4j is still in a "version zero" (0.14) so always insuring downward compability in this early state would pollute the api more and more.

johnthetireless:
OK.  I will move the 3 Aroon indicators to a new package and put AroonFacade there.  BollingerBandFacade and KeltnerChannelFacade will go into the existing packages.  The macd facade will go away.  Its effect can easily be achieved like this.

     NumericIndicator macd = NumericIndicator.of(new MACD(bs, 12, 26));
     Indictor signal = macd.ema(9);
     Indicator histogram = macd.minus(signal) ;

Using the facade I wrote takes just as many lines and doesn't read much better.

johnthetireless:
> I thought a lot about breaking the API here

I have moved things as suggested.  Please have a look. The message below tells me there are merge conflicts but also tells me I cannot resolve them without write access.

team172011:
>The message below tells me there are merge conflicts but also tells me I cannot resolve them without write access. 

maybe because of updated license headers. You could resolve the conflicts be merging the current stat of the `develop` branch into this feature branch locally.

>Using the facade I wrote takes just as many lines and doesn't read much better.

I agree, macd facade would be a good case for the example project. The library should only contain really helpful facades

team172011:
> You could resolve the conflicts by merging the current stat of the develop branch into this feature branch locally.

I tried and it worked :)

johnthetireless:
> > You could resolve the conflicts by merging the current stat of the develop branch into this feature branch locally.
> 
> I tried and it worked :)

Thanks

team172011 (commented):
We need unit tests at least for:

- ta4j-core/src/main/java/org/ta4j/core/indicators/numeric/UnaryOperation.java
- ta4j-core/src/main/java/org/ta4j/core/indicators/numeric/NumericIndicator.java
- ta4j-core/src/main/java/org/ta4j/core/indicators/numeric/BinaryOperation.java

johnthetireless:
> We need unit tests...

The Binary and UnaryOperation classes cannot be instantiated outside the numeric package.  Objects of these classes are only created by NumericIndicator methods.  Testing the public methods like plus() and sqrt() will test these operation objects.

I see a couple of testing approaches have been used.  Some tests use a mock bar series and others use Excel files.   I do not have Excel; I doubt I can create files for that kind of test.

There are a lot of methods in the NumericIndicator class.   Most of them do simple math; tests such as adding 2 "mock indicators" at index 0 are easy and sufficient, IMO.  A few methods simply create other ta4j indicators.  I can test to make sure they create the right classes; I think this is enough.  

Thoughts?

team172011:
I pushed a draft to this PR [commit](https://github.com/ta4j/ta4j/pull/817/commits/f61e5ae32ee0aaf19750d13ed9c10de8d3ce9ea7)

johnthetireless:
> I pushed a draft to this PR [commit](https://github.com/ta4j/ta4j/pull/817/commits/f61e5ae32ee0aaf19750d13ed9c10de8d3ce9ea7)

I did a git pull on my end and I have your new file.  Thanks; this is very helpful.  I will fill in the tests.

johnthetireless:
There are probably 10-12 new indicators I can define for ta4j when this is merged.     
I will look through my little lib and see what's easy to do for the ta4j demo project.
Here is a simple one...  

	//Force Index -- https://school.stockcharts.com/doku.php?id=technical_indicators:force_index
	BarSeries bs = ...
	NumericIndicator close = NumericIndicator.closePrice(bs);
	NumericIndicator change = close.minus(close.previous));
	NumericIndicator volume = NumericIndicator.volume(bs);
	Indicator<Num> forceIndex = change.multipliedBy(volume).ema(21);

ta4j-core/src/main/java/org/ta4j/core/indicators/numeric/BinaryOperation.java
johnthetireless: > remove last (empty) line in comments.

I think the javadoc is clean now

nimo23:
>It won't "thrash" the way a list-backed implementation will, deleting oldest to add new items.

Not directly related to this PR, but a good point!

@johnthetireless The link mentioned in https://github.com/ta4j/ta4j/pull/817#issuecomment-999616018 is broken, I wanted to have a look at it..maybe we can use this in ta4j internally or improve ta4j that way.

nimo23:
@johnthetireless Closing a PR for purely personal and non-factual reasons in which the community, especially @team172011  has already invested **so much time**  is extremely rude. Btw, you do benefit in using this project and we also spend much time in improving the code.

nimo23:
@team172011 did you have saved this branch in your local repo? Maybe you can recover this PR from your local branch to build a new PR similar to this. I think, the `NumericIndicator` would be a great addition to this project. Can you create  a new PR similar to this one?

team172011:
> Can you create a new PR similar to this one?

Done with #832

PR #827, opened 2022-01-07
Update BarSeriesManager.java: consider finishIndex

nimo23:
- If `finishIndex` is set, then closing the last position should **NOT** be possible, if it happens after finishIndex.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/BarSeriesManager.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BarSeriesManager.java
@@ -172,10 +172,9 @@ public class BarSeriesManager {
             }
         }
 
-        if (!tradingRecord.isClosed()) {
+        if (!tradingRecord.isClosed() && finishIndex == barSeries.getEndIndex()) {
             // If the last position is still opened, we search out of the run end index.
-            // May works if the end index for this run was inferior to the actual number of
-            // bars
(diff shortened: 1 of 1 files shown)


nimo23:
@team172011 please check if my solution is right.

For example:

If `startIndex = 0 and finishIndex = 10`, then the tradingRecord should only contain trades within the range of 0 till 10. But this is actually not the case. Actually, when trying to close a position, then it also tries to close the position even if it happens after the `finishIndex`. But if the user sets a `finishIndex`, then the user expects that trades should **not** occur after the `finishIndex` - this is currently not the case and such trades are **wrongly** recorded in the `tradingRecord`. Thus, I changed the code, please check if my solution is right.

nimo23:
I came up with two versions, however, I don't know which one, if any, is correct:


```
..
int runEndIndex = Math.min(finishIndex, barSeries.getEndIndex());
..
if (!tradingRecord.isClosed()) {
    // If the last position is still opened, we search out of the run end index.
    // It may work if the end index for this run was inferior to the actual number of bars.
    int seriesMaxSize = Math.min(runEndIndex, Math.max(barSeries.getEndIndex() + 1, barSeries.getBarData().size()));
..
}
```


or

if (!tradingRecord.isClosed() && finishIndex == barSeries.getEndIndex()) {
..
}

nimo23:
replaced by https://github.com/ta4j/ta4j/pull/882

PR #834, opened 2022-01-17
add indices to TradingRecord to track start and end recording

nimo23:
- added `TradingRecord#getStartIndex()` to track the start of the recording 
- added `TradingRecord#getEndIndex()` to track the end of the recording 
- correct `EnterAndHoldReturnCriterion`
- rename `VersusBuyAndHoldCriterion` to `VersusEnterAndHoldCriterion`
- correct `VersusEnterAndHoldCriterion`
- ability to add `TradeType` to `VersusEnterAndHoldCriterion`

Using https://github.com/ta4j/ta4j/blob/9a3bb213364424f92ce0fd9501f47e45b2c33c21/ta4j-core/src/main/java/org/ta4j/core/BarSeriesManager.java#L158 

with:
- `startIndex` > series.getBeginIndex()
- `endIndex` < barSeries.getEndIndex()

returns a tradingRecord within `startIndex` and `endIndex`, however, all those Criteria which calculates stuff within the whole barSeries (from `series.getBeginIndex()` till `barSeries.getEndIndex()`) are not aware of the fact that the tradingRecord was only run within `startIndex` and  `endIndex` and should return its analyzes only within this segment. 

For example, the following Criteria:

- VersusBuyAndHoldCriterion
- EnterAndHoldReturnCriterion
- ..

returns misleading values because it assumes the **full** barSeries instead of only restricting its calculation only from `startIndex` till  `endIndex`. Of course, I can use `series.subSeries(startIndex, endIndex)`, but then the indices are changed which is not always appropriate. 

The tradingRecord must be analyzed correctly by any criterion if I use custom `startIndex` and  `endIndex`.

We can provide two properties within the TradingRecord-interface
- endIndex with `TradingRecord#getEndIndex()`
- startIndex with `TradingRecord#getStartIndex()`

those properties will be set within the run-method.

With this, we can access the real startIndex and endIndex of the TradingRecord within any Criteria. What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 ### Breaking
 - **DifferencePercentage** renamed to **`DifferencePercentageIndicator`**
 - **BuyAndHoldCriterion** renamed to **`EnterAndHoldCriterion`**
+- **VersusBuyAndHoldCriterion** renamed to **`VersusEnterAndHoldCriterion`**
 - **DXIndicator** moved to adx-package
 - **PlusDMIndicator** moved to adx-package
 - **MinusDMIndicator** moved to adx-package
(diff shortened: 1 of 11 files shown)


nimo23:
@team172011 I already di

PR #835, opened 2022-01-17
TradingRecord: add startIndex/endIndex

nimo23:
- added `TradingRecord#getStartIndex()` to track the start of the recording 
- added `TradingRecord#getEndIndex()` to track the end of the recording
- (includes some minor code cleanups (javadoc, replaced IllegalArgumentException with `Objects.requireNonNull`))

to correct the calculations of all Criteria when custom startIndex and endIndex is used. All criteria which make use of `barSeries.getBeginIndex `and `barSeries.getEndIndex` must be adapted to use the right startIndex and endIndex.


nimo23:
Using https://github.com/ta4j/ta4j/blob/9a3bb213364424f92ce0fd9501f47e45b2c33c21/ta4j-core/src/main/java/org/ta4j/core/BarSeriesManager.java#L158 

with:
- `startIndex` > series.getBeginIndex()
- `endIndex` < barSeries.getEndIndex()

returns a tradingRecord within `startIndex` and `endIndex`, however, all those Criteria which calculates stuff within the whole barSeries (from `series.getBeginIndex()` till `barSeries.getEndIndex()`) are not aware of the fact that the tradingRecord was only run within `startIndex` and  `endIndex` and should return its analyzes only within this segment. 

For example, the following Criteria:

- VersusBuyAndHoldCriterion
- EnterAndHoldReturnCriterion
- ..

returns misleading values because it assumes the **full** barSeries instead of only restricting its calculation only from `startIndex` till  `endIndex`. Of course, I can use `series.subSeries(startIndex, endIndex)`, but then the indices are changed which is not always appropriate. 

The tradingRecord must be analyzed correctly by any criterion if I use custom `startIndex` and  `endIndex`.

We can provide two properties within the TradingRecord-interface
- endIndex with `TradingRecord#getEndIndex()`
- startIndex with `TradingRecord#getStartIndex()`

those properties will be set within the run-method.

With this, we can access the real startIndex and endIndex of the TradingRecord within any Criteria. What do you think?

nimo23:
If this PR is accepted, then we must look for all existing criteria where to adapt this `startIndex` and `endIndex`. For example, `MaximumDrawdownCriterion` also uses "series.getBeginIndex()" and "series.getEndIndex()" to calculate its cashFlow.

nimo23:
@team172011 This PR is finished (except the license error build which should be solved by you).

nimo23:
@team172011 could you please review this PR as soon as possible. I want to create other PRs with new criteria that makes use of the capabilities of this PR.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **MinusDMIndicator** moved to adx-package
 - `analysis/criterion`-package moved to root
 - `cost`-package moved to `analysis/cost`-package
+- **VersusBuyAndHoldCriterion** renamed to **`VersusEnterAndHoldCriterion`**
```
team172011: I think this breaking change should not be part of this PR

team172011 (changes requested):

CHANGELOG.md
nimo23: Ok. I did https://github.com/ta4j/ta4j/pull/885. Please merge this PR after https://github.com/ta4j/ta4j/pull/885.

PR #837, opened 2022-01-20
Rename pnl files

nimo23:

- **ProfitLossCriterion** renamed to **`NetProfitLossCriterion`**
- **ProfitLossPercentageCriterion** renamed to **`NetProfitLossPercentageCriterion`**
- **ProfitLossRatioCriterion** renamed to **`GrossProfitLossRatioCriterion`**
-  small javadoc corrections


team172011 (changes requested):
This breaking changes do not really add comparable value

nimo23:
>This breaking changes do not really add comparable value

Unfortunately, the names are actually vague, because you cannot determine from looking only at the class name what it does (It could be GrossPnl or NetPnl), because the class name is not specific. And we also use Net and Gross as Prefix for the other classes, so it would suite to the naming scheme. 

I also think, that closing a PR immediately without any discussion is not constructive! I think, I will create a new issue for that, so the community can participate into this decision.

PR #838, opened 2022-01-20
Num: rename multipliedBy() and dividedBy()

nimo23:
- **Num#multipliedBy** renamed to **`Num#multiply`**
- **Num#dividedBy** renamed to **`Num#divide`**

It's common to say `multiply` and `divide` instead of the puffy words `multipliedBy` and `dividedBy`. Btw, BigDecimal also calls it `divide` and `multipy`. It's better and it also eliminates a lot of characters in the ta4j-source-code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 ### Breaking
 - **DifferencePercentage** renamed to **`DifferencePercentageIndicator`**
 - **BuyAndHoldCriterion** renamed to **`EnterAndHoldCriterion`**
+- **Num#multipliedBy** renamed to **`Num#multiply`**
+- **Num#dividedBy** renamed to **`Num#divide`**
 - **DXIndicator** moved to adx-package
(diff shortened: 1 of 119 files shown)


team172011 (changes requested):
Same for this PR. I do not agree with breaking changes in hundreds of files only for a small method renaming. We could do this kind of renaming for hundreds of classes and methods all the time ...

nimo23:
@team172011 don't you think that especially those two operators deserves smaller names? We use them all over the places and using it has only benefits (less keystrokes, less characters, less source code size,..) **and we are still in version 0.14**, so breaking changes, as you also said, is justifiable. 

And it has the side effects, that the community needs to look into this project..btw,  it wouldn't bother anyone anyway, on the contrary, because nobody (expect you) mentioned it, that this breaking change causes any problems.

nimo23:
@team172011 I also think, that closing a PR immediately **without any discussion** is not constructive! I think, I will create a new issue for that, so the community can participate into this decision.

nimo23:
>We could do this kind of renaming for hundreds of classes and methods all the time ...

Yes, but the renamings especially for those two methods are reasonable because in long term it only has benefits.

PR #844, opened 2022-02-11
WaitForRule: add noWaitIfNoTradeExists

nimo23:
Fixes #.

- :tada: **Enhancement** added **`WaitForRule#noWaitIfNoTradeExists`**  (i.e. no need to wait if no trade exists).


team172011:
In my opinion this rule seems to be too specific to be part of the library.

PR #846, opened 2022-02-20
Add constructor for double input

romanbsd:
The Double.parseDouble(i.toString()) is a considerable performance hit.
This surfaced when doing profiling to our code, and I couldn't figure why there was no explicit constructor with a double argument.


team172011:
@romanbsd PRs to the master branch will be rejected. Please open a PR to the develop branch.

Why are you not using the existing `DoubleNum.valueOf(Number)` method? 
``` java
        DoubleNum doubleNum = DoubleNum.valueOf(3d);
        DoubleNum dDoubleNum = DoubleNum.valueOf(Double.valueOf(3d));
```

romanbsd:
DoubleNum.valueOf(Number) does Double.parseDouble(i.toString()) which creates a noticeable performance hit for no reason.

PR #864, opened 2022-06-21
Examine DoubleNum performance + Misc performance intellij inspections

jnorthrup:
I am working with tens of billions of candles.  I wanted to peek into the internals and see if there was anything that might make a difference for performance.  

intellij catches the following: deletion of bars/results by subList range, an enum list that pushes the varargs into EnumSet without removing the original ctor, and I took an interest in DoubleNum math performance, and am testing a few things ~~not submitted here yet~~.

DoubleNum references are created with a sort of Key->Value relationship to the Double values.  that said, there is no Map or prior re-use of these refs at large.  while this may or may not make a difference with other Num variants, the DoubleNum is the express tradeoff of precision for performance and some attention has been given to scavenging existing references wherever the logic arithmetic presents an opportunity. 

Additionally a WeakHashMap cache has been introduced to potentially make hot recursive inner loops even hotter.  My initial testing on jdk 18 bears this to be the case (I see 10-20% faster loops) with `DoubleNum.useCache=true` but the garbage collection costs are significantly greater cost in the one Par-GC test given.  This cache is disabled by default for this reason but if you want to test maximum thruput with weak refs and giant heaps it could potentially decrease latency in the short term calculations.

Prior issues mentioning `CachedInstrument` efficacy also bring to light the inconsistent or unecessary caching occurring; with this DoubleNum cache enabled (and similar for larger Num delagates) it is likely that the Instrument inheritance hierarchy could be flatter in certain cases.

the gc expense comes when the DoubleNum's leave scope but the DoubleNum.cache items are still present.  this is a sort of transactional rigor, calling `DoubleNum.cache.clear()` after a series is finished but before it is GC'd is advised to avoid large GC trace slowdowns.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/BaseBarSeries.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BaseBarSeries.java
@@ -475,8 +475,8 @@ public class BaseBarSeries implements BarSeries {
         if (barCount > maximumBarCount) {
             // Removing old bars
             int nbBarsToRemove = barCount - maximumBarCount;
-            for (int i = 0; i < nbBarsToRemove; i++) {
-                bars.remove(0);
+            if (nbBarsToRemove > 0) {
+                bars.subList(0, nbBarsToRemove).clear();
             }
             // Updating removed bars count
             removedBarsCount += nbBarsToRemove;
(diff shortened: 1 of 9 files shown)


team172011:
@jnorthrup thanks, your changes look good. Can you please make a PR to the develop branch (based on a current version of the develop branch)? PRs to the master branch will not be accepted.

Issue #868, opened 2022-08-16
using BarSeries interface explodes in undocumented barlimit array indexing.


jnorthrup:
I recently attempted to integrate my ISAM backend toolkit to the BarSeries by java interface to provide the kline data on the fly from storage and ran into impassable barrier of undocumented complexity with CachingIndicator(s) for which an off-heap ISAM tookit barely has any need for.  

I threw the code out and crossed my fingers that i could proceed with unlimited in-memory arrays which did succeed, but this is a big dataset and could easily go south with the wrong recursive Indicators.  

I would love to have some guidance how to magically produce the bar data from an interface, like all data interface designs should support, and  somehow avoid the cache array indexing gotchas completely.

TheCookieLab:
@jnorthrup I understand your frustration and while it may be warranted please keep in mind that maintaining OSS is often a thank-less job...which is why I've signed up as a new team member. Here to make TA great again!

TheCookieLab:
OK I looked over the proposed change set and find it mostly reasonable, although I do have some concerns about the cache. 

If you create a new PR that includes unit test coverage, an update to the CHANGELOG, and the other requirements in the [contribution guide](https://ta4j.github.io/ta4j-wiki/How-to-contribute) then we can continue the discussion there.

TheCookieLab:
TBH I'm a bit confused by your creative grammar but am I to understand that
you're calling it quits?

TheCookieLab:
Well in that case good luck. You're welcome back if you change your mind.

PR #869, opened 2022-08-16
add Doxygen to supplement the wiki coverage

jnorthrup:
I find myself asking questions for which the wiki is topical but not thorough and the sample code is terse, sometimes diverging from the very same unit test.

with this doxygen flie i can look at all the instantiations of Bar, BarSeries, etc. and look at the sum of test-cases and examples tree graphically and narrow down to what is relevant for learning more, like with #868 

image


team172011:
@jnorthrup looks good for me, can you please make a PR to the develop branch? PRs to the master branch will not be accepted

PR #882, opened 2022-09-22
BarSeriesManager: consider finishIndex

nimo23:

- If the parameter `finishIndex` is given with`finishIndex >0 && finishIndex <barSeries.getEndIndex()` then closing the last position should NOT be possible, if it happens after `finishIndex`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangelog.com/en/1.0.0/) from version 0.9 onwards.
 
 ## 0.16 (unreleased)
+### Fixed
+- **BarSeriesManager** consider finishIndex when running backtest
 
 ## 0.15 (released September 11, 2022)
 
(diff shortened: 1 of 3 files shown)


nimo23:
For example:

If `startIndex` = 0 and `finishIndex` = 10, then the tradingRecord should only contain trades within the range of 0 till 10 (=> 11 bars). But this is actually not the case. Actually, when trying to close a position, then it also tries to close the position even if it happens after the `finishIndex`. But if the user sets a valid `finishIndex`, then the user expects that trades should not occur after the `finishIndex` - this is currently not the case and such trades are wrongly recorded in the tradingRecord. Thus, I changed the code, please check if my solution is right (maybe you find another solution which is more appropriate).

Using future bars after a valid `finishIndex`, is a kind of **look-ahead bias** (= using future data to act in past). We have to debate two points of view:

1. The actual behaviour could be senseful, because we really want to know if the last _entry-trade_ is behaving well or badly by looking at its future _exit-trade_. Without this information, the success rate of the last entry trade is unknown.

2. However, one could argue whether it's legal to use this future information within the time period being tested (since it's a "look-ahead bias") which should be avoided when doing backtests.

I think, point 2 is correct. What do you think?

_Note that the actual behaviour does only change if a **valid finishIndex** is given. Otherwise, the code does exactly what it has always done._

team172011:
I agree that the current implementation is not correct. But I am wondering why it has been implemented this way. Is there a use case that should be considered within this fix?

team172011 (approved):
Please add an entry to the CHANGELOG.md

nimo23:
> But I am wondering why it has been implemented this way. Is there a use case that should be considered within this fix?

The use case is nothing more than doing backtesting right when a `finishIndex` is given (It's wrong to record a trade which happens outside of a given `finishIndex`.)

team172011 (approved):

PR #894, opened 2022-10-05
Add positions ratio criterion

nimo23:
﻿﻿﻿﻿﻿Changes proposed in this pull request:
- replaced `WinningPositionsRatioCriterion` and `LosingPositionsRatioCriterion` by `PositionsRatioCriterion`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **WinningPositionsRatioCriterion** replaced by **`PositionsRatioCriterion`**
 
 ### Fixed
+-  **Fixed** **ParabolicSarIndicator** fixed calculation for sporadic indices
+
+### Changed
+- **BarSeriesManager** consider finishIndex when running backtest
+
+### Removed/Deprecated
+
+### Added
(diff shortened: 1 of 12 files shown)


team172011:
When merging this and #895 the design of the criterions gets more inconsistent. We would still have criterions like

I would prefer to have a consistent api for similar cases and I am unsure if breaking the api here is it worth:

**pro**
- less classes

**contra**
- instantiation of a criterion gets more complex and could be confusing with selecting parameters like `PositionFilter`
- criterions would become much more complex than they are in the current approach (PositionsRatioCriterion)
or
- decreased performance and increased code smell by checking something like `if (isLoss) do this else do that` in each calculation step (LossCriterion)
- with this pr and #895 we woul mix both approaches together and would still have the following:
    - NumberOfLosingPositionsCriterion
    - NumberOfPositionsCriterion
    - NumberOfWinningPositionsCriterion
    - NumberOfBreakEvenPositionsCriterion
    - NumberOfConsecutivePositionsCriterion
- and:
    - AverageLossCriterion
    - AverageProfitCriterion
- and:
    - WinningPositionsRatioCriterion
    - LosingPositionsRatioCriterion

team172011:
But than I would prefer to do the refactoring within one single PR and with a transparent transition. The solution with a static constructor looks good for me.

PR #907, opened 2022-10-26
First suggestion of pluggable cache

team172011:


ta4j-core/src/main/java/org/ta4j/core/indicators/RecursiveCachedIndicator.java
```diff
@@ -69,12 +70,10 @@ public T getValue(int index) {
...
             if (index <= seriesEndIndex) {
                 // We are not after the end of the series
-                final int removedBarsCount = series.getRemovedBarsCount();
-                int startIndex = Math.max(removedBarsCount, highestResultIndex);
-                if (index - startIndex > RECURSION_THRESHOLD) {
+                if (index - series.getRemovedBarsCount() > RECURSION_THRESHOLD) {
                     // Too many uncalculated values; the risk for a StackOverflowError becomes high.
                     // Calculating the previous values iteratively
-                    for (int prevIdx = startIndex; prevIdx < index; prevIdx++) {
+                    for (int prevIdx = series.getRemovedBarsCount(); prevIdx < index; prevIdx++) {
```
nimo23: I just wanted to make sure you knowingly changed this. Formerly, it was `startIndex` which can differ from `series.getRemovedBarsCount()`.

ta4j-core/pom.xml
```diff
@@ -41,8 +41,14 @@
 		<dependency>
 			<groupId>ch.qos.logback</groupId>
 			<artifactId>logback-classic</artifactId>
+			<version>1.2.9</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>com.github.ben-manes.caffeine</groupId>
+			<artifactId>caffeine</artifactId>
+			<version>2.9.0</version>
```
nimo23: Caffeine has already have version **3.1.1.** I think, you should firstly merge https://github.com/ta4j/ta4j/pull/910 and then change this version to  3.1.1.

TheCookieLab:
I thought I read in a thread somewhere that the bar series size management functionality (i.e. setting a max bar count, removing old bars, etc) was a blocker for this feature. Is that still the case here or is there something else?

team172011:
> Is that still the case here or is there something else

@TheCookieLab Usually it is possible to set a max cache size in third party caching libraries, especially in coffeine cache. 
So far this is just a suggestion from me on how we could replace the custom solution by a third party library. This is a big change and so far there as not been an appropriate discussion about this topic

ta4j-core/src/main/java/org/ta4j/core/indicators/RecursiveCachedIndicator.java
team172011: Yes, so far I am not 100% sure on how to adapt the RecursiveCachedIndicator when the CachedIndicator will be changed this way

team172011:
I updated this PR:


- The `CachedIndicator` and the `RecursiveCachedIndicator` have been removed. They will be replaced by an external cache `IndicatorValueCache` that will be used in the `AbstractIndicator`. Current cache implementations are:
    - `BaseIndicatorValueCache`
    - `NoIndicatorValueCache` 
    - **TODO**: Something like `RecursiveIndicatorCache` to avoid the stackoverflow problem (see failing tests)

Caching in `AbstractIndicator` looks like this:

```java
    @Override
    public T getValue(final int index) {
        final BarSeries series = getBarSeries();
        if (series == null || index >= series.getEndIndex()) {
            // Don't cache result if last bar or no available bar
            return calculate(index);
        }

        if (index < series.getBeginIndex()) {
            return calculate(0);
        }

        final Bar bar = series.getBar(index);
        return cache.get(bar.getEndTime(), (endTime) -> calculate(index));
    }
```

Usage example
```java
      SMAIndicator smaIndicator = new SMAIndicator(new ClosePriceIndicator(barSeries), 5); // default with an BaseIndicatorValueCache
      
      smaIndicator.clearCache(); // clear the cache
      smaIndicator.setCache(new BaseIndicatorValueCache<>(new IndicatorValueCacheConfig(300))); // set cache with custom config
      smaIndicator.setCache(new MyCustomCache()); // set custom cache implementation
      Map<ZonedDateTime, Num> values = smaIndicator.getCache().getValues(); // get all cached values
```

team172011:
It seems that my IDE has swapped a lot of imports in files that are not related to changes of this PR. I will try to revert this soon -.-

ta4j-core/src/main/java/org/ta4j/core/indicators/caching/BaseIndicatorValueCache.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Collections;
+import java.util.Map;
+import java.util.function.Function;
+
+public class BaseIndicatorValueCache<T> implements IndicatorValueCache<T> {
+
+    private final Cache<ZonedDateTime, T> cache;
+
+    public BaseIndicatorValueCache(BarSeries series) {
+        this(new IndicatorValueCacheConfig(series.getMaximumBarCount()));
```
TheCookieLab, line 42: IIRC when the bar series maximum bar count isn't set it'll default to `Integer.MAX_VALUE`. When this value is passed to Caffeine as the maximumSize it'll implement the cache using a `BoundedLocalCache` (a relatively heavy-weight object) as opposed to an `UnboundedLocalCache` (a light-weight wrapper around ConcurrentHashMap). In practice I don't know how much performance impact it'll have but if we have a lot of indicators active it could be something. 

Adding a check here (or perhaps within the `IndicatorValueCacheConfig` constructor) to identify the `Integer.MAX_VALUE` scenario and then either not set Caffeine's maximumSize or pass -1 (what it uses for unbounded) could be worthwhile.

team172011: Good point, using the `unboundedLocaleCache` seems to be the better default option here

team172011:
> What about the old caching implementations. If you include the old caching logic into another `IndicatorValueCache`-implementation within
> 
> caching/`NativeIndicatorValueCache` caching/`NativeRecursiveIndicatorValueCache` _(with changes as needed, e.g. `endTime` instead of `index`)_
> 
> We need this, to
> 
> * have a controllable transition
> * have a native cache implementation (without 3rd party)
> * to be able to compare benchmarks (between native and caffein)
> * to especially have **"regression tests"** while making this PR) to the new caching provider
> 
> To clarify in steps:
> 
> (1) create `NativeIndicatorValueCache`, `NativeRecursiveIndicatorValueCache`, then use that caching and let the tests run. (2) When tests are sucessfully, then change the caching provider to use the `BaseIndicatorValueCache` and `BaseRecursiveIndicatorValueCache` and let the tests run. (3) When tests are successfully, then create a benchmark to compare both caching providers. It would be bad if the newer caffeine cache would be now slower than before (with the native version).
> 
> **We need to prove that the new caching implementation is not only correct, but also not measurably slower or more memory intensive!**

I added two further cache implementations:

    - `NativeIndicatorValueCache` the former cache using custom circular array implementation for caching
    - `NativeRecursiveIndicatorValueCache` the former recursive cache extending the `NativeIndicatorValueCache`

and added the `NativeRecursiveIndicatorValueCache` to all indicators that where extending the `RecursiveCachedIndicator`.  All tests are passing now

ta4j-core/src/main/java/org/ta4j/core/BacktestExecutor.java
```diff
@@ -25,7 +25,6 @@
 
 import java.util.List;
 import java.util.stream.Collectors;
-
```
nimo23, line 28: Please do **not** remove the "blank lines" between the "package groups" (also called "top level packages") as this helps to visualize the groups better. For example, in Eclipse these "blank lines" are automatically added as a "visual separator" when using "Source->Organize Imports" - this is common and best practice. I think, intelliJ has something similar (it's found in File/Settings menu, under something like "Editor/Code Style/Java, -> "Import Layout" section)

ta4j-core/src/main/java/org/ta4j/core/indicators/CachedIndicator.java
```diff
@@ -21,27 +21,17 @@
...
- * Cached {@link Indicator indicator}.
- *
- * <p>
- * Caches the calculated results of the indicator to avoid calculating the same
- * index of the indicator twice. The caching drastically speeds up access to
- * indicator values. Caching is especially recommended when indicators calculate
- * their values based on the values of other indicators. Such nested indicators
- * can call {@link #getValue(int)} multiple times without the need to
- * {@link #calculate(int)} again.
- */
```
nimo23, line 43: Please do not remove the existing javadocs of this class.

ta4j-core/src/main/java/org/ta4j/core/indicators/CachedIndicator.java
```diff
@@ -52,40 +42,20 @@ public abstract class CachedIndicator<T> extends AbstractIndicator<T> {
      */
     protected int highestResultIndex = -1;
 
-    /**
-     * Constructor.
-     *
-     * @param series the bar series
-     */
```
nimo23, line 59: Please do not remove the javadocs of this class.

ta4j-core/src/main/java/org/ta4j/core/indicators/RecursiveCachedIndicator.java
```diff
@@ -21,55 +21,35 @@
...
- *
- * <p>
- * Recursive indicators should extend this class.
- * 
- * <p>
- * This class is only here to avoid (OK, to postpone) the StackOverflowError
- * that may be thrown on the first getValue(int) call of a recursive indicator.
- * Concretely when an index value is asked, if the last cached value is too
- * old/far, the computation of all the values between the last cached and the
- * asked one is executed iteratively.
```
nimo23, line 40: Please do not remove the javadocs of this class.

ta4j-core/src/main/java/org/ta4j/core/indicators/RecursiveCachedIndicator.java
```diff
@@ -21,55 +21,35 @@
...
- */
-public abstract class RecursiveCachedIndicator<T> extends CachedIndicator<T> {
+public class NativeRecursiveIndicatorValueCache<T> extends NativeIndicatorValueCache<T> {
 
-    /**
-     * The recursion threshold for which an iterative calculation is executed.
-     * 
-     * TODO: Should be variable (depending on the sub-indicators used in this
```
nimo23, line 47: Do not remove the javadocs of this class.

ta4j-core/src/test/java/org/ta4j/core/indicators/bollinger/BollingerBandWidthIndicatorTest.java
```diff
@@ -38,6 +35,8 @@
 import org.ta4j.core.mocks.MockBarSeries;
 import org.ta4j.core.num.Num;
 
+import static org.ta4j.core.TestUtils.assertNumEquals;
```
nimo23, line 38: Please organize the imports according common practice. As it was before, "Static imports" should be always at the top and not at the bottom.  I guess in Intellij you can find it in the "File/Settings" menu under "Editor/Code style/Java" -> "Import layout" section where you can edit the import order.

TheCookieLab:
Let’s not normalize raising concerns on blank lines in between imports,
comment style, etc in code reviews, that’s frankly anal retentive and
causes more harm than good. There’s a reason code style is typically
considered off-limits in code reviews for anything not explicitly defined
upfront as a universal style guide.

TheCookieLab:
No thanks.

ta4j-core/src/main/java/org/ta4j/core/criteria/VersusEnterAndHoldCriterion.java
```diff
@@ -23,12 +23,8 @@
  */
 package org.ta4j.core.criteria;
 
-import org.ta4j.core.AnalysisCriterion;
-import org.ta4j.core.BarSeries;
-import org.ta4j.core.BaseTradingRecord;
-import org.ta4j.core.Position;
+import org.ta4j.core.*;
```
nimo23, line 26: Please remove the "`*`" (asterisk) and import only required classes.

team172011:
> Let’s not normalize raising concerns on blank lines in between imports, comment style, etc in code reviews, that’s frankly anal retentive and causes more harm than good. There’s a reason code style is typically considered off-limits in code reviews for anything not explicitly defined upfront as a universal style guide.
> […](#)

@TheCookieLab I aggree, especially in PRs like this where the focus is on the general design and solving different challenges to reach the objectives.
But we should also ensure not to touch hundreds of unrelated files, so I am okay with having to correct those import issues

team172011:
Current problem I face ist that one goal of this PR is to eliminate alle the difficulties we have with the indexed based caching and the sync between the circular array in `CachedIndicators` and in the `BarSeries`. Having
the `NativeIndicatorValueCache`, `NativeRecursiveIndicatorValueCache` is good for comparison, but If we still have those indexed based caches nothing has been really fixed regarding this problem. So I do not want to have them in the final version of this PR.

This leads to the second challenge, how to replace the `RecursiveCachedIndicator` with a version that is not index based. The `RecursiveCachedIndicator` needs to now the latest available value to decided if it should iteratively fill the cache.

team172011:
> With such a solution there are no more indexes and no more iterating in the cache.

But isn't this exactly the same what I already did in this PR? I just used the endTime of each bar as primary key, it is mandatory and it makes no sense to have two bars for the same endTime -> primary key.

And how would the sync work in constant time? At the moment, the BarSeries and CachedIndicator are using the maximumBarCount and removedBarCount property to calculate the "real" index. If we would only store a unique key (like my PR already does with endTime) we do not know how many bars or which bars may have been removed from the BarSeries. We would have to iterate over the whole cache values and check if they are still in the barSeries

However the concept or the RecursiveCachedIndicator would not work without known and storing indices of values

team172011:
> The indicator calculates a value for a bar and then saves this value in its cache. The next time the bar asks the indicator for the value, the indicator looks in its cache to see if it has already been calculated.

This idea sounds very interesting. This would allow an indicator not just to cache the final value for index n but also use it for meta information like the `ParabolicSarIndicator` would need them for the next index n+1 (isUptrend, lastExtreme, acceleration facto, see #975).

That means we would change the Indicator interface to something like:

```java
public interface Indicator<T> {

    T getValue(Bar bar);
    
    // ...
}
```

And the Bar interface to something like

```java
public interface Bar extends Serializable {
    
    IndicatorValueCache getCache(IndicatorInstanceKey key)

    // ...
}
```

The API of the IndicatorValueCache could look like this, providing also a possibility to store meta information required for
further calculations

```java
public interface IndicatorValueCache<T> {
    
    Optional<T> getValue(); // returns the value for this indicator if already calculated and stored
    void putValue(); // stores the value
    Map<String, Object> getMetaValues(); // possibility to add meta information
   putMetaValues(Map<String, Object>); // possibility to get meta information
  // and/or: 
  Optional <Object> getMetaValue(String key);
  void putMetaValue(String key, Object value);

}
```

And of course each indicator instance needs to have its own key for the cache in the Bar

```java
public interface IndicatorInstanceKey {
     
   // needs to be stored in each indicator instance
   default String getKey() {
       return new UUID()
  }
}
```

team172011:
But I remember one issue I had in a previous try to refactor the caching: 

As far as you leave the `Indicator<T>` scope (by not extending it or putting the cache in another class like `Bar`) you are loosing the information about the generic class parameter _T_. And that means it gets very ugly because you can handle the results only as Objects and need to do a lot of casting and type checking.

So this suggestion would not work, because the `Bar` does not know which `Indicator` with wich kind of type `T` is using it.

Because of that, when playing around with [implementing some concepts in swift](https://github.com/team172011/ta4swift/tree/main)  I completely dropped the concept of having a generic class parameter for Indicators.

My personal attitude is also to avoid using generics or just use them in cases where you really are able to gentrify something in a usefull way.

But beside of my personal flavor I really thing that **having a generic class parameter `T` in `Indicator` is useless for this lib**. At a first look ist sounds great to _have the possibility to create indicators that return any type _T__. But when having a closer look, the ta4j library really only supports having `Boolean` and `Num` type for `T`:
For example almost all rules expect Indicator<Num> as parameters. Also almost all the criterions returning `Num` and working with `Num` results. Analysis classes like `CashFlow` and `Returns` are also implementing the `Indicator<Num>` interface. 

**As far as I can see, there is no reasonable way to use this library with `Indicator<T>` where `T` is something else than `Boolean` or `Num`.** 

So one option could also be to remove the generic class parameter T by introducing two interfaces (for the time being ignoring `DateTimeIndicator` and I think one or two indicates returning a long or so):
- `NumIndicator` with getValue -> Num
- `BooleanIndicator` with getValue -> Boolean

TheCookieLab:
> Current problem I face ist that one goal of this PR is to eliminate alle the difficulties we have with the indexed based caching and the sync between the circular array in `CachedIndicators` and in the `BarSeries`. Still having the `NativeIndicatorValueCache`, `NativeRecursiveIndicatorValueCache` is good for comparison, **but If we still have those indexed based caches nothing has been really fixed** regarding this problem. So I do not want to have them in the final version of this PR.
> 
> This leads to a second challenge: How to replace the `RecursiveCachedIndicator` with a version that is not index based. The `RecursiveCachedIndicator` needs to now the latest available index to decided if it should iteratively fill the cache. From a specific index to another index.
> 
> Only idea I have so far is to create a cache that always checks/calculates if every previous index is already in the cache -> like suggested in #973 (#975) this would be a "cache" that is always iteratively calculating all values from a (configurable?) threshold to the current index if the current index is not already in the cache

@team172011 Are we sure we still need the `RecursiveCachedIndicator`? I modified 'ZLEMAIndicator` to extend `CachedIndicator` and then ran `CachedIndicatorTest.recursiveCachedIndicatorOnMovingBarSeriesShouldNotCauseStackOverflow` which passed successfully. I'm not terribly familiar with the origin story here so may be missing some important context. Can you confirm?

team172011:
@TheCookieLab it depends on the stack size of your jvm.For me the org.ta4j.core.indicators.volume.OnBalanceVolumeIndicatorTest#stackOverflowError test fails, when replacing only extending the cached indicator.

team172011:
So far this approach does not really solve all mentioned problems, especially the question regarding recursive calculations is still unsolved. I will revert this to a draft PR

sgflt:
May I ask, what access patterns does solve caching?

From my research, I have found a few main scenarios.

**LiveTrading**
There we are interested in last state. We usually want to update state of indicators on external event (tick, candle, other).
In this case caching does a little job. Only multiple indicators with the same parameters may benefit from it.

`new SMA(1 ,2, close)` for entry strategy and `new SMA(1, 2, close)` for exit may reuse same values. But we can workaround it and reuse instance of SMA(1,2,close) indicator.

Also array may be replaced with only pointer to last bar.

**Backtesting**
In this case I usually run optimization algorihms that estimates the best parameters of strategy. It is done in serial manner, so actually indexed array is superior also as cache. I have done little research experiments with caffeine and as cache for calculated values it was musch slower, because hashCode and equals are of similar costf of coputation as cached value. Adding locks and synchronized blocks it was less performatn than indexed array.

Secondly each optimization run have different parameters so it can not benefit from cache calculated in previous run.


RecursiveIndicator in serial acces does not need to be recursive, it depends only on previous value that may be stored inside instance. See approach in https://github.com/ta4j/ta4j/pull/1140

I have dropped all "caching indicators" in https://github.com/ta4j/ta4j/pull/1148 and after completion I am going to investigate caching needs of each indicator separately to prove that global cache does not make real benefit. Only optimization of each indicator does matter (I think).

TheCookieLab:
After a few spikes on this I don't think we'll be able to get significant performance gains out of adopting a different caching solution. I may ultimately circle back around to this but closing the loop on this for the time being.

Issue #893, opened 2022-10-05
Improve BarSeries with numType


nimo23:
See https://github.com/ta4j/ta4j/pull/753:

This issue is here to discuss the following:

Currently, BarSeries needs Function<Number, Num> function. Instead of this function, we can replace it by providing **an example of a Num** (the property is called `numType` which is a common `Num`). Having the direct glue with Num instead of its Function (when constructing the BarSeries), provides the following benefits:

- we use `numType#function()` to get its function: no need to instantiate a new Function for every BarSeries-constructors or BarSeriesBuilders.
- we use static defined Num's (for example, `DecimalNum.ZERO`): no need to instantiate a new Num for every BarSeries-constructors or BarSeriesBuilders
- with `Function<Number, Num> function`, it's not possible to determine its (runtime) Num-Type. With `numType`, it is easy to determine its runtime type.
- using `numType` instead of `DecimalNum::valueOf,` we can provide both the function (by numType.function()) and also determine its type - vice versa, it's not possible.
- we can use series.zero(), series.one(), series().hundred() (etc.) which get its value by its static defined properties in the dedicated `Num.class` (this avoids the need to instantiate, for example, series.valueOf(1)).
- we can use those static series.zero() on any other place (for example, indicators). Actually, we need at least one number to derive its constants (0,1,100, etc). Having `numType` instead of numFunction as a first class property in any barSeries solves this issue and we can access those static nums without the need of an instantiated num. So we can call easily series.zero() instead of numberOfTen.zero() (which is a little irritating).
- it eliminates the excessive instance creation of num-function when putting it, for example, in BarSeries-constructor or Bar-constructor. We always use one static constant from the Num-class (for example, DoubleNum.ZERO) and put this into those constructors.
- w can easily replace all those valueOf(0), valueOf(1), valueOf(100) by series.zero(), series.one(), series.hundred() in the next step. Actually, we need at least one instantiated Num to access its constants. (A BarSeries can be empty, thus indicators cannot actually acccess the Num constants. Having `numType` solves this issue.).
- with `numType`, there is no more need to instantiate a Num and/or a Num-function on every construction (because both, numType and numfunction can be easily accessed by constants defined in the Num-classes, for example: within BaseBarConstructor, I can put Decimal.ZERO to provide numType and implicitly numFunction)

[...]

PR #909, opened 2022-10-30
BarSeries with NumType

nimo23:
Fixes https://github.com/ta4j/ta4j/issues/893.

- BarSeries uses any instance of Num to determine its Num function.
- Any indicator, criterion or any other class uses static Num's (where possible) instead of creating new instances of Num.


ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
```diff
@@ -35,7 +35,7 @@
  */
 public class DoubleNum implements Num {
 
-    private static final DoubleNum ZERO = DoubleNum.valueOf(0);
+    public static final DoubleNum ZERO = DoubleNum.valueOf(0);
```
team172011, line 38: why making (only) this one public?

team172011 (approved):

ta4j-core/src/main/java/org/ta4j/core/num/DoubleNum.java
nimo23: Because within any indicator/criteria/series-calculation, we should access these static values _​​indirectly_ via the methods of `BarSeries` and the `Num`  (`one()`, `hundred()`, etc.), as this ensures that these nums have the correct precision in all cases. 

For example:

```
@Override
public Num one() {
        return mathContext.getPrecision() == DEFAULT_PRECISION ? ONE : function().apply(1);
 }
```

If we allow the user to access `ONE` directly within any indicator/criteria/series-calculation, then the right precision could be ignored. Therefore, users must use the indirect methods when accessing the static numbers. The public `ZERO` should only be used for the `BarSeries`-constructor (it just makes it easier for the user to use a default Num within the BarSeries-constructor). Note that the user can always use another `ZERO` with another precision within the `BarSeries`-Constructor, for example: `MY_ZERO = DecimalNum.valueOf(0.0000000000, 10); `

nimo23: Another alternative is to use default BarSeries-constructors for `DecimalNum` and `DoubleNum`: 

`static DecimalNumBarSeries(..)` // uses default precision
`static DoubleNumBarSeries`(..).` // uses default precision

With this, we can switch Num.ZERO to `private`.

What do you think?

team172011: I am unsure if it makes sense at all to specify the precision to another value than the default value. Can you imagine a use case that would work e.g. where you could work with a higher or lower precision.

Example of calculation in `SMAIndicator`:
``` java
    @Override
        Num sum = zero();                                            // default precision from Num.ZERO
        for (int i = Math.max(0, index - barCount + 1); i <= index; i++) {
            sum = sum.plus(indicator.getValue(i));       // precision from Num.ZERO used
        }
        final int realBarCount = Math.min(barCount, index + 1);
        return sum.dividedBy(numOf(realBarCount));      // precision from Num.ZERO used as well
    }
    @Override
    public String toString() {
        return getClass().getSimpleName() + " barCount: " + barCount;
    }
```

nimo23: For `DecimalNum`, we have the DEFAULT_PRECISION of `32`. I think, if someone uses a lower precision (e.g. `16`), then the **memory consumption** for holding DecimalNums _decreases_ and the **speed** for operating with those DecimalNums _increases_. Or? 

For `DoubleNum` we have a default precision of `5` (by having an `EPS = 0.00001`). Actually, we have no possiblity to use another precision, but I can provide a PR (after this was merged) where the user can also set another precision for `DoubleNum`. With this, the user can increase the precision of `DoubleNum` to a maximum precision of `16`. `DoubleNum` is always faster and uses less memory than `DecimalNum`. For example, instead of having DecimalNum with  `precision=16`, the user can use `DoubleNum` with `precision=16` (eps=1/16).

**To sum up:** 
By having the possiblity to configure the precision of `DecimalNum` and `DoubleNum`, the user can change _speed_ and _memory consumption_ of the Nums, or? That would be a benefit. 


Btw, your example says:

`Num sum = zero();    / default precision from Num.ZERO`

The comment is not right, because it uses the `zero()` with the **inherited** precision (which could be the default precision or another precision):

```
@Override
public Num zero() {
return mathContext.getPrecision() == DEFAULT_PRECISION ? ZERO : function().apply(0);
 }
```

team172011: >  because it uses the zero() with the inherited precision 
okay right, now I see how once could configure the precision. 

static DecimalNumBarSeries(..) // uses default precision
static DoubleNumBarSeries(..).` // uses default precision

sounds interesting, but I think so far it is enough to have the BaseBarSeriesBuilder with `DecimalNum` as default and with the possibility to set it to `DoubleNum` if required.

> I can provide a PR (after this was merged) where the user can also set another precision for DoubleNum.

Regarding this I am unsure if there is a demand for this feature. In general I must admit that I have the feeling that the whole `Num` abstraction is not relevant for the vast majority of users.

nimo23: > I must admit that I have the feeling that the whole Num abstraction is not relevant for the vast majority of users.

Good point! I created https://github.com/ta4j/ta4j/issues/916 for further  discussion.

PR #918, opened 2022-11-28
add Indicator.unstablePeriod

nimo23:
Fixes #.

- add` Indicator#getUnstablePeriod()`

The main purpose of the PR is:

Actually, the user must _know_ which value to set for `Strategy#unstablePeriod`. However, this is not easy to _know_, because the user is not always aware of the technical details of the indicators (e.g. SARIndicator needs at least 200 bars before the values can be calculated correctly, other indicators needs at least its barCount, etc.). It's better to have this information (about the unstablePeriod of an indicator) implemented and documented **within each Indicator.** The implementor of the indicator must provide this information within the source code of the indicator. With this, the user can retrieve this information (by `Indicator#getUnstablePeriod()`) for various use cases.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendUpperBandIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendUpperBandIndicator.java
@@ -64,7 +64,7 @@ public class SuperTrendUpperBandIndicator extends RecursiveCachedIndicator<Num> 
         else
             return this.getValue(index - 1);
     }
-    
+
     @Override
     public int getUnstablePeriod() {
         return 0;


ta4j-core/src/main/java/org/ta4j/core/Indicator.java
```diff
@@ -42,6 +42,11 @@
      */
     T getValue(int index);
 
+    /**
+     * @return the number of bars up to which the indicator calculates wrong values
+     */
+    int getUnstablePeriod();
```
team172011, line 48: Why not adding a default implementation here? So we do not have to touch > 120 classes

nimo23: I didn't add a default method because I wanted to force the indicator developer **to explicitly specify** an **unstablePeriod** appropriate for this indicator. If we use a standard method, we (or the developer of the technical indicator) could avoid adding a suitable `unstablePeriod` to the indicator.

Btw, it would be good to review each indicator's unstable period so we can be sure that the unstablePeriod for this indicator is really right. Yes, it's a little work (120 classes), but I think we must do that, otherwise we cannot assure that each indicator has indeed it's right unstable period set.

Btw, I mostly used the `barCount` for setting the `unstablePeriod` for each indicator. However, I do not know if this was always right. So you see it's better to really check `getUnstablePeriod() `for each indicator..

ta4j-core/src/main/java/org/ta4j/core/indicators/EMAIndicator.java
```diff
@@ -43,4 +43,9 @@ public class EMAIndicator extends AbstractEMAIndicator {
     public EMAIndicator(Indicator<Num> indicator, int barCount) {
         super(indicator, barCount, (2.0 / (barCount + 1)));
     }
+
+    @Override
+    public int getUnstablePeriod() {
+        return 0;
```
gelhaimer: EMA requires more than 0 periods to be computed.

Depending on how the EMA is initialized, it might require at least the number of period.


Ex: 
For a 9 EMA, the seed might be initialized with the 9 SMA.
This means that, to compute the 9 EMA, you'll need at least 9 periods for the SMA 


This topic on reddit was very insightful !

https://www.reddit.com/r/algotrading/comments/bbnv8c/how_many_candles_minimum_to_get_an_accurate_ema/

nimo23: Thanks. 

@team172011 You see, each indicator must treat its own correct `unstablePeriod` thus it is crucial to implement this method as mandatory and not as optional.

I think, we really have to go through each indicator and set unstablePeriod explicitly.

team172011:
I am still unsure about this enhancement, especially for indicators that have an unstable period, it should be up to the user to decide when the data of the indicator is useful. For example a user with 190 bars would like to use the `SARIndicator` in his strategy anyway. 

Furthermore just adding the `getUnstablePeriod()` without any usages does not change anything nor? If introduced this concept must also be  implemented in the BarSeriesManager where the `unstablePeriod` will be checked for each index

team172011:
> Another thing: should we call it `getUnstablePeriod()` or `getUnstableBars()`? Which name is more appropriate?

I would prefer getUnstableBars()

hhashim1:
Any updates/ETA on this?

PR #930, opened 2023-02-06
SuperTrend Indicator implementation

### LINKED ISSUE #923: Supertrend Indicator

Is it possible to add the Supertrend indicator?
I found a beginning of the solution in https://stackoverflow.com/questions/68063996/java-implementation-of-supertrend

Thank you

preetishmadalia:
Fixes #923 

- Added code for SuperTrendIndicator
- Added unit tests.
- Updated CHANGELOG.md

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/LowerBandIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/LowerBandIndicator.java
@@ -1,0 +1,67 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017-2022 Ta4j Organization & respective
+ * authors (see AUTHORS)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package org.ta4j.core.indicators.supertrend;
+
+import org.ta4j.core.Bar;
+import org.ta4j.core.BarSeries;
+import org.ta4j.core.indicators.ATRIndicator;
+import org.ta4j.core.indicators.RecursiveCachedIndicator;
+import org.ta4j.core.indicators.helpers.MedianPriceIndicator;
+import org.ta4j.core.num.DoubleNum;
+import org.ta4j.core.num.Num;
+
+public class LowerBandIndicator extends RecursiveCachedIndicator<Num> {
+
+    private final ATRIndicator atrIndicator;
+    private final Integer multiplier;
+    private final MedianPriceIndicator medianPriceIndicator;
+
+    public LowerBandIndicator(final BarSeries barSeries) {
+        this(barSeries, new ATRIndicator(barSeries, 10), 3);
+    }
+
+    public LowerBandIndicator(final BarSeries barSeries, final ATRIndicator atrIndicator, final Integer multiplier) {
+        super(barSeries);
+        this.atrIndicator = atrIndicator;
+        this.multiplier = multiplier;
+        this.medianPriceIndicator = new MedianPriceIndicator(barSeries);
+    }
+
+    @Override
+    protected Num calculate(int index) {
+        if (index == 0)
+            return DoubleNum.valueOf(0);
+
+        Bar bar = getBarSeries().getBar(index - 1);
+
+        Num currentBasic = this.medianPriceIndicator.getValue(index)
+                .minus(DoubleNum.valueOf(this.multiplier).multipliedBy(this.atrIndicator.getValue(index)));
+
+        if (currentBasic.isGreaterThan(this.getValue(index - 1))
+                || bar.getClosePrice().isLessThan(this.getValue(index - 1))) {
+            return currentBasic;
+        } else
+            return this.getValue(index - 1);
+    }
+}
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
@@ -56,7 +56,7 @@ public class SuperTrendIndicator extends RecursiveCachedIndicator<Num> {
         Bar bar = getBarSeries().getBar(i);
 
         if (this.getValue(i - 1).isEqual(this.superTrendUpperBandIndicator.getValue(i - 1))
-                && bar.getClosePrice().isLessThanOrEqual(this.superTrendUpperBandIndicator.getValue(i))) {
+                && bar.getClosePrice().isLessThan(this.superTrendUpperBandIndicator.getValue(i))) {
             value = this.superTrendUpperBandIndicator.getValue(i);
         }
 
@@ -64,10 +64,10 @@ public class SuperTrendIndicator extends RecursiveCachedIndicator<Num> {
                 && bar.getClosePrice().isGreaterThan(this.superTrendUpperBandIndicator.getValue(i))) {
             value = this.superTrendLowerBandIndicator.getValue(i);
         } else if (this.getValue(i - 1).isEqual(this.superTrendLowerBandIndicator.getValue(i - 1))
-                && bar.getClosePrice().isGreaterThanOrEqual(this.superTrendLowerBandIndicator.getValue(i))) {
+                && bar.getClosePrice().isGreaterThan(this.superTrendLowerBandIndicator.getValue(i))) {
             value = this.superTrendLowerBandIndicator.getValue(i);
         } else if (this.getValue(i - 1).isEqual(this.superTrendLowerBandIndicator.getValue(i - 1))
-                && bar.getClosePrice().isLessThan(this.superTrendUpperBandIndicator.getValue(i))) {
+                && bar.getClosePrice().isLessThan(this.superTrendLowerBandIndicator.getValue(i))) {
             value = this.superTrendUpperBandIndicator.getValue(i);
         }
 
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
@@ -83,4 +83,12 @@ public class SuperTrendIndicator extends RecursiveCachedIndicator<Num> {
         }
         return "Buy";
     }
+
+    public LowerBandIndicator getLowerBandIndicator() {
+        return lowerBandIndicator;
+    }
+
+    public UpperBandIndicator getUpperBandIndicator() {
+        return upperBandIndicator;
+    }
 }
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
@@ -43,7 +43,7 @@ public class SuperTrendIndicator extends RecursiveCachedIndicator<Num> {
     }
 
     public SuperTrendIndicator(final BarSeries series) {
-        this(series, 3, 10);
+        this(series, 10, 3);
     }
 
     @Override
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
@@ -25,10 +25,8 @@ package org.ta4j.core.indicators.supertrend;
 
 import org.ta4j.core.Bar;
 import org.ta4j.core.BarSeries;
-import org.ta4j.core.Trade;
 import org.ta4j.core.indicators.ATRIndicator;
 import org.ta4j.core.indicators.RecursiveCachedIndicator;
-import org.ta4j.core.num.DoubleNum;
 import org.ta4j.core.num.Num;
 
 public class SuperTrendIndicator extends RecursiveCachedIndicator<Num> {
@@ -48,7 +46,7 @@ public class SuperTrendIndicator extends RecursiveCachedIndicator<Num> {
 
     @Override
     protected Num calculate(int i) {
-        Num value = DoubleNum.valueOf(0);
+        Num value = zero();
 
         if (i == 0)
             return value;
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
@@ -72,6 +72,7 @@ public class SuperTrendIndicator extends RecursiveCachedIndicator<Num> {
 
         return value;
     }
+
     public SuperTrendLowerBandIndicator getSuperTrendLowerBandIndicator() {
         return superTrendLowerBandIndicator;
     }
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
@@ -25,20 +25,21 @@ package org.ta4j.core.indicators.supertrend;
 
 import org.ta4j.core.Bar;
 import org.ta4j.core.BarSeries;
+import org.ta4j.core.Trade;
 import org.ta4j.core.indicators.ATRIndicator;
 import org.ta4j.core.indicators.RecursiveCachedIndicator;
 import org.ta4j.core.num.DoubleNum;
 import org.ta4j.core.num.Num;
 
 public class SuperTrendIndicator extends RecursiveCachedIndicator<Num> {
-    private final UpperBandIndicator upperBandIndicator;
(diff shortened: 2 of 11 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/LowerBandIndicator.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.ta4j.core.Bar;
+import org.ta4j.core.BarSeries;
+import org.ta4j.core.indicators.ATRIndicator;
+import org.ta4j.core.indicators.RecursiveCachedIndicator;
+import org.ta4j.core.indicators.helpers.MedianPriceIndicator;
+import org.ta4j.core.num.DoubleNum;
+import org.ta4j.core.num.Num;
+
+public class LowerBandIndicator extends RecursiveCachedIndicator<Num> {
```
TheCookieLab: Can you rename this to **SuperTrend**LowerBandIndicator? Several indicators have lower (and upper) bands so just LowerBandIndicator is ambiguous. Granted it can be identified via the package but that requires an extra step. Being immediately obvious is a boon to readability and goes a long to to reducing cognitive load.

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/UpperBandIndicator.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.ta4j.core.Bar;
+import org.ta4j.core.BarSeries;
+import org.ta4j.core.indicators.ATRIndicator;
+import org.ta4j.core.indicators.RecursiveCachedIndicator;
+import org.ta4j.core.indicators.helpers.MedianPriceIndicator;
+import org.ta4j.core.num.DoubleNum;
+import org.ta4j.core.num.Num;
+
+public class UpperBandIndicator extends RecursiveCachedIndicator<Num> {
```
TheCookieLab: Please rename to SuperTrendUpperBandIndicator

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Override
+    protected Num calculate(int i) {
+        Num value = DoubleNum.valueOf(0);
+
+        if (i == 0)
+            return value;
+        Bar bar = getBarSeries().getBar(i);
+        /*
+         * if(bar.getClosePrice().isLessThanOrEqual(this.upperBandIndicator.getValue(i))
```
TheCookieLab: Is this still needed? Please remove if not. Leaving commented code becomes a point of confusion especially as time moves on.

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
```diff
@@ -0,0 +1,94 @@
...
+                && bar.getClosePrice().isLessThan(this.upperBandIndicator.getValue(i))) {
+            value = this.upperBandIndicator.getValue(i);
+        }
+
+        return value;
+    }
+
+    public String getSignal(int index) {
+        if (getBarSeries().getBar(index).getClosePrice().isLessThanOrEqual(this.getValue(index))) {
+            return "Sell";
```
TheCookieLab: Would it make sense to use the Trade.TradeType enum here instead of the string literals?

TheCookieLab (changes requested):

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/UpperBandIndicator.java
preetishmadalia: renamed class name and references.

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/LowerBandIndicator.java
preetishmadalia: renamed class name and references.

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
preetishmadalia: removed commented code.

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
preetishmadalia: Used Trade.TradeType as per your suggestion. Kept method return type as String.

TheCookieLab (commented):
@preetishmadalia thanks for making those changes. I am a bit torn on the `getSignal` function, primarily for 2 reasons:

1. AFAIK it would be the first meaningful (i.e. non-getter) public function that exists outside of the Indicator interface. Clients who use this function would thus be coupled to _implementation_ rather than API, which is not a precedent we want to set without careful consideration.
2. The current implementation returns a BUY on index 0 which while an edge case, is a false signal and potentially a costly one if followed at face value in a live trading scenario.

TBH it may just be better to let clients compare the Super Trend calculated value to the close price themselves. @team172011 @nimo23  do you have any opinions either way?

gtskaushik:
@TheCookieLab @preetishmadalia How about this approach
+1 --> Buy signal
any other +ive integer --> the +ive integer "n" is n candles past the last Buy signal
-1 --> Sell signal
any other -ive integer --> the -ive integer "n" is n candles past the last Buy signal

TheCookieLab:
@gtskaushik to clarify do you mean we remove the `getSignal` function and modify the `calculate` to return those values as signals?

I question if this is the appropriate place for interpreting the signal at all as any non-trivial interpretation will be opinionated and may not match how clients want to use them. For example, why should a negative integer represent the number of candles past the last Buy signal instead of number of candles since the last Sell signal? 

I'm assuming you chose that because the # of candles since the last buy is the more meaningful signal within your app but we can't assume that's true for everyone. What if there's a use case where one uses the SuperTrend calculated value not as a simple BUY/SELL signal but as an input into a more complex calculation?

IMO a library like TA4J (especially at the Indicator level) should focus on providing the core building blocks and let clients interpret according to their own use cases. My objection to the `getSignal` function is that it has more cons than pros:

Pros:
1. It saves the client 1 line of code assuming they are using the SuperTrend indicator as a simple BUY/SELL signal.

Cons:
1. It's extra to the Indicator interface which is a concerning precedent to set as it can be a slippery slope to a codebase full of `instanceof` and casting instead of the solid abstractions we have now. I do however see there's precedent as the DeMarkPivotPointIndicator has a `getBarsOfPreviousPeriod` and there's probably others as well. Still it's worth explicitly discussing whether it's an acceptable practice or something we should nip in the bud.
3. The false positive signal at Index 0 requires a fix, but I suspect any fix would by nature be opinionated, leading to extra complexity and potential incompatibility with use cases.

preetishmadalia:
Thanks for the inputs. I can remove the 'getSignal()' from Indicator class.

TheCookieLab (approved):
LGTM

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
```diff
@@ -0,0 +1,83 @@
...
+        this.superTrendLowerBandIndicator = new SuperTrendLowerBandIndicator(series, atrIndicator, multiplier);
+    }
+
+    public SuperTrendIndicator(final BarSeries series) {
+        this(series, 3, 10);
+    }
+
+    @Override
+    protected Num calculate(int i) {
+        Num value = DoubleNum.valueOf(0);
```
team172011: you cannot use `DoubleNum` here because the indicator should be independent of the underlying `Num` implementation. Please use the `Num zero()` method and reuse this value within an instance variable

team172011 (changes requested):

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
```diff
@@ -0,0 +1,81 @@
...
+        this.superTrendLowerBandIndicator = new SuperTrendLowerBandIndicator(series, atrIndicator, multiplier);
+    }
+
+    public SuperTrendIndicator(final BarSeries series) {
+        this(series, 3, 10);
+    }
+
+    @Override
+    protected Num calculate(int i) {
+        Num value = zero();
```
team172011: This looks good now. One further improvement would be to store this constant as instance variable (2). For example like the AbstractEMAIndicator does:

https://github.com/ta4j/ta4j/blob/09cb8397161052403438c22c46404e2bc0c454ab/ta4j-core/src/main/java/org/ta4j/core/indicators/AbstractEMAIndicator.java#L42

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
```diff
@@ -0,0 +1,81 @@
...
+
+        if (i == 0)
+            return value;
```
team172011, line 54: Code style (1): please always add braces to `if` statements.

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendLowerBandIndicator.java
```diff
@@ -0,0 +1,68 @@
...
+            final Integer multiplier) {
+        super(barSeries);
+        this.atrIndicator = atrIndicator;
+        this.multiplier = multiplier;
+        this.medianPriceIndicator = new MedianPriceIndicator(barSeries);
+    }
+
+    @Override
+    protected Num calculate(int index) {
+        if (index == 0)
```
team172011, line 53: Code style (1) and please use (2) here as well

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendUpperBandIndicator.java
```diff
@@ -0,0 +1,68 @@
...
+        this.medianPriceIndicator = new MedianPriceIndicator(barSeries);
+    }
+
+    @Override
+    protected Num calculate(int index) {
+
+        Num currentBasic = this.medianPriceIndicator.getValue(index)
+                .plus(DoubleNum.valueOf(this.multiplier).multipliedBy(this.atrIndicator.getValue(index)));
+
+        if (index == 0)
```
team172011, line 56: Code style (1) and please use (2) here as well

team172011 (approved):

team172011 (changes requested):
@preetishmadalia If you do the recommended changes I will merge this PR

preetishmadalia:
Please give me couple of days. Is that ok?

gtskaushik:
@preetishmadalia I tried this code against 2 scripts that I'm working on. I found a wierd issue in one of those scripts.

1. Tried to load data from 2015 to today and send create BarSeries for this whole data and send it to SuperTrendIndicator. I did not get any Buy or Sell signal since 2017.
2. Then tried loading data for each year and send year wise data to SuperTrendIndicator. Then the results are as expected.

Any insights on this?

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
preetishmadalia: Just to clarify, the suggestion is to make `multiplier` as class variable in SuperTrendIndicator.java?

gtskaushik:
@preetishmadalia We can pair up offline if needed

preetishmadalia:
> * Tried to load data from 2015 to today and send create BarSeries for this whole data and send it to SuperTrendIndicator. I did not get any Buy or Sell signal since 2017.

@gtskaushik can you please tell me the ticker/data you are testing against? I will try to test it.

gtskaushik:
> > * Tried to load data from 2015 to today and send create BarSeries for this whole data and send it to SuperTrendIndicator. I did not get any Buy or Sell signal since 2017.
> 
> @gtskaushik can you please tell me the ticker/data you are testing against? I will try to test it.

I was testing NIFTY index in India

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
team172011: I mean to make zero an instance variable, so it has not to be calculated for every method call

nimo23: >One further improvement would be to store this constant as instance variable (2).

@team172011 `zero()` already accesses a **static** property (defined in the dedicated `Num.class`), so there is no benefit to put an instance variable again only to point to a static field (this is possible since https://github.com/ta4j/ta4j/pull/909).

gtskaushik:
> > > * Tried to load data from 2015 to today and send create BarSeries for this whole data and send it to SuperTrendIndicator. I did not get any Buy or Sell signal since 2017.
> > 
> > 
> > @gtskaushik can you please tell me the ticker/data you are testing against? I will try to test it.
> 
> I was testing NIFTY index in India

@preetishmadalia 
I have attached a csv file which has the following columns
- Timestamp
- Open
- High
- Low
- Close
- Supertrend
This data is from 2015 to today for Nifty index in India. If you notice, the last Sell signal was at 16/03/17 14:57. But if I load the data of only say 2020, I'm getting different results. You can use the OHLC data in the csv to check it out

[supertrend_nifty_india.csv](https://github.com/ta4j/ta4j/files/10767423/supertrend_nifty_india.csv)

preetishmadalia:
> > > > * Tried to load data from 2015 to today and send create BarSeries for this whole data and send it to SuperTrendIndicator. I did not get any Buy or Sell signal since 2017.
> > > 
> > > 
> > > @gtskaushik can you please tell me the ticker/data you are testing against? I will try to test it.
> > 
> > 
> > I was testing NIFTY index in India
> 
> @preetishmadalia I have attached a csv file which has the following columns
> 
> * Timestamp
> * Open
> * High
> * Low
> * Close
> * Supertrend
> 
> This data is from 2015 to today for Nifty index in India. If you notice, the last Sell signal was at 16/03/2017 14:57. But if I load the data of only say 2020, I'm getting different results. You can use the OHLC data in the csv to check it out
> 
> [supertrend_nifty_india.csv](https://github.com/ta4j/ta4j/files/10767423/supertrend_nifty_india.csv)

Thanks. I will look into this over weekend. What's the best way to reach you?

gtskaushik:
> > > > > * Tried to load data from 2015 to today and send create BarSeries for this whole data and send it to SuperTrendIndicator. I did not get any Buy or Sell signal since 2017.
> > > > 
> > > > 
> > > > @gtskaushik can you please tell me the ticker/data you are testing against? I will try to test it.
> > > 
> > > 
> > > I was testing NIFTY index in India
> > 
> > 
> > @preetishmadalia I have attached a csv file which has the following columns
> > 
> > * Timestamp
> > * Open
> > * High
> > * Low
> > * Close
> > * Supertrend
> > 
> > This data is from 2015 to today for Nifty index in India. If you notice, the last Sell signal was at 16/03/2017 14:57. But if I load the data of only say 2020, I'm getting different results. You can use the OHLC data in the csv to check it out
> > [supertrend_nifty_india.csv](https://github.com/ta4j/ta4j/files/10767423/supertrend_nifty_india.csv)
> 
> Thanks. I will look into this over weekend. What's the best way to reach you?

Email: kaushik.gunasekaran@dotconn.net
Whatsapp: +91 9538140126

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
team172011: this is only true for the default implementations `DoubleNum` and `DecimalNum` but I agree that using zero() is ok because I think the probability of using another implementation is also zero()..

team172011:
I will wait with merging this PR unit the issue from @gtskaushik is resolved @preetishmadalia

preetishmadalia:
There was a typo in the default constructor. Waiting for @gtskaushik to test and confirm.

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
```diff
@@ -0,0 +1,84 @@
...
+        if (this.getValue(i - 1).isEqual(this.superTrendUpperBandIndicator.getValue(i - 1))
+                && bar.getClosePrice().isGreaterThan(this.superTrendUpperBandIndicator.getValue(i))) {
+            value = this.superTrendLowerBandIndicator.getValue(i);
+        } else if (this.getValue(i - 1).isEqual(this.superTrendLowerBandIndicator.getValue(i - 1))
+                && bar.getClosePrice().isGreaterThanOrEqual(this.superTrendLowerBandIndicator.getValue(i))) {
+            value = this.superTrendLowerBandIndicator.getValue(i);
+        } else if (this.getValue(i - 1).isEqual(this.superTrendLowerBandIndicator.getValue(i - 1))
+                && bar.getClosePrice().isLessThan(this.superTrendUpperBandIndicator.getValue(i))) {
+            value = this.superTrendUpperBandIndicator.getValue(i);
+        }
```
gtskaushik: There are 3 issues.

1. Line 59 should use isLessThan instead of isLessThanOrEqual
2. Line 67 should use isGreaterThan instead of isGreaterThanOrEqual 
3. Line 70 should use superTrendLowerBandIndicator instead of superTrendUpperBandIndicator

This is the correct conditionals
```
if (this.getValue(i - 1).isEqual(this.superTrendUpperBandIndicator.getValue(i - 1))
        && bar.getClosePrice().isLessThan(this.superTrendUpperBandIndicator.getValue(i))) {
    value = this.superTrendUpperBandIndicator.getValue(i);
} else if (this.getValue(i - 1).isEqual(this.superTrendUpperBandIndicator.getValue(i - 1))
        && bar.getClosePrice().isGreaterThan(this.superTrendUpperBandIndicator.getValue(i))) {
    value = this.superTrendLowerBandIndicator.getValue(i);
} else if (this.getValue(i - 1).isEqual(this.superTrendLowerBandIndicator.getValue(i - 1))
        && bar.getClosePrice().isGreaterThan(this.superTrendLowerBandIndicator.getValue(i))) {
    value = this.superTrendLowerBandIndicator.getValue(i);
} else if (this.getValue(i - 1).isEqual(this.superTrendLowerBandIndicator.getValue(i - 1))
        && bar.getClosePrice().isLessThan(this.superTrendLowerBandIndicator.getValue(i))) {
    value = this.superTrendUpperBandIndicator.getValue(i);
}
```

Verified with this reference - https://medium.com/codex/step-by-step-implementation-of-the-supertrend-indicator-in-python-656aa678c111

gtskaushik (changes requested):

ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java
preetishmadalia: I can change the first 2 conditions. Can you check the third condition again? To me it seems correct.

gtskaushik: Screenshot 2023-02-24 at 9 44 40 AM
Please check the lines circled in red in the attached screenshot. 
Also when we do not change third point mentioned above, I'm getting the same issue as earlier

gtskaushik (approved):
+1

Issue #957, opened 2023-04-13
[BUG] StandardDeviationCriterion betterThan function considers larger values better than smaller values


TheCookieLab:
**I have checked existing issues and wiki**
- [X] I could not find similar [issues](https://github.com/ta4j/ta4j/issues?utf8=%E2%9C%93&q=)
- [X] I could not find a solution in the [wiki](https://ta4j.github.io/ta4j-wiki/) or [faq section](https://ta4j.github.io/ta4j-wiki/FAQ.html)

**Describe the bug**
**StandardDeviationCriterion**'s betterThan logic considers higher values better than lower values. While this is true for the majority of criteria, those related to standard deviation should consider lower values better as they imply a more consistent data set. The same issue also applies to:

- **RelativeStandardDeviationCriterion**
- **VarianceCriterion**

TBH this is may be more of a semantic issue than a bug. What tipped me over is that **StandardErrorCriterion** also treats lower values as better and I figure these similar-in-kind criteria should behave consistently. I'm open to arguments challenging these premises (ex. not similar in kind) and that this fix is not needed.

TheCookieLab:
Yeah, subjectivity is kinda unavoidable when we are talking in terms of "better" and "worse". Before going further let me take a step back and level-set. 

The fact that `betterThan` is an interface function suggests to me that it plays a prominent role but I actually have not yet used it myself so I'm probably missing important context. Is the intention to relieve clients of the burden of interpreting the values they are calculating, or is there a specific use case where the various criteria are managed as generic interfaces?

PR #960, opened 2023-04-14
Criteria: add lessBetter to dedicated criteria

nimo23:
Fixes https://github.com/ta4j/ta4j/issues/957.

- added "lessBetter"-property for **AverageCriterion**
- added "lessBetter"-property for **RelativeStandardDeviation**
- added "lessBetter"-property for **StandardDeviationCriterion**
- added "lessBetter"-property for **StandardErrorCriterion**
- added "lessBetter"-property for **VarianceCriterion**
- added "lessBetter"-property for **NumberOfPositionsCriterion**

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/criteria/NumberOfPositionsCriterion.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/criteria/NumberOfPositionsCriterion.java
@@ -38,22 +38,22 @@ public class NumberOfPositionsCriterion extends AbstractAnalysisCriterion {
      * the criterion value the better. This property is only used for
      * {@link #betterThan(Num, Num)}.
      */
-    private final boolean lessBetter;
+    private final boolean lessIsBetter;
 
     /**
-     * Constructor with {@link #lessBetter} == true.
+     * Constructor with {@link #lessIsBetter} == true.
      */
     public NumberOfPositionsCriterion() {
-        this.lessBetter = true;
+        this.lessIsBetter = true;
     }
 
     /**
      * Constructor.
      * 
-     * @param lessBetter the {@link #lessBetter}
+     * @param lessIsBetter the {@link #lessIsBetter}
      */
-    public NumberOfPositionsCriterion(boolean lessBetter) {
-        this.lessBetter = lessBetter;
+    public NumberOfPositionsCriterion(boolean lessIsBetter) {
+        this.lessIsBetter = lessIsBetter;
     }
 
     @Override
@@ -67,12 +67,12 @@ public class NumberOfPositionsCriterion extends AbstractAnalysisCriterion {
     }
 
     /**
-     * If {@link #lessBetter} == false, then the lower the criterion value, the
+     * If {@link #lessIsBetter} == false, then the lower the criterion value, the
      * better, otherwise the higher the criterion value the better.
      */
     @Override
     public boolean betterThan(Num criterionValue1, Num criterionValue2) {
-        return lessBetter ? criterionValue1.isLessThan(criterionValue2)
+        return lessIsBetter ? criterionValue1.isLessThan(criterionValue2)
                 : criterionValue1.isGreaterThan(criterionValue2);
     }
 }
--- a/ta4j-core/src/main/java/org/ta4j/core/criteria/NumberOfPositionsCriterion.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/criteria/NumberOfPositionsCriterion.java
@@ -33,6 +33,29 @@ import org.ta4j.core.num.Num;
  */
 public class NumberOfPositionsCriterion extends AbstractAnalysisCriterion {
 
+    /**
+     * If true, then the lower the criterion value the better, otherwise the higher
+     * the criterion value the better. This property is only used for
+     * {@link #betterThan(Num, Num)}.
+     */
+    private final boolean lessBetter;
+
+    /**
+     * Constructor with {@link #lessBetter} == true.
+     */
+    public NumberOfPositionsCriterion() {
+        this.lessBetter = true;
+    }
+
+    /**
+     * Constructor.
+     * 
+     * @param lessBetter the {@link #lessBetter}
+     */
+    public NumberOfPositionsCriterion(boolean lessBetter) {
+        this.lessBetter = lessBetter;
+    }
+
     @Override
     public Num calculate(BarSeries series, Position position) {
         return series.one();
@@ -43,9 +66,13 @@ public class NumberOfPositionsCriterion extends AbstractAnalysisCriterion {
         return series.numOf(tradingRecord.getPositionCount());
     }
 
-    /** The lower the criterion value, the better. */
+    /**
+     * If {@link #lessBetter} == false, then the lower the criterion value, the
+     * better, otherwise the higher the criterion value the better.
+     */
     @Override
     public boolean betterThan(Num criterionValue1, Num criterionValue2) {
-        return criterionValue1.isLessThan(criterionValue2);
+        return lessBetter ? criterionValue1.isLessThan(criterionValue2)
+                : criterionValue1.isGreaterThan(criterionValue2);
     }
 }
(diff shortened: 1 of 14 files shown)


ta4j-core/src/main/java/org/ta4j/core/criteria/NumberOfPositionsCriterion.java
```diff
@@ -43,9 +66,13 @@ public Num calculate(BarSeries series, TradingRecord tradingRecord) {
...
-    /** The lower the criterion value, the better. */
+    /**
+     * If {@link #lessBetter} == false, then the lower the criterion value, the
+     * better, otherwise the higher the criterion value the better.
+     */
     @Override
     public boolean betterThan(Num criterionValue1, Num criterionValue2) {
-        return criterionValue1.isLessThan(criterionValue2);
+        return lessBetter ? criterionValue1.isLessThan(criterionValue2)
+                : criterionValue1.isGreaterThan(criterionValue2);
```
team172011, line 76: What about using an `int` for the `lessBetter` variable (for instance 0 or -1) and comparing the criterion values with `criterionValue1.compareTo(criterionValue2) < lessBetter`

nimo23: What is the benefit in using `Num#compareTo` directly instead of `Num#isLessThan` or `Num#isGreaterThan ` (which also calls `Num#compareTo`).  I cannot imagine that there is any performance gains.

Btw, using a `boolean` property in constructor is more intuitive than using -1,+1..

nimo23: @team172011 Using `int lessBetter` results in this code:

```
/**
 * Average criterion.
 * 
 * <p>
 * Calculates the average of a Criterion by dividing it by the number of
 * positions.
 */
public class AverageCriterion extends AbstractAnalysisCriterion {

    /**
     * If {@code 0}, then the lower the criterion value the better. If {@code -1}, 
     * then the higher the criterion value the better. This property is only used for
     * {@link #betterThan(Num, Num)}.
     */
    private final int lessBetter;

    private final AnalysisCriterion criterion;
    private final NumberOfPositionsCriterion numberOfPositionsCriterion = new NumberOfPositionsCriterion();

    /**
     * Constructor with {@link #lessBetter} == false.
     * 
     * @param criterion the criterion from which the "average" is calculated
     */
    public AverageCriterion(AnalysisCriterion criterion) {
        this.criterion = criterion;
        this.lessBetter = -1;
    }

    /**
     * Constructor.
     * 
     * @param criterion  the criterion from which the "average" is calculated
     * @param lessBetter if true, then sets {@link #lessBetter} to {@code 0}, otherwise to {@code -1}
     */
    public AverageCriterion(AnalysisCriterion criterion, boolean lessBetter) {
        this.criterion = criterion;
        this.lessBetter = lessBetter ? 0 : -1;
    }

    @Override
    public Num calculate(BarSeries series, Position position) {
        Num numberOfPositions = numberOfPositionsCriterion.calculate(series, position);
        return criterion.calculate(series, position).dividedBy(numberOfPositions);
    }

    @Override
    public Num calculate(BarSeries series, TradingRecord tradingRecord) {
        if (tradingRecord.getPositions().isEmpty()) {
            return series.zero();
        }
        Num numberOfPositions = numberOfPositionsCriterion.calculate(series, tradingRecord);
        return criterion.calculate(series, tradingRecord).dividedBy(numberOfPositions);
    }

    /**
     * If {@link #lessBetter} == false, then the lower the criterion value, the
     * better, otherwise the higher the criterion value the better.
     */
    @Override
    public boolean betterThan(Num criterionValue1, Num criterionValue2) {
        return criterionValue1.compareTo(criterionValue2) > lessBetter;
    }

}
```


Is using `int lessBetter` instead of `boolean lessBetter` really better? 
I know I would bypass a method  call(`Num#isLessThan`/`Num#isBetterThan`) and bypass an additional check for `NaN` if I use `int lessBetter`. But is it worth it in terms of measurable performance?

team172011: I do not know if it is better or not. I like to reduce the amount of if-else statements or ternary operators if possible. I think it is just a matter of taste and I am also ok with this approach

TheCookieLab:
I feels like we are trying to band-aid over a design contradiction. An interface function should be universally applicable but given we now have a significant sub-set of criteria where the concept of "better than" is either ambiguous, subjective, or non-applicable it raises the question of whether we should continue down this path of increasing complexity or would it be cleaner to remove the betterThan function altogether and let the client compare the num values directly?

ta4j-core/src/main/java/org/ta4j/core/criteria/NumberOfPositionsCriterion.java
nimo23: I do not like the idea to give to user an `int` as a choice because an `int` can have other values than `0` and `-1`. Thus, it is better to use a `boolean`. 

@team172011 You can merge this PR. I am finished.

TheCookieLab:
I agree that "better than" should be applicable to any given criteria, but
this should be the user's responsibility, not the toolset's. The point is
that we, at the framework level, should focus on providing tools that
generate objective data while leaving interpretation to the users.
Deviating from this principle harms everyone: users, by artificially
restricting their use cases (e.g., NumberOfBarsCriterion or
LinearTransactionCostCriterion - what if the client is a broker simulating
the other side of trades?), and maintainers, by unnecessarily increasing
code complexity. Stateless functions are one thing, but adding state would
be a significant step down that slippery slope.

Philosophical concerns aside, from a practical standpoint I'm still unclear
what value this functionality provides me, as a user, over the baseline of
comparing criteria values directly. Previously one could argue that
'betterThan' abstracts the user from certain semantic knowledge, but this
proposed change would negate that altogether. If I have the necessary
semantic knowledge at runtime to set the lessBetter flag, then I have
enough knowledge to compare and interpret the values directly, which is
simpler to boot. What am I missing here?

TheCookieLab (changes requested):
The changes in this fix are net-negative for the code-base's readability, consistency, and would further entrench a source of subjectivity and recurring future contention.

I would like to see a different approach, if not complete removal of the `betterThan` interface function then *something* else.

TheCookieLab:
It's certainly possible I simply don't understand betterThan's value, so
that's why I've inquired several times for use cases that would help me
understand it. I've laid out my arguments against it in what I believe to
be clear cost/benefit terms but I've yet to see a rebuttal from you that
goes further than "because I'm using it.".

TheCookieLab:
Please refrain from ...reviewing PRs? Because that's what I'm doing, reviewing your PR.

Do you have any use cases demonstrating the value of `betterThan` so as to enlighten those of us who don't understand? This is the third time asking.

TheCookieLab:
Yes, if the reviewer doesn't agree with the proposed change-set then he/she is not going to approve. How do code reviews work at your job?

I've provided reasoning for why I do not agree with these changes. I've also provided rationale why I believe removing `betterThan` altogether is the best, holistic solution. If you disagree, you then provide counter-points - flaws in my premises, reasoning, etc. That's how this is supposed to work. 

The fact that you've not provided any rationale on why I'm wrong or why `betterThan` should exit leads me to believe you don't have any, and your defensiveness is due to a mix of sunk-cost fallacy and taking critique of code you wrote as a personal attack. 

As the party who first opened this issue as a bug I'd actually prefer no action over this change-set.

TheCookieLab:
It is unneeded because it provides minimal benefit over simply comparing
the criteria values directly. Code has carrying cost so all things being
equal less code is better than more code.

This fix further strains the value proposition by negating any and all of
the existing benefit while simultaneously adding additional state to
manage.

TheCookieLab:
If it’s implementation is not known at runtime, then now can we correctly
set the ‘lessBetter’ param?

TheCookieLab:
Last call for community feedback. I'm planning on merging this PR within the next day or so.

PR #961, opened 2023-04-15
adds spread extension in bar class

marsteff:
- Adds a new member to the bar class. The spread is an important value that must be considered when opening new trades. By adding the spread to the current bar and using the new SpreadIndicator it can be used in Rules.
Not so useful for backtesting, but extremely important for live trading.

- [x] added an entry with related ticket number(s) to the unreleased section of `CHANGES.md`  (no related ticket)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 
 ### Changed
 - **BarSeriesManager** consider finishIndex when running backtest
+- **Extends Bar with spread**
 
 ### Removed/Deprecated
 
--- a/ta4j-core/src/main/java/org/ta4j/core/Bar.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/Bar.java
@@ -70,6 +70,11 @@ public interface Bar extends Serializable {
     long getTrades();
 
     /**
+     * @return the spread size
+     */
+    Num getSpread();
+
+    /**
      * @return the whole traded amount of the period
      */
     Num getAmount();
@@ -171,4 +176,6 @@ public interface Bar extends Serializable {
     }
 
     void addPrice(Num price);
+
+    void addPrice(Num price, Num spread);
 }
--- a/ta4j-core/src/main/java/org/ta4j/core/Bar.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/Bar.java
@@ -175,6 +175,10 @@ public interface Bar extends Serializable {
         addPrice(numFunction.apply(price));
     }
 
+    default void addPrice(Number price, Number spread, Function<Number, Num> numFunction) {
+        addPrice(numFunction.apply(price), numFunction.apply(spread));
+    }
+
     void addPrice(Num price);
 
     void addPrice(Num price, Num spread);
--- a/ta4j-core/src/main/java/org/ta4j/core/BarSeries.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BarSeries.java
@@ -371,7 +371,7 @@ public interface BarSeries extends Serializable {
     }
 
     default void addPrice(Number price, Number spread) {
-        addPrice(numOf(price), numOf(price));
+        addPrice(numOf(price), numOf(spread));
     }
 
     /**
--- a/ta4j-core/src/main/java/org/ta4j/core/BarSeries.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BarSeries.java
@@ -360,12 +360,18 @@ public interface BarSeries extends Serializable {
      */
     void addPrice(Num price);
 
+    void addPrice(Num price, Num spread);
+
     default void addPrice(String price) {
         addPrice(new BigDecimal(price));
     }
 
     default void addPrice(Number price) {
         addPrice(numOf(price));
+    }
+
+    default void addPrice(Number price, Number spread) {
+        addPrice(numOf(price), numOf(price));
     }
 
     /**
--- a/ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
@@ -303,7 +303,7 @@ public class BaseBar implements Bar {
      * @param spread     the actual spread size
      */
     public BaseBar(Duration timePeriod, ZonedDateTime endTime, Num openPrice, Num highPrice, Num lowPrice,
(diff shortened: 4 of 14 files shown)


marsteff:
yeah sure, will write some tests!

marsteff:
UnitTests created!

TheCookieLab:
I like the idea as this is useful information to have, but I'm wondering if it wouldn't be better to add properties for 'bid` and 'ask' rather than spread. In the crypto space several exchanges return bid and ask on their OHLC data so this would match up better. Spread is also derivable from the bid and ask, whereas it's not possible to go in the other direction. Thoughts?

marsteff:
Yeah, my intention was to change not that much.
But sure, the most broker bring up bid and ask values.
But when we doing this, we need it for all OHLC values I think and that would be a big change.
We also need then to calculate the middle value for the indicators. Don't know if this will be bad for the performance.

Maybe we can make a second implementation of Bar with OHLC Bid and Ask values..

marsteff:
When trading FX instruments (like I did), the spread changes very often (in time of high volatility). And to use it in a indicator, i think it is necessary to have it in the bar class (correct me if I'm wrong). And if we don't store bid and ask values (which are coming from most brokers) it is the only possibility to get this value.
And with the spread you can (if you like) calculate the bid and ask..

And like you say, it has to be in the addPrice method, but I think it is not that bad when you can create a bar directly with a initial spread.

And when you don't like to use a spread in your bars, feel free. I don't changed any existing constructor.

TheCookieLab:
How do you calculate the bid and ask from the spread? One would need at least another data point no?

I think @nimo23 makes a good point in that spread doesn't really belong with OHLCV data. A (finalized) bar only has one open price, one close price, the volume is defined over the bar's duration, etc etc. The Spread value on the other hand is a snapshot in time so it's questionable if we can define a single, definitive spread value for a given bar. Would it be convenient to have the bid/ask/spread data within the same scope as OHLCV? Certainly, however it doesn't really fit from an intuitive design standpoint.

marsteff:
> How do you calculate the bid and ask from the spread? One would need at least another data point no?
If you only have the Bid value, you can calculate the Ask value (bid + spread = ask)
If you have only the closePrice, you can calculate the ask and bid (close + spread/2 = ask , close - spread/2 = bid).

I agree, that the spread is a snapshot, but when you want to trade on candle end, then the spread at the end of the candle is the final value, because you will open a position at this moment.

It sould also be posible to using the spread to calculate the costs of a trade which is strongly combined to the actual real costs. With this approach, you don´t have a fixed costs modell.

TheCookieLab:
> > How do you calculate the bid and ask from the spread? One would need at least another data point no?
> 
> If you only have the Bid value, you can calculate the Ask value (bid + spread = ask) If you have only the closePrice, you can calculate the ask and bid (close + spread/2 = ask , close - spread/2 = bid).
> 

Indeed, from your own example we see it's not possible to know the Bid and Ask if you only have the Spread. The calculation using closePrice assumes the close price is equidistant from the best bid and ask, which is by no means guaranteed. Example:

Capture

It's a moot point now but assuming for a second we wanted to add some top level order book info to the Bar, it's preferable to add properties for both Bid and Ask rather than a single Spread property. One can derive the Spread from the Bid and Ask not to mention bid and ask are useful data points in their own right. 

> I agree, that the spread is a snapshot, but when you want to trade on candle end, then the spread at the end of the candle is the final value, because you will open a position at this moment.

Given bid/ask/spread data is fundamentally different in nature compared to the rest of the properties in `Bar` and its desired interpretation can vary (ex. you want it to represent the "closing" snapshot, but someone else may want the opening) this code is better suited to live in your client app (i.e. extend Bar) rather than the base Ta4j library. Ultimately we need a new `OrderBook` model interface.

marsteff:
The Broker IG is streaming only the Bid/Ask Values of an instument (like EURUSD). [Look here at the bottom](https://labs.ig.com/streaming-api-reference).
So at the moment i have to calculate the middle price of bid and ask price.

I agree that the spread is different to the other values, but the bid/ask values are the same as when you have only on closePrice.
When you getting tick-data over APIs, you getting mostly bid/ask values 

Here the description from [Histdata](http://www.histdata.com/f-a-q/) (where you can get historical data for forex pairs):
> What are the order of the fields in the exported data?
In each file you’ll get from us, here are the fields order of all bar based data:

> DateTime,Open,High,Low,Close,Volume

> For tick data files, here are the fields order of the supplied data:

> DateTime,Bid,Ask,Volume

I think a OrderBook is only interesting for crypto-trading but not for "standard" broker trading, because there is no OrderBook.

Can we find a way to integrate Ask/Bid values in Bar? With this approach i could calculate the spread in a SpreadIndicator by myself.

TheCookieLab:
> The Broker IG is streaming only the Bid/Ask Values of an instument (like EURUSD). [Look here at the bottom](https://labs.ig.com/streaming-api-reference). So at the moment i have to calculate the middle price of bid and ask price.
> 
> I agree that the spread is different to the other values, but the bid/ask values are the same as when you have only on closePrice. When you getting tick-data over APIs, you getting mostly bid/ask values
> 
> Here the description from [Histdata](http://www.histdata.com/f-a-q/) (where you can get historical data for forex pairs):
> 
> > What are the order of the fields in the exported data?
> > In each file you’ll get from us, here are the fields order of all bar based data:
> 
> > DateTime,Open,High,Low,Close,Volume
> 
> > For tick data files, here are the fields order of the supplied data:
> 
> > DateTime,Bid,Ask,Volume
> 
> I think a OrderBook is only interesting for crypto-trading but not for "standard" broker trading, because there is no OrderBook.
> 
> Can we find a way to integrate Ask/Bid values in Bar? With this approach i could calculate the spread in a SpreadIndicator by myself.

What do you mean by "standard broker trading" - equities and such? There are definitely order books in the world of stock trading. If you've ever placed a limit order then you've interacted with an order book. Granted historically many retail-facing brokers didn't bother providing access to Level 2 data (and forget about L3) but many increasingly are. Search around your broker's website/interface for Level 2 data, you may just be surprised.

TheCookieLab:
I'm going to close this PR barring any further developments. While I agree having the bid and ask info is valuable, I don't think this is the right approach for a base library. @marsteff I agree there's convenience in having the bid and ask data alongside the bar data, however it's not quite right from a data model standpoint and thus creates fundamental ambiguities and representation irregularities. Your particular use case may not care however as a base library we need to keep a broader perspective. If we implement it like this it's going to be a foundation of sand for future devs and inevitably will require a large refactor. The right way to do this is to introduce OrderBook to the data model with development TBD.

Issue #962, opened 2023-04-16
Case for Java generics, Bar Money, Volume, Duration, Time abstractions and Immutable bars


sergproua:
I'd like to list few of the challenges I'm facing with using TA4J and propose possible solutions

Challenge 1:
Duration does not work well for months, January duration is different then February and so on
Duration and Period classes do not extend the same interface
Solution:
All temporal types such as Instant, ZonedDateTime, LocalDateTime, etc have methods
 temporal.minus(long, TemporalUnit)
 temporal.plus(long, TemporalUnit)

```
interface BarDuration {
    ChronoUnit getUnit();
    int getMultiplier();
}
```

Challenge 2:
I don't want to use heavy ZonedDateTime since most of the market data providers use Epoch time or simply Instant
Solution:
Use Generics:
TimeType extends Temporal & Comparable<? super TimeType>>

Challenge 3:
I want to be able to use different types for Price and Volume depending on the instrument
Equity markets, volumes always integer, Crypto Markets varies depending on the instrument
Option strike prices could be integers and I don't want to be stuck with decimals

Solution:
Use Generics in Bars and BarSeries
Bar<PriceType extends Number & Comparable<PriceType>,
    VolumeType extends Number & Comparable<VolumeType>,
    DurationType extends BarDuration,
    TimeType extends Temporal & Comparable<? super TimeType>>
Potentially eliminates Num interface since you're able to perform compare and arithmetic operators using chosen Number implementation.

Thought on mutating Bars:
I believe that Bars should be immutable records and addTrade/addPrice methods should be removed. Instead, I should be able to create 
empty BarSeries and use barSeries.addPrice and/or barSeries.addTrade with a timestamp which will manage internal bar creation
and be the factory for it's bars.
BarSeries could well be Iterable and implement other collections interfaces

Here's an partial example of what it might look like and what existing BaseBar and BaseBarSeries would look like using generics and not require major refactoring.

```
import java.io.Serializable;
import java.math.BigDecimal;
import java.time.ZonedDateTime;
import java.time.temporal.ChronoUnit;
import java.time.temporal.Temporal;
import java.util.List;

interface BarDuration {
    ChronoUnit getUnit();
    int getMultiplier();
}

interface Bar<PriceType extends Number & Comparable<PriceType>,
        VolumeType extends Number & Comparable<VolumeType>,
        DurationType extends BarDuration,
        TimeType extends Temporal & Comparable<? super TimeType>>
        extends Serializable {

[...]

PR #967, opened 2023-04-18
Java Generics for Bar BarSeries

sergproua:
- Introduction of Java generics for Bar and BarSeries
- `BarSeries` changed to `BarSeries<T extends Bar>`
- `BaseBarSeries` implementation moved to `BaseBarSeriesAbstract<T extends BaseBar> implements BarSeries<T>`
- `BaseBarSeries` refactored to `BaseBarSeries<BaseBar> extends BaseBarSeriesAbstract<T extends BaseBar>` for those wishing to use `BaseBarSeries` with own implementation of `BaseBar`

- All test are passing

- This update introduces a warning for all indicators `Raw use of parameterized class BarSeries`

- [ ] added an entry with related ticket number(s) to the unreleased section of `CHANGES.md` #962

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/BarSeriesAbstract.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BarSeriesAbstract.java
@@ -41,7 +41,7 @@ import static org.ta4j.core.num.NaN.NaN;
  * Abstract base implementation of a {@link BarSeries}.
  * </p>
  */
-public abstract class BaseBarSeriesAbstract<T extends BaseBar> implements BarSeries<T> {
+public abstract class BarSeriesAbstract<T extends Bar> implements BarSeries<T> {
 
     /**
      * The logger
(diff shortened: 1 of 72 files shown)


ta4j-core/src/main/java/org/ta4j/core/BarSeriesAbstract.java
TheCookieLab, line 1: For consistency with existing convention `Abstract` should be a prefix rather than a suffix (ex. `BaseBarSeries`). That said it may be confusing to have both an 'AbstractBarSeries` and a `BaseBarSeries` as `Abstract` and `Base` are often used interchangeably.

~~To preempt any such confusion an alternative could be this new class takes the name `BaseBarSeries` and we rename the current `BaseBarSeries` to something more "concrete". @team172011 @nimo23 do you have any opinions here?~~

On second thought that would break consistency with all the other Base* interfaces. Let's just go with changing this to `AbstractBarSeries`.

TheCookieLab (changes requested):

team172011:
What is the advantage of this breaking change? You can use your own implementation by implementing the `Bar` interface.

> This update introduces a warning for all indicators Raw use of parameterized class BarSeries

In my opinion adding those warning is not acceptable

PR #977, opened 2023-05-26
Update ReturnCriterion to exclude starting capital of 100% (i.e. 1.0)

### LINKED ISSUE #976: [BUG] ReturnCriterion returns higher than expected by 1.0
**Describe the bug**
ReturnCriterion returns a value 1.0 higher than it should.

**Expected behavior**
Currently, if a backtest results in a 10% profit , the expected value is 0.1, but this currently returns 1.1

In my opinion, the correct calculation needs to subtract the intial capital amount (100%), i.e. 1.0 .
Otherwise this function needs to be renamed to something that more closely resembles its purpose, as when I first used the ReturnCriterion, I was expecting it to output the percent profit or loss. Having 100% profit or 100% returns means doubling the original account balance.

phong-phuong:
Fixes #976

- Reduce return amount by 1.0
- Update unit tests to also account for this

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/criteria/pnl/ReturnCriterion.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/criteria/pnl/ReturnCriterion.java
@@ -49,7 +49,8 @@ public class ReturnCriterion extends AbstractAnalysisCriterion {
         return tradingRecord.getPositions()
                 .stream()
                 .map(position -> calculateProfit(series, position))
-                .reduce(series.one(), Num::multipliedBy);
+                .reduce(series.one(), Num::multipliedBy)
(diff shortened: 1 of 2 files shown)


phong-phuong:
Canelled pull request as this would break existing code bases.

TheCookieLab:
Just because it’s a breaking change doesn’t necessarily mean it shouldn’t
be done. There is a “breaking changes” section in the CHANGELOG after all.

As with any change we must weigh the value add vs the cost in order to
guide a decision. Don’t let the lack of immediate feedback discourage you,
at least not until ~2 weeks have elapsed ;P

phong-phuong:
Ok, re-issued pull request: #978

Issue #982, opened 2023-06-01
Add ability to set begin time in the bar


phong-phuong:
Begin time currently of a bar is assumed to be the closeTime - bar duration, which is not always the case, but it is useful as users have one less field to deal with.

Begin Time is already in the BaseBarBuilder.

I will create a PR for this feature.

phong-phuong:
By the way, this library can't even properly support monthly candles as a result of this **fixed duration constraint**, since days in a month can vary.

Here are some snippets of logs for the open time:

Begin time: 2017-09-01T00:00Z
Begin time: 2017-10-0**2**T00:00Z
Begin time: 2017-11-01T00:00Z
Begin time: 2017-12-0**2**T00:00Z
Begin time: 2018-01-0**2**T00:00Z
Begin time: 2018-01-**30**T00:00Z

PR #983, opened 2023-06-01
Feature #982 ability to set begin time of a bar

phong-phuong:
Implementation for Feature #982.

- Add ability to optionally set beginTime using BaseBarBuilder instead of relying on the calculated Duration

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BaseBar.java
@@ -286,6 +286,34 @@ public class BaseBar implements Bar {
     }
 
     /**
+     * Constructor.
+     * 
+     * @param timePeriod the time period
+     * @param beginTime  the begin time of the bar period
+     * @param endTime    the end time of the bar period
+     * @param openPrice  the open price of the bar period
+     * @param highPrice  the highest price of the bar period
+     * @param lowPrice   the lowest price of the bar period
+     * @param closePrice the close price of the bar period
+     * @param volume     the volume of the bar period
+     * @param amount     the amount of the bar period
+     * @param trades     the trades count of the bar period
+     */
+    public BaseBar(Duration timePeriod, ZonedDateTime beginTime, ZonedDateTime endTime, Num openPrice, Num highPrice,
+            Num lowPrice, Num closePrice, Num volume, Num amount, long trades) {
+        this.timePeriod = timePeriod;
+        this.beginTime = beginTime;
+        this.endTime = endTime;
+        this.openPrice = openPrice;
+        this.highPrice = highPrice;
+        this.lowPrice = lowPrice;
+        this.closePrice = closePrice;
+        this.volume = volume;
+        this.amount = amount;
+        this.trades = trades;
+    }
+
+    /**
      * Returns BaseBarBuilder
      * 
      * @return builder of class BaseBarBuilder
--- a/ta4j-core/src/main/java/org/ta4j/core/BaseBarBuilder.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BaseBarBuilder.java
@@ -31,6 +31,7 @@ import org.ta4j.core.num.Num;
 public class BaseBarBuilder {
 
     private Duration timePeriod;
+    private ZonedDateTime beginTime;
     private ZonedDateTime endTime;
     private Num openPrice;
     private Num closePrice;
@@ -45,6 +46,11 @@ public class BaseBarBuilder {
 
     public BaseBarBuilder timePeriod(Duration timePeriod) {
         this.timePeriod = timePeriod;
+        return this;
+    }
+
+    public BaseBarBuilder beginTime(ZonedDateTime beginTime) {
+        this.beginTime = beginTime;
         return this;
     }
 
@@ -89,6 +95,11 @@ public class BaseBarBuilder {
     }
 
     public BaseBar build() {
-        return new BaseBar(timePeriod, endTime, openPrice, highPrice, lowPrice, closePrice, volume, amount, trades);
+        if (beginTime != null) {
+            return new BaseBar(timePeriod, beginTime, endTime, openPrice, highPrice, lowPrice, closePrice, volume,
+                    amount, trades);
+        } else {
+            return new BaseBar(timePeriod, endTime, openPrice, highPrice, lowPrice, closePrice, volume, amount, trades);
+        }
     }
 }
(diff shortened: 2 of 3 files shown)


phong-phuong:
Hold up, need to fix something

phong-phuong:
This use case is mainly aesthetic as binance uses close time for example 1st June 23 hours 59 minutes and 59.999999999999 seconds, resulting in the open time to be 31st May 23 hours 59 minutes and 59.9999999999999 seconds

phong-phuong:
However there could be use other cases in other types of bars that don't use strict time periods such as Renko bars.

phong-phuong:
> Why don't you round up such seconds? It's easy to do so..

If someone owes you 0.000000001 trillion dollars, would you round down? It's easier to do so? =)

team172011 (approved):

team172011:
@nimo23 but in general it makes sense to be able to add begin time to a Bar/BarBuilder nor?

team172011:
okay, so with this PR we would break the invariant that the `duration` always describes the time frame between `endTime` and `startTime`. 

> This use case is mainly aesthetic as binance uses close time for example on a daily candle of 1st June 23 hours 59 minutes and 59.999999999999 seconds, resulting in the open time to be 31st May 23 hours 59 minutes and 59.9999999999999 seconds. While I could just add 1 nanosecond to the close time and be done with it, this is technically not right either.

@phong-phuong does this mean that in your case biance provides you with bars of **different durations** or is the duration always the same but not equal to `Duration.ofDays(1)` but `Duration.ofDays(1).plusMillis(1)`

phong-phuong:
Neither. Binance uses an time which is 1 millisecond less than the duration. So ta4j users would have to either add 1 millisecond to end time to correct this.

However, with Renko bars, the duration is not fixed, without this feature, charting Renko bars is not possible.

To chart Renko bars in real time, updating the end time is also required for the very last bar to be extended in duration.

This leads to another design question. If the duration is fixed, why does every bar need a duration? It can be saved as an attribute in BarSeries to avoid duplication.

phong-phuong:
Duration can also be changed to be a method derived from end time minus start time. Prevents the need for a few extra constructors in the builder when mixing setting start time or setting duration. Too many contructors with double, num and string versions gets very messy and hard to know which constructor template to use when using autocomplete.
Just an opinion.

team172011:
> I don't know if this PR is enough to be able to use "Renko bars" or "Tick bars" (including use in indicators).

I think "renko bars" are no ohlcv bars. So once should not use ta4j `Bars` api to plot renko chart system. Renko bars are another charting system type. Theirfore this PR would not enalbe renko bars, it would only allow to break the condition, that the duration is always describing the intervall of the bars endTime & starTime.

I do not see any value in this enhancement and I would preffer to not merge this PR.

phong-phuong:
Dear developers of ta4j,

I appreciate the effort you have put into developing ta4j, and I understand that you have designed the constructors and methods based on certain considerations. However, I would like to present a counter argument regarding the openTime setting in bars.

While I understand that the duration of time may be considered fixed for certain use cases, it is important to acknowledge that bars serve as the fundamental building blocks for various financial analysis techniques, such as Renko bars and tick bars. These techniques require the ability to set the openTime explicitly, as the duration may not always be fixed or predictable.

By restricting the openTime and enforcing the use of closeTime and duration, you are introducing a dependency between bars and other classes, such as fixed time series. This dependency limits the flexibility and reusability of the bar implementation, as it becomes tightly coupled with other classes.

A more abstract and flexible approach would be to decouple the bar implementation from the specific use cases and allow users to set the openTime directly. This would enable the creation of different bar types without relying on fixed time series or making assumptions about the duration of time.

Furthermore, considering that bars are the lowest building blocks in financial analysis, they should provide a level of abstraction that allows users to define and manipulate them based on their specific requirements. By providing the freedom to set the openTime, you would empower users to design custom bar types and strategies that are not limited by predefined durations.

In conclusion, I believe that providing the option to set the openTime in bars would enhance the flexibility, reusability, and extensibility of ta4j. It would enable the implementation of various bar types, including Renko bars and tick bars, without introducing unnecessary dependencies or assumptions about fixed time durations. I hope you will consider this counter argument and evaluate the potential benefits of incorporating this feature into ta4j.

Thank you for your consideration.

phong-phuong:
> > It would enable the implementation of various bar types, including Renko bars and tick bars, without introducing unnecessary dependencies or assumptions about fixed time durations.
> 
> I also want such support. However, only providing the manual setting of `bar.openTime` does not provide support for `RenkoBar` or `TickBar`, nor?
> 
> The `Bar.class` is indeed an interface but only for ohlcv-bars, thus we need a more general approach, e.g. `Bar` (adapted interface) and rename `BaseBar` to `OhlcvBar` and create two new classes `RenkoBar` and `TickBar`. Or we add a new `enum BarType {Ohlcv, Tick, Renko;} `and use the property `BarType` within `Bar.class` . However, then we must (maybe) restrict the use of a set of indicators only usable with `OhlcvBar` and so forth, etc.
> 
> The point is: we need a more holistic solution to provide different kind of bar-types. I actually have no time to develop such a solution, but @phong-phuong feel free to find a full solution and we can see.

I have not performed an exhaustive test, however, a tick, Renko or even Binance's bars work with the current indicators without modification since indicators have no knowledge or concept of time, but instead rely on indexes. I don't see see the need to restrict users who generate their own type of bars from using the existing indicators, but feel free to explain why.

The only place where Bar.getPeriod() is used is in BarAggregator. For tick and Renko bars, you are correct, these will not work with BarAggregator and will mostly likely throw illegal argument exception when the multiplier is not correct.

I won't be using BarAggregator, so it wouldn't bother me if it doesn't work with Renko or Binance bars, but maybe you are right to think about the overall picture.

However the intent of this PR is giving that flexibility for the user to create these types of bars without having to generate a new BarSeriesBuilder class that is practically a copy and paste of an existing class with a few modifications. Either way, this extended discussion is not worth the effort for a simple modification.

phong-phuong:
> The Indicators use Ohlcv-Bars and have an instrinsic knowlege of time, i.e. the scale must be well formed to be able to measure. You cannot compare a 20 minute ohlcv-bar with a 17 minute ohlcv-bar because those two bars are NOT normalized (in this case: by time).

TradingView supports mixing Renko bars with every other indicator. As you mentioned, time is not uniform with Renko Bars, a Renko user knows this, instead Renko is based on price movement. Ultimately, a user should not restricted to use technical analysis with their own type of bar, a user wants to use the technical analysis in order to test their theory. Whether, it is right or wrong, whether it will be profitable, or whether the behaviour differs in backtest vs forward test matches is another long winded discussion that I'd rather not get into.

image

image

> But others could use it..so we should always think about the overall picture..however, we could simply restrict the use of BarAggregator to ohlcv-bars..

It would likely throw the IllegalArgumentException due to the multiplier, and that is fine, as I don't see why a person using Renko bars would even use this BarAggregator. From my quick glance, it's used for example to convert 1 hour bars to 4 hours, which serves no purpose in Renko.

PR #984, opened 2023-06-03
Add TradeExecutionModel to modify backtest behaviour, Fixes #598

### LINKED ISSUE #598: Results on Backtesting
Hi,

First of all I would like to thank you for all of your effort you spend in this lib - great work.
I have tested the backtesting  - when you run a strategy and analyse tradingRecord i was wondering if you take the right bar for the entry:

 - in the trading records the index of entry is the index where the indicator rule is true (index x)
 - you save the close price(AssetPrice) of the bar x in the trade order

But in general that is not true for live trading:
 - you get bar X in the stream and analyze your indicator and decide to enter the trade
 - then you decide to enter the market and you will get the Open price of bar x+1

Do I get something wrong ? If this is the case then the backtesting results are not correct?

phong-phuong:
Fixes #598.

- Adds a new interface: TradeExecutionModel, which allows trade execution behaviour to be modified
- Two implementations of TradeExecutionModels are provided, `TradeOnOpenModel`, and `TradeOnCloseModel` 
- Update BarSeriesManager to use `TradeOnOpenModel` by default if no TradeExecutionModel is specified
- Updates existing tests to use `TradeOnCloseModel` so logic is not affected
- Basic test added to ensure `TradeOnOpenModel` opens on the next bar and at the open price
- Moved `BarSeriesManagerTest` from root package to backtest package

Breaking:
- Moved `BarSeriesManager` and `BacktestExecutor` from root package to backtest package


phong-phuong:
This interface is under the trade package. The implementation is under trade/execution package according to my memory. I chose those folders because these can potentially be used in live trading bot too, not just for backtesting.

phong-phuong:
Well, there are a minority of traders may want to execute close to the closing time. This interface allows the user to define their own behaviour if the don't like either implementation. Some trade ticks, some want to simulate random slippage.

This obeys the open/closed principle, rather than having to copy paste and implement their own custom BarSeriesManager that does something slightly different.

phong-phuong:
> Could we rename the following classes:?
> 
> * `TradeExecutionModel` to `TradeExecution`
> * `ExecuteOnClosePriceModel` to `TradeOnClosePrice`
> * `ExecuteOnNextBarOpenPriceModel` to `TradeOnNextBarOpenPrice`
> 
> Btw, we don't use the suffix `*Model` for any of our other classes in ta4j.

This is what I'm seeing:

image

phong-phuong:
The naming is now as follows:
TradeExecutionModel - interface
**TradeOnCloseModel** - implementation
**TradeOnOpenModel** - implementation

Removed the description `OnNextBar` for a shorter name, and it is inferred to be the next bar as user can't back in time and trade on the open of a bar that has already closed. The class comments explain what each does though when considering bar at index `t`.

ta4j-core/src/main/java/org/ta4j/core/BarSeriesManager.java
```diff
@@ -48,13 +49,26 @@ public class BarSeriesManager {
     private final CostModel transactionCostModel;
     private final CostModel holdingCostModel;
 
+    /** The trade execution model to use */
+    private final TradeExecutionModel tradeExecutionModel;
+
     /**
      * Constructor.
```
nimo23: Please write javadoc 

> Constructor with {@link TradeExecutionModel} == {@link TradeExecutionModel#getDefault()} .

This allows the user to read which TradeExecutionModel is being used without looking at the source code.

ta4j-core/src/main/java/org/ta4j/core/execution/trade/TradeExecutionModel.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.ta4j.core.BarSeries;
+import org.ta4j.core.TradingRecord;
+import org.ta4j.core.num.Num;
+
+/**
+ * An execution model for {@link BarSeriesManager} objects.
+ *
+ * Used for backtesting. Instructs {@link BarSeriesManager} on how to execute
+ * trades
```
nimo23: Please add a point after trades.

ta4j-core/src/main/java/org/ta4j/core/execution/trade/TradeOnOpenModel.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * An execution model for {@link BarSeriesManager} objects.
+ *
+ * Executes trades on the next bar at the open price.
+ * 
+ * This is used for strategies that explicitly trade just after a new bar opens
+ * at bar index `t + 1`, in order to execute new or close existing trades as
+ * close as possible to the opening price.
+ */
+public class TradeOnOpenModel implements TradeExecutionModel {
```
nimo23: The behavior of `TradeOnOpenModel` **for the last bar in the barSeries** is that a possible trade is missed due to `t+1`, while `TradeOnCloseModel` (at t) trades with the current bar's close price. Could we add such information to the javadoc?

Another alternative is: Exceptionally, we can use the `closePrice` of the `currentBar` when the `currentBar` is the `lastBar` in the barSeries. But I'm not sure if this idea is ok. 

Another alternative is: Exceptionally, we can use the `closePrice` of the `currentBar` when the `currentBar` is the `lastBar` in the barSeries AND the last trade is already open and it could be closed by using the closePrice of the last Bar. But I'm not sure if this idea is ok. In fact, we already have this behavior (trying to close the last trade in the series) but this possibility would actually be bypassed when using `TradeOnOpenModel`.

We should think about it.

phong-phuong: I'm unsure why this behaviour was implemented as it did more than what a user would it would do.

However, I'll re-implement that feature by closing at the last closing price, if the strategy says to close.
```
                if (strategy.shouldOperate(i, tradingRecord)) {
                    // if this is the very last bar in the series
                    **if(i == barSeries.getEndIndex()) {
                        // close it at the last available price
                        tradingRecord.operate(i, barSeries.getBar(i).getClosePrice(), amount);
                    }** else {
                        tradeExecutionModel.apply(i, tradingRecord, barSeries, amount);
                    }
```

ta4j-core/src/main/java/org/ta4j/core/BarSeriesManager.java
```diff
@@ -198,7 +229,13 @@ public TradingRecord run(Strategy strategy, TradeType tradeType, Num amount, int
                 // For each bar after the end index of this run...
                 // --> Trying to close the last position
                 if (strategy.shouldOperate(i, tradingRecord)) {
-                    tradingRecord.operate(i, barSeries.getBar(i).getClosePrice(), amount);
+                    // if this is the very last bar in the series
+                    if(i == barSeries.getEndIndex()) {
```
nimo23: Please revert this back to: 

```
for (int i = runEndIndex + 1; i < seriesMaxSize; i++) {
                // For each bar after the end index of this run...
                // --> Trying to close the last position
                if (strategy.shouldOperate(i, tradingRecord)) {
                    tradeExecutionModel.apply(i, tradingRecord, barSeries, amount);
                    break;
                }
            }
```

This was already correct.

The code `if(i == barSeries.getEndIndex())` does not have any effect as `i` could be after `barSeries.getEndIndex()`, that's the reason why we try to close the last position..

The javadoc explains it:

```
// If the last position is still open and there are still bars after the
// endIndex of the barSeries, then we execute the strategy on these bars
// to give an opportunity to close this position.
```

My assumption to use the `closePrice` also for `TradeOnOpenModel` was wrong. It should also use the `openPrice` for the bar even after the run index..

phong-phuong:
There is also a BarSeriesManagerTest which would also need to move to its own backtest package, however there are some issues:

- I'm not using MockBar because the existing `MockBar` sets the open, high, low to zero
- The existing Doji related tests were built based on this behaviour

So instead, I've used `new BaseBarSeries()` but the BaseBarSeries constructor is not public, and is not accessible when I move the package to backtest

ta4j-core/src/main/java/org/ta4j/core/BaseBarBuilder.java
```diff
@@ -40,7 +40,7 @@ public class BaseBarBuilder {
     private Num volume;
     private long trades;
 
-    BaseBarBuilder() {
+    public BaseBarBuilder() {
```
nimo23: Sorry, you must revert it (remove `public`) and leave it without modifier. 

You should build your BaseBar by  `BaseBar.build(..)`:

https://github.com/ta4j/ta4j/blob/68c3d2c0cf20aaeff923ae9f993243c90be7520a/ta4j-core/src/main/java/org/ta4j/core/BaseBar.java#L293

ta4j-core/src/main/java/org/ta4j/core/backtest/BarSeriesManager.java
```diff
@@ -48,13 +52,26 @@ public class BarSeriesManager {
...
+     * Constructor.
+     * 
+     * @param barSeries            the bar series to be managed
+     * @param tradeExecutionModel the trade execution model to use
+     */
+    public BarSeriesManager(BarSeries barSeries, TradeExecutionModel tradeExecutionModel) {
+        this(barSeries, new ZeroCostModel(), new ZeroCostModel(), tradeExecutionModel);
     }
 
     /**
```
nimo23, line 77: Add the following javadoc to the following constructor:

```
/**
 * Constructor with {@link #tradeExecutionModel} = {@link TradeOnOpenModel}.
 * ..
*/
public BarSeriesManager(BarSeries barSeries, CostModel transactionCostModel, CostModel holdingCostModel){
..
}
```


With this, the user knows which `TradeExecutionModel` is used within this constructor by reading the javadoc (and without looking into the source code).

ta4j-core/src/main/java/org/ta4j/core/backtest/BarSeriesManager.java
```diff
@@ -68,6 +85,23 @@ public BarSeriesManager(BarSeries barSeries, CostModel transactionCostModel, Cos
         this.barSeries = barSeries;
```
nimo23, line 85: Add the javadoc to its surrounding constructor:

`Constructor with {@link #tradeExecutionModel} = {@link TradeOnOpenModel}.`

ta4j-core/src/main/java/org/ta4j/core/backtest/BarSeriesManager.java
```diff
@@ -48,26 +52,56 @@ public class BarSeriesManager {
...
     /**
-     * Constructor.
-     * 
+     * Constructor with {@link #tradeExecutionModel} = {@link TradeOnOpenModel}.
+     *
      * @param barSeries the bar series to be managed
      */
     public BarSeriesManager(BarSeries barSeries) {
-        this(barSeries, new ZeroCostModel(), new ZeroCostModel());
+        this(barSeries, new ZeroCostModel(), new ZeroCostModel(), TradeExecutionModel.getDefault());
```
nimo23: Please change `TradeExecutionModel.getDefault()` to `new TradeOnOpenModel()`

ta4j-core/src/main/java/org/ta4j/core/backtest/TradeExecutionModel.java
```diff
@@ -0,0 +1,42 @@
...
+import org.ta4j.core.TradingRecord;
+import org.ta4j.core.num.Num;
+
+/**
+ * An execution model for {@link BarSeriesManager} objects.
+ *
+ * Used for backtesting. Instructs {@link BarSeriesManager} on how to execute
+ * trades.
+ * 
+ * Constructor with {@link TradeExecutionModel} == {@link TradeExecutionModel#getDefault()}.
```
nimo23: Please remove the sentence `Constructor with {@link TradeExecutionModel} == {@link TradeExecutionModel#getDefault()}.` from the javadoc of the class.

PR #995, opened 2023-06-21
Feature/slade wilson indicators trailing atr stop loss

TheCookieLab:
- added **TrailingATRStopLossRule**


team172011:
@TheCookieLab isn‘t this rule a little bit to specific for adding it to this lib. Following this we could add a rule that is somehow based on every existing indicator. Or is this a common use case?

TheCookieLab:
Yes to both. While a trailing ATR based stop is a very standard technique
it is also somewhat precedent breaking in its specificity when compared to
our existing rules.

It really depends on the philosophy of the library’s rule approach. Is the
existing rule set supposed to be a finished product or just a starting
point? This PR proposes the latter.

On the other hand since specific strategy implementations are located in
the ta4j examples project and not the main library we can likely infer that
the original intent was the former. That said having indicators (which
exist across a spectrum of specificity) in the main library really blurs
the lines.

Do you think the ta4j examples is a better fit for this?

TheCookieLab:
Per discussion, the ta4j-examples project is the preferred home for something like this. Closing PR.

PR #1001, opened 2023-06-23
update pom.xml

nimo23:
Fixes #.

- updates plugins and dependencies of pom.xml 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,9 +28,12 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - updated pom.xml: slf4j-api to 2.0.7
 - updated pom.xml: org.apache.poi to 5.2.3
 - updated pom.xml: maven-jar-plugin to 3.3.0
+- add `final` to properties where possible
 - improved javadoc
+- add missing `@Override` annotation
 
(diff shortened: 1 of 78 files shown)


team172011 (approved):

TheCookieLab:
> The `pom.xml` has a very old version of `bnd-maven-plugin`.
> 
> Is it okay to change
> 
> ```
> <dependency>
>     <groupId>biz.aQute.bnd</groupId>
>     <artifactId>bnd-maven-plugin</artifactId>
>     <version>3.3.0</version>
>    ...
> </dependency>
> ```
> 
> to
> 
> ```
> <dependency>
>     <groupId>biz.aQute.bnd</groupId>
>     <artifactId>biz.aQute.bndlib</artifactId>
>     <version>6.4.1</version>
>     ...
> </dependency>
> ```

Why change the artifactId? There are some additional changes needed if we update to 6.4.0 so best to leave this off for now.



> Btw, would it be possible to change this as well:
> 
> https://github.com/ta4j/ta4j/blob/0c604253f09601849a9fe4ad0e30ae1301957f83/.travis.yml#L9
> 
> from `openjdk8` to `openjdk11`. Ok?

I don't see why not but the only way to know for sure is to let the CI run and see the result.

PR #1004, opened 2023-06-24
added `NumIndicator`

nimo23:
Fixes #.

- added **NumIndicator** to calculate any `Num`-value for a `Bar`

_The `NumIndicator` complements the existing `DateTimeIndicator`.
With the `NumIndicator`, we can extract or calculate any `Num`-value from/for a bar._

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/NumIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/NumIndicator.java
@@ -1,0 +1,73 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017-2023 Ta4j Organization & respective
+ * authors (see AUTHORS)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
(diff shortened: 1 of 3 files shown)


team172011:
I am okay with this, but we should keep in mind not to generify existing code with this indicator. If I remember correct we had times where we removed all `Close/Open/High/Low/VolumeIndicators` usages by such a generic indicator and afterwards reverted everything because it got to complex and unreadable

team172011 (approved):

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/NumIndicator.java
```diff
@@ -0,0 +1,72 @@
...
+public class NumIndicator extends CachedIndicator<Num> {
+
+    private final Function<Bar, Num> action;
+
+    /**
+     * @param barSeries the bar series
+     * @return close price of a bar
+     */
+    public static NumIndicator ClosePriceIndicator(BarSeries barSeries) {
+        return new NumIndicator(barSeries, Bar::getClosePrice);
```
team172011: I would remove this static method

team172011 (approved):

PR #1008, opened 2023-06-25
Move average indicators to package `indicators/average`

nimo23:
Fixes #.

- created a new package `indicators/average`
- **AbstractEMAIndicator** moved to package `indicators/average`
- **DistanceFromMAIndicator** moved to package `indicators/average`
- **DoubleEMAIndicator** moved to package `indicators/average`
- **EMAIndicator** moved to package `indicators/average`
- **HMAIndicator** moved to package `indicators/average`
- **KAMAIndicator** moved to package `indicators/average`
- **LWMAIndicator** moved to package `indicators/average`
- **MACDIndicator** moved to package `indicators/average`
- **MMAIndicator** moved to package `indicators/average`
- **SMAIndicator** moved to package `indicators/average`
- **TripleEMAIndicator** moved to package `indicators/average`
- **WMAIndicator** moved to package `indicators/average`
- **ZLEMAIndicator** moved to package `indicators/average`

_Now that we have a package `indicators/average`, we can quickly see which indicators are of type "average indicators". Previously, these indicators were scattered untidily in the root package `indicators`. Organization is half the battle._


team172011:
I do not agree with this change, because imo it is not worth to break the api with this new package

TheCookieLab:
We'll just have to agree to disagree then. As maintainers it's our job to use our personal judgement (aka bias) to steer the code-base in what we believe to be the best direction. I do not see net value in this so will not be merging it. Please feel free to make this change in your own fork where you have full discretion on the direction the code-base takes.

team172011:
>  I will not continue to argue or engage in conversation until my request is granted.

I think everything has been said here, especially about objectivity

PR #1036, opened 2023-07-14
InPipeRule: improved calculation

nimo23:
Fixes #.

- **InPipeRule** improved calculation by avoiding redundant calculation
- 
- 

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/rules/DayOfWeekRule.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/rules/DayOfWeekRule.java
@@ -55,7 +55,7 @@ public class DayOfWeekRule extends AbstractRule {
     /** This rule does not use the {@code tradingRecord}. */
     @Override
     public boolean isSatisfied(int index, TradingRecord tradingRecord) {
-        ZonedDateTime dateTime = this.timeIndicator.getValue(index);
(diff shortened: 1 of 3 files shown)


ta4j-core/src/main/java/org/ta4j/core/rules/DayOfWeekRule.java
```diff
@@ -55,7 +55,7 @@ public DayOfWeekRule(DateTimeIndicator timeIndicator, DayOfWeek... daysOfWeek) {
     /** This rule does not use the {@code tradingRecord}. */
     @Override
     public boolean isSatisfied(int index, TradingRecord tradingRecord) {
-        ZonedDateTime dateTime = this.timeIndicator.getValue(index);
```
team172011, line 58: why are you removing `this` keyword?

nimo23: It's redundant and unnecessarily bulks up the code. Should I do a new PR just because of something so small?

I can also ask, why not adding `this` to the variable below:

`final boolean satisfied = this.daysOfWeekSet.contains(this.dateTime.getDayOfWeek());`

:)

Only use "this" in constructors and methods where necessary, otherwise it's just completely redundant and unnecessarily lengthens the code and weakens readability.

team172011 (approved):

PR #1062, opened 2023-07-22
remove-unused-imports

nimo23:
Fixes #.

- remove unused imports
- move static imports to the top (putting them at the bottom is not a big issue, but in all other files we have static imports at the top, so I changed that here too for unification).

- [x] added an entry with related ticket number(s) to the unreleased section of `CHANGES.md` (already done with "clean code")

Please folks, as a matter of decency, take a look at your code in your editor and please **remove unused import declarations before doing PR** so I or your successor don't have to do this all the time. Thanks.


TheCookieLab:
While removing unused imports is OK, this PR has too many unnecessary changes. Re-arranging import lines is pointless and an exercise in re-arranging deck chairs. Going forward all PRs that are simply "re-arranging deck chairs" will be rejected. If you want to re-create this PR with only the removal of unused imports then by all means.

PR #1078, opened 2023-07-26
Position.java: improve code

nimo23:
Fixes #.

- make some methods sleeker and faster


TheCookieLab:
I don't see this as improving the code, rather it's just trading readability for brevity. Is it objectively shorter? Yes, but is it objectively better? I don't think this is necessary.

PR #1080, opened 2023-07-26
Position#getGrossReturn-addBase

nimo23:
Fixes #.

- added the parameter `addBase` to the following methods to be able to include or exlude the base from the result. The developer must now make a conscious decision whether to set `addBase` to `true` or `false` to avoid implicit bugs:
    - changed `Position.getGrossReturn()` to `Position.getGrossReturn(boolean addBase)`
    - changed `Position.getGrossReturn(Num finalPrice)` to `Position.getGrossReturn(Num finalPrice, boolean addBase)`
    - changed `Position.getGrossReturn(BarSeries barSeries)` to `Position.getGrossReturn(BarSeries barSeries, boolean addBase)`
    - changed `Position.getGrossReturn(Num entryPrice, Num exitPrice)` to `Position.getGrossReturn(Num entryPrice, Num exitPrice, boolean addBase)`
- With the above methods, we now make it explicit, whether a method (e.g. `EnterAndHoldCriterion`) is calculated with or without a basis - this will better avoid future bugs as the developer now must make a conscious choice whether to set `addBase` to true or false.
- With the above methods, we also solved a bug in  `ReturnCriterion#calculate(BarSeries series, Position position) `as we are now able to include or exclude the base per position.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
       - `BarSeriesManager manager = new BarSeriesManager(barSeries, new TradeOnCurrentCloseModel())`
       - `BarSeriesManager manager = new BarSeriesManager(barSeries, transactionCostModel, holdingCostModel, tradeExecutionModel)`
 - **BarSeriesManager** and **BacktestExecutor** moved to packge **`backtest`**
+- added the parameter `addBase` to the following methods to be able to include or exlude the base from the result. The developer must now make a conscious decision whether to set `addBase` to `true` or `false` to avoid implicit bugs:
+    - changed `Position.getGrossReturn()` to `Position.getGrossReturn(boolean addBase)`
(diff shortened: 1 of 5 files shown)


TheCookieLab:
The fact that we include the base/principal in the return value is a choice of convention and not some kind of bug. Unless egregiously off the wall, conventions are neither right or wrong, they are simply a choice. "Correcting" a convention just because it may not have been the one you would have chosen is misguided and only leads to unnecessary complexity. 

All frameworks and libraries have conventions, if a user is initially confused then the solution is simply to learn and adapt and not cater to every subjective preference out there. By doing the latter we establish precedent and potentially create a slippery slope of supporting all manners of representation variations.

Another potential problem with this change is negative returns. In a scenario where one wants to calculate aggregate returns of multiple investments, say, +10%, -2%, and +5% the current representation can correctly calculate it as 1.1 * 0.98 * 1.05 = 1.1254. 
 If we exclude the base then this becomes 0.1 * -0.02 * 0.05 = -0.0001 which is nonsensical.

PR #1100, opened 2023-08-23
Add a Running Total aka Cumulative Sum indicator

romanbsd:
This adds a utility indicator RunningTotalndicator

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/RunningTotalIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/RunningTotalIndicator.java
@@ -1,0 +1,64 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017-2023 Ta4j Organization & respective
+ * authors (see AUTHORS)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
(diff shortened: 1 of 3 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/RunningTotalIndicator.java
```diff
@@ -0,0 +1,63 @@
...
+
+    protected RunningTotalIndicator(Indicator<Num> indicator, int barCount) {
+        super(indicator);
+        this.indicator = indicator;
+        this.barCount = barCount;
+    }
+
+    @Override
+    protected Num calculate(int index) {
+        Num sum = numOf(0);
```
nimo23: use:

`Num sum = zero();`

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/RunningTotalIndicator.java
```diff
@@ -0,0 +1,63 @@
...
+    protected RunningTotalIndicator(Indicator<Num> indicator, int barCount) {
+        super(indicator);
+        this.indicator = indicator;
+        this.barCount = barCount;
+    }
+
+    @Override
+    protected Num calculate(int index) {
+        Num sum = zero();
+        for (int i = Math.max(0, index - barCount + 1); i <= index; i++) {
```
team172011: For getting the first valid index, please use `BarSeries#getBeginIndex` instead of hard coded zero

team172011 (changes requested):

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/RunningTotalIndicator.java
romanbsd: Should I also change the SMAIndicator? https://github.com/ta4j/ta4j/blob/master/ta4j-core/src/main/java/org/ta4j/core/indicators/SMAIndicator.java#L55

romanbsd:
Fixed the comments, improved the test class and squashed commits.

team172011 (approved):

PR #1109, opened 2023-09-04
add-Num#bigDecimalValue

nimo23:
Fixes #.

- added `Num.bigDecimalValue()` to convert the `Num` to a `BigDecimal`
    - For `DecimalNum` it returns its `delegate`
    - For `DoubleNum` it converts by `BigDecimal.valueOf(delegate)` (if delegate is `NaN` or `Infinity`, then it simply returns `null`)
    - For `NaN` it simply returns `null`.

_Actually, it's not easy to get the `BigDecimal`-representation of a `Num`. We have `intValue()`, `longValue()`, `doubleValue()`, etc. so it also makes sense to add `bigDecimalValue()`._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,24 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 
 ## 0.17
 
-- Added signal line and histogram to **MACDIndicator**
-- Implemented inner cache for **SMAIndicator**
-- Added getTransactionCostModel, getHoldingCostModel, getTrades in **TradingRecord**
-- **BooleanTransformIndicator** remove enum constraint in favor of more flexible `Predicate`
-- Fixed **ta4jexamples** project still pointing to old (0.16) version of **ta4j-core**
+### Breaking
 - Renamed **SMAIndicatorMovingSerieTest** to **SMAIndicatorMovingSeriesTest**
+
+### Fixed
+- Fixed **ta4jexamples** project still pointing to old (0.16) version of **ta4j-core**
 - Fixed **SMAIndicatorMovingSeriesTest** test flakiness where on fast enough build machines the mock bars are created with the exact same end time
+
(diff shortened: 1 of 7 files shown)


TheCookieLab:
> Would it be better to overwrite `bigDecimalValue()` in its subclasses instead of providing a default method?
> 
> For example, in `DecimalNum`:
> 
> ```
>  @Override
>  public BigDecimal bigDecimalValue() {
>         return delegate;
> }
> ```
> 
> With this, no **if** and **cast** is needed..

Yes, that would be ideal.

CHANGELOG.md
```diff
@@ -132,6 +132,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - added "TradeExecutionModel" to modify trade execution during backtesting
 - added **NumIndicator** to calculate any `Num`-value for a `Bar`
 - added **RunningTotalIndicator** to calculate a cumulative sum for a period.
+- added `Num.bigDecimalValue(DoubleNum)` to convert Num to a BigDecimal
```
TheCookieLab: This should be moved up to the 0.17 section

ta4j-core/src/main/java/org/ta4j/core/num/Num.java
```diff
@@ -354,6 +345,31 @@ default float floatValue() {
...
+        return getDelegate().doubleValue();
+    }
+
+    /**
+     * Converts this {@code Num} to a {@code BigDecimal}.
+     *
+     * @return this {@code Num} converted to a {@code BigDecimal}
+     */
+    default BigDecimal bigDecimalValue() {
+        Number num = getDelegate();
```
TheCookieLab: Can we add unit tests to **DoubleNumTest** and **DecimalNumTest** to exercise this branching logic, including the doubleNum edge cases?

TheCookieLab (approved):

PR #1121, opened 2023-10-26
Fixes erroneous SuperTrend calculation

### LINKED ISSUE #1120: [BUG] Supertrend Indicator value goes to 0 and stays 0 forever

**Describe the bug**
On the Supertrend indicator, when the close price value is the same as the previous supertrend indicator value, the supertrend value goes to zero as no code is assigning any value to it (initial assignment is = 0). After this happens once, the Supertrend value is always 0, never changes.

I think there's a branch missing here (the equal case): https://github.com/ta4j/ta4j/blob/master/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java#L81 
and here (also the equal case): https://github.com/ta4j/ta4j/blob/master/ta4j-core/src/main/java/org/ta4j/core/indicators/supertrend/SuperTrendIndicator.java#L89

**To Reproduce**
Please view this test I amended that proves the behavior I described above: https://github.com/ta4j/ta4j/compare/master...francescopeloi:ta4j:master

**Expected behavior**
I am not 100% sure what the right behavior is, but I would expect that in this special case either the Super Trend lower-band or the Super Trend upper-band gets assigned as Super Trend value. I can amend the code if this is accepted as the right behavior.

Possible fix:
```
        if (previousValue.isEqual(superTrendUpperBandIndicator.getValue(i - 1))) {
            // if equal and before the upperBand was used, keep the upperband as new value
            if (closePrice.isLessThanOrEqual(upperBand)) {
                value = upperBand;
            } else if (closePrice.isGreaterThan(upperBand)) {
                value = lowerBand;
            }
        }

        if (previousValue.isEqual(superTrendLowerBandIndicator.getValue(i - 1))) {
            // if equal and before the lowerBand was used, keep the lowerBand as new value
            if (closePrice.isGreaterThanOrEqual(lowerBand)) {
                value = lowerBand;
            } else if (closePrice.isLessThan(lowerBand)) {
                value = upperBand;
            }
        }
```

francescopeloi:
Fixes #1120 .

- On SuperTrendIndicator fixed calculation when close price is the same as the previous Super Trend indicator value

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,8 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - **BarSeries#getBeginIndex()** methode returns correct begin index for bar series with max bar count
 
 ### Fixed
--  **Fixed** **ParabolicSarIndicator** fixed calculation for sporadic indices
+- **Fixed** **SuperTrendIndicator** fixed calculation when close price is the same as the previous Super Trend indicator value
+- **Fixed** **ParabolicSarIndicator** fixed calculation for sporadic indices
 - **ExpectancyCriterion** fixed calculation
 - catch NumberFormatException if `DecimalNum.valueOf(Number)` is `NaN`
 - **ProfitCriterion** fixed excludeCosts functionality as it was reversed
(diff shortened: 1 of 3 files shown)


TheCookieLab (approved):

TheCookieLab:
@francescopeloi can you run the formatter i.e. `mvn -B clean license:format formatter:format test`

francescopeloi:
@TheCookieLab sure I can, done

TheCookieLab:
Thanks, appreciate your contribution.

PR #1127, opened 2024-01-05
Update Num interface to implement Serializable

gretard:
Fixes #.

- Updated Num interface to implement Serializable
- Updated classes implementing Num interface
- Added test

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
@@ -49,6 +49,8 @@ import org.slf4j.LoggerFactory;
  */
 public final class DecimalNum implements Num {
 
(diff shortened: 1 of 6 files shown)


ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
```diff
@@ -49,6 +49,8 @@
  */
 public final class DecimalNum implements Num {
 
+    private static final long serialVersionUID = 387498017007756808L;
```
TheCookieLab: Any reason not to use '1L` for the initial UIDs?

gretard: Great spot, updated to 1L.

TheCookieLab (approved):

PR #1149, opened 2024-04-18
Introduced live trading LiveIndicator

sgflt:
- which may fetch data from external source

- ta4j may analyse external data sources and create rules from them

For example we may use CPU temperature, AI generated numeric advision, current profit from exchange API as stop loss source.


sgflt:
Maybe choose better name?

ExternalIndicator, SuppliedIndicator...

sgflt:
Maybe it should be Rule.

sgflt:
Invented alternative named RuntimeContext within event based branch.

PR #1176, opened 2024-09-07
reused tr in constructor instead of creating new instance

TheCookieLab:
Fixes #[Any reason tr isn't being passed to MMA instead of creating a new instance?](https://github.com/ta4j/ta4j/commit/594c4e464418e162c1b1dcdd1bbed3f218c740ee#r146073872)

- re-use `tr `by passing it as a constructor param when initializing `averageTrueRangeIndicator`


jschoeff81690:
Thanks, was about to do this.

PR #1189, opened 2024-10-03
#691 Implemented TickBarBuilder

### LINKED ISSUE #691: Can I construct tick bars?
It is unclear to me if there is a way to construct tick bars? For example I have a tick data stream I get data from and would like to aggregate bars for live trading on every 70th tick. I think I could create fake bars with the same value for OHLC and then write my own BarAggregator. However im not sure if this is possible or if this breaks something else.

Is it possible to construct tick based bars and use them for live trading? If so could you show an example?

sgflt:
Fixes #691.

- it does add bar if certain amount of ticks passed
- removed convertible bar builder, because api incompatibility. Now each builder can accept various data types.


sgflt:
So should I modify actual code to count volume, or do you propose another mutually exclusive approach?

ta4j-core/src/main/java/org/ta4j/core/BarBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+package org.ta4j.core;
+
+import java.time.Duration;
+import java.time.ZonedDateTime;
+
+import org.ta4j.core.num.Num;
+
+/**
+ * @author Lukáš Kvídera
```
nimo23: Unfortunately, we should not use the `@autor`-tag within the class at the file can be changed/extended any time in the future. Please provide your name in https://github.com/ta4j/ta4j/blob/master/AUTHORS. It was also said once by @team172011.

ta4j-core/src/main/java/org/ta4j/core/BaseBarBuilder.java
```diff
@@ -71,16 +76,42 @@ public BaseBarBuilder endTime(ZonedDateTime endTime) {
...
+    @Override
+    public BarBuilder highPrice(final String highPrice) {
+        highPrice(this.numFactory.numOf(highPrice));
+        return this;
+    }
+
     /**
      * @param highPrice the highest price of the bar period
      * @return {@code this}
      */
```
nimo23: Please move all the general javadocs from the subclasses to the dedicated interface methods and leave only specific javadocs which overrides the behaviour (e.g. if a method throws an Exception because the value cannot be set on that kind of Bar).

ta4j-core/src/main/java/org/ta4j/core/TickBarBuilder.java
```diff
@@ -0,0 +1,213 @@
...
+    private Num closePrice;
+    private Num lowPrice;
+    private Num amount;
+    private long trades;
+
+    public TickBarBuilder(final NumFactory numFactory, final int countOfTicks) {
+        this.numFactory = numFactory;
+        this.volume = numFactory.zero();
+        this.highPrice = numFactory.zero();
+        this.lowPrice = numFactory.numOf(Integer.MAX_VALUE);
```
nimo23: Why not setting lowPrice to zero? Is it a tickBar-requirement?

ta4j-core/src/test/java/org/ta4j/core/TickBarBuilderTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+public class TickBarBuilderTest {
+
+    @Test
+    public void add() {
+        final var series = new BaseBarSeriesBuilder().withBarBuilderFactory(new TickBarBuilderFactory(5)).build();
+        series.barBuilder().timePeriod(Duration.ofDays(1)).endTime(ZonedDateTime.now()).closePrice(1).volume(1).add();
+        series.barBuilder()
+                .timePeriod(Duration.ofDays(1))
+                .endTime(ZonedDateTime.now().plusDays(1))
```
nimo23: Please use a `final var now = ZonedDateTime.now()` within the method and reference `now` within each bar, e.g.` .endTime(now.plusDays(1))`.

Do this also for all other methods or places of your tests.

ta4j-core/src/test/java/org/ta4j/core/TickBarBuilderTest.java
```diff
@@ -0,0 +1,112 @@
...
+                .add();
+        series.barBuilder()
+                .timePeriod(Duration.ofDays(1))
+                .endTime(ZonedDateTime.now().plusDays(8))
+                .closePrice(2)
+                .volume(1)
+                .add();
+        series.barBuilder()
+                .timePeriod(Duration.ofDays(1))
+                .endTime(ZonedDateTime.now().plusDays(9))
```
nimo23: Please use `.endTime(now.plusDays(9))` as described above..

ta4j-core/src/test/java/org/ta4j/core/mocks/MockBarBuilder.java
```diff
@@ -30,13 +30,13 @@
 import java.time.ZonedDateTime;
 
 import org.ta4j.core.BaseBar;
-import org.ta4j.core.BaseBarConvertibleBuilder;
+import org.ta4j.core.BaseBarBuilder;
 import org.ta4j.core.num.NumFactory;
 
 /**
  * @author Lukáš Kvídera
```
nimo23: Please remove the `@author` tag as explained above and  put your name in official AUTHORS-file.

ta4j-core/src/main/java/org/ta4j/core/BarBuilder.java
sgflt: Ah, I have to reconfigure IDE. It is added automatically.

sgflt:
I can't decide between a) 60% and b) 40% right now so I am going to do PoC and subjectively evaluate usability. 

Why do you think a new bar class is needed? I think all data can fit into existing bar implementation.

ta4j-core/src/main/java/org/ta4j/core/BarBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+public interface BarBuilder {
+    BarBuilder timePeriod(Duration timePeriod);
+
+    BarBuilder endTime(ZonedDateTime endTime);
```
TheCookieLab: Please merge in the latest from master

ta4j-core/src/main/java/org/ta4j/core/TickBarBuilder.java
```diff
@@ -0,0 +1,213 @@
...
+import java.time.ZonedDateTime;
+import java.util.Objects;
+
+import org.ta4j.core.num.Num;
+import org.ta4j.core.num.NumFactory;
+
+class TickBarBuilder implements BarBuilder {
+
+    private final NumFactory numFactory;
+    private final int countOfTicks;
```
TheCookieLab: How about tickCount?

ta4j-core/src/main/java/org/ta4j/core/TickBarBuilder.java
```diff
@@ -0,0 +1,213 @@
...
+    private int countOfTicksPassed;
```
TheCookieLab: and here, passedTickCount

ta4j-core/src/main/java/org/ta4j/core/TickBarBuilderFactory.java
```diff
@@ -0,0 +1,43 @@
...
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package org.ta4j.core;
+
+public class TickBarBuilderFactory implements BarBuilderFactory {
+
+    private final int countOfTicks;
```
TheCookieLab: naming, same as above.

ta4j-core/src/main/java/org/ta4j/core/aggregator/DurationBarAggregator.java
```diff
@@ -135,7 +135,7 @@ public List<Bar> aggregate(List<Bar> bars) {
             }
 
             if (!onlyFinalBars || i <= bars.size()) {
-                final Bar aggregatedBar = new BaseBarBuilder().timePeriod(timePeriod)
+                final Bar aggregatedBar = new BaseBarBuilder(null).timePeriod(timePeriod)
```
TheCookieLab: Why null here? If a null `numFactory `param is a common use case when instantiating `BaseBarBuilder`, instead of passing null as part of its API usage, `BaseBarBuilder `should have a default, no param constructor which then sets the null value internally.

ta4j-core/src/main/java/org/ta4j/core/TickBarBuilder.java
```diff
@@ -0,0 +1,213 @@
...
+    public Bar build() {
+        return new BaseBar(this.timePeriod, this.endTime, this.openPrice, this.highPrice, this.lowPrice,
+                this.closePrice, this.volume, this.amount, this.trades);
+    }
+
+    @Override
+    public void add() {
+        if (++this.countOfTicksPassed % this.countOfTicks == 0) {
+            this.barSeries.addBar(build());
+            this.openPrice = null;
```
TheCookieLab: This "reset" sequence could be in it's own method as it's done in multiple places. If it were 3+ places I would ask for it explicitly, but if it's only 2 I'll leave it to your discretion.

Help me understand the rationale of the differing reset values. `openPrice` and `closePrice` are reset to null, but `highPrice` and `lowPrice` are set to semantically "nonsense" values. Why the difference?

I don't do tick-based trading myself. What is the significance of checking `if (++this.countOfTicksPassed % this.countOfTicks == 0)` rather than a strict equality check. Wouldn't this allow bars with varying multiples of ticks?

TheCookieLab (changes requested):
Understand this is still in draft but added some comments. Agree in principle with the approach.

ta4j-core/src/main/java/org/ta4j/core/BaseBarBuilder.java
sgflt: done

ta4j-core/src/main/java/org/ta4j/core/TickBarBuilder.java
sgflt: It would break min/max algorithm.

ta4j-core/src/main/java/org/ta4j/core/TickBarBuilder.java
sgflt: "nonsense" values are chosen for min/max algorithm. High goes from zero to actual high and low goes from "infinity" to actual value.

Open/close are set to null as undefined value and no need of calculation.

For equality check I would need to reset countOfTicksPassed to zero. Modulo is reset free.

sgflt:
> There are two options. But I don't know which approach is better:
> 
> a) I thought about abstracting the TickBarBuilder to something like a `CountBasedBarBuilder` which counts anything you like by providing either a Function or an enum {Volume, Ticks, EndPrice, Duration, ..}.
> 
> b) Provide another BarBuilder-implementation called VolumeBarBuilder.
> 
> Which way do you prefer?

b) is winner, implementations need flexible handling of their specifics

ta4j-core/src/test/java/org/ta4j/core/TickBarBuilderTest.java
sgflt: done

ta4j-core/src/test/java/org/ta4j/core/mocks/MockBarBuilder.java
sgflt: done

ta4j-core/src/main/java/org/ta4j/core/TickBarBuilder.java
```diff
@@ -0,0 +1,215 @@
...
+    @Override
+    public void add() {
+        if (++this.passedTicksCount % this.tickCount == 0) {
+            this.barSeries.addBar(build());
+            reset();
+        }
+    }
+
+    private void reset() {
+        this.openPrice = null;
```
nimo23, line 210: Do:
```
var zero = numFactory.zero();
..
this.highPrice = zero;
..
this.volume = zero;
```

ta4j-core/src/main/java/org/ta4j/core/BarBuilder.java
sgflt: done

ta4j-core/src/main/java/org/ta4j/core/TickBarBuilder.java
sgflt: done

ta4j-core/src/main/java/org/ta4j/core/aggregator/DurationBarAggregator.java
sgflt: done

TheCookieLab (approved):
Looks good.

PR #1148, opened 2024-04-14
Simplify ux optimize performance of backtesting and livetrading

sgflt:
Fixes #1141 
Fixes #1134.

Changes proposed in this pull request:

- optimize performance of backtesting
  - `Indicator.getStableBars()` replaced by `boolean isStable()`
    - implemented propagation of stability 
  - optimized running sums, uses previous partial sum
  - number factories caches constants with runtime defined precision
  - Optimized `RecursiveIndicator` -> no more StackOverflowException
  - Optimized `LowestValueIndicator`
  - Optimized `CorrelationCoefficientIndicator`, `VarianceIndicator`,`CovarianceIndicator` using streaming algorithms
- optimize memory footprint of live trading
  - raspberry pi power save mode enabled
- enhance parallelization to indicator level
- bump to JDK 21
- general code clean up
  - numeric indicators uses `NumericIndicator` as a base, not as wrapper
   - `SeriesRelatedNumericIndicator` - indicator pulls data from series directly
   - `NumericIndicator` - pulls data from another indicator
   - `SeriesRelatedBooleanIndicator` - pulls data from series directly and produces boolean
   - `BooleanIndicator` - pulls data from another indicator
   - `Chandelier*ExitIndicator` - moved from Num to Boolean, is it is meaningful only as Rule
  - package clean up
    - moving averages moved to own package
    - channels moved to own package
    - oscillators moved to won package
    - all numeric indicators are under numeric package
    - all boolean indicators are under bool package
  - fixed javadocs
    - License headers should be comments 
    - `DifferenceIndicator`
- Used java.time.Instant as time source instead of ZonedDateTime
- new assert methods
  - assertNext - moves to next bar and compares values
  - fastForward - passes N ticks to future 
- Introduced `NormalizedBar` for Machine Learning / AI training

IDEAS:
- untangle indicators from BarSeries but preserve Statefulness
  - allows creation of MACD from sun spot count and other sorcery things 
- allow custom naming of indicaors
  - allows persistence of calculations into TimescaleDB for AI analysis 

Cons:
- dropped serialization support
- dropped mutable last bar (why trade on "unstable" candle?)
  - mutability has performance impact (maybe exists another solution)

- [ ] I like the final API
- [ ] added tests for stability API
- [ ] all tests passed
- [ ] Performance requirements met
- [ ] added an entry with related ticket number(s) to the unreleased section of `CHANGES.md`

PR #1190, opened 2024-10-05
Feature/heikinashi

### LINKED ISSUE #1186: Implement Heikin ashi bar

https://www.investopedia.com/trading/heikin-ashi-better-candlestick/

- [ ] Maybe BaseBar should be renamed to OHLCBar.
- [ ] Maybe just use different BarBuilder.

prackky:
Fixes #1186 

- Converts given bar to HeikinAshi if BaseBarSeries is bound
- Test Cases added
- Example to create HeikinAshi Bar added

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/BaseBarBuilder.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BaseBarBuilder.java
@@ -27,6 +27,8 @@ import java.time.Duration;
 import java.time.ZonedDateTime;
 import java.util.Objects;
 
+import org.ta4j.core.num.DecimalNum;
+import org.ta4j.core.num.DoubleNum;
 import org.ta4j.core.num.Num;
 
 /**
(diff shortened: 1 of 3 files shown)


sgflt:
Personaly I would rather implement HeikinAshiBarBuilder, HeikinAshiBarBuilderFactory as it would be usable transparently like:

```java
var series = seriesBuilder.withBarBuilderFactory(HeikinAshiBarBuilderFactory).build();
series.barBuilder()  
  ....
  .add();
```

I am not sure how toHeikinAshiBar() fits in this fluent style.

But I am just contributor. Leaders of this project have the last word.

prackky:
So, rather than modifying the core classes we should create a BarBuilderFactory? And this will be helpful when adding other Bar implementations as well. Thanks @sgflt for the advice.

sgflt:
From my point of view yes. It does not make sense mix multiple types of Bars in single BarSeries. BarBuilderFactory ensures this consistency.

Also I would hold reference to previous Bar inside factory directly instead of using indexes on BarSeries, because of [PoC](https://github.com/ta4j/ta4j/pull/1148) that does not use indexes and treats series as discrete simulation.

PR #1192, opened 2024-10-08
cleaned code by using var, record, text-blocks

nimo23:
Fixes #.

- Cleaned code by using new java syntax like `var`, `record`, `text blocks`
(which leads to better maintainability, better readability, less imports, less size, ..)

- Faster test execution by using new `String.lines()` where possible
(better performance, better readability and on my machine it saves at least approx. 4 seconds testing)


dakics:
var = death

I don't get what would be the benefits of removal of types from the entire code...

I get it's more convenient, but I see stronger types as a feature. Everything being 'var' makes it less readable ("what is the type of X?...lets find the method") and potentially less stable for this particular use case of code potentially managing money.

dakics:
I didn't say using 'var' would turn it into JavaScript at runtime:) I stated my personal opinion that removal of explicit type declarations doesn't help, and in this particular project I would prefer as little ambiguity as possible.

But your opinion on this differs and, as I am not a contributor here, I respect that mine is not as important, so you do you and I wish you well.

TheCookieLab:
I agree with @dakics my preference is for explicit typing but I'll chalk it up to stylistic choice. Going forward contributors can use '`var` if they want, however removing the existing explicit typing for no reason beyond simply doing so will not be accepted.

Issue #1132, opened 2024-02-18
[BUG] VersusEnterAndHoldCriterion should use 'amount' to match comparsion


Cashmeiser:
**I have checked existing issues and wiki**
- [X] I could not find similar [issues](https://github.com/ta4j/ta4j/issues?utf8=%E2%9C%93&q=)
- [X] I could not find a solution in the [wiki](https://ta4j.github.io/ta4j-wiki/) or [faq section](https://ta4j.github.io/ta4j-wiki/FAQ.html)

**Describe the bug**
When running a Backtest with `BarSeriesManager` a `amount` is provided and used as the quantity of each trade.
When running a `VersusEnterAndHoldCriterion` it always uses 1 as the `amount` (or quantity) for the buy and sell trade.
If comparing with a non-percentage criterion it will be a non-useful comparison.

Lets see an example: 
Run a backtest with `BarSeriesManager` with `amount` 10 (quantity).
Then run a `new VersusEnterAndHoldCriterion(new ProfitLossCriterion()).calculate(barSeries, tradingRecord)` 
and it will produce a value 10 times greater than it actually is, due to not creating the "enter and hold" with same `amount` (quantity).

**To Reproduce**
1. Run a backtest with `BarSeriesManager` and use `amount=10` and assign result to variable `tradingRecordTen`
2. Run a new backtest with `BarSeriesManager` and use  `amount=1` and assign result to variable `tradingRecordOne`
3. Run `new VersusEnterAndHoldCriterion(new ProfitLossCriterion()).calculate(barSeries, tradingRecordTen)`  and compare result when you run `new VersusEnterAndHoldCriterion(new ProfitLossCriterion()).calculate(barSeries, tradingRecordOne)`.
4. Be amazed that the higher quantity used, the better you are than Buy and Hold

**Expected behavior**
Expect that `VersusEnterAndHoldCriterion` comparison should give same result no matter what `amount` is used per trade, since I expect enter and hold always uses same `amount` as its compared to.

PR #1193, opened 2024-10-11
EnterAndHoldCriterion: add optional amount

nimo23:
Fixes https://github.com/ta4j/ta4j/issues/1132

- Added constructor with `amount` for  `EnterAndHoldCriterion`
- Added constructor with `amount` for  `VersusEnterAndHoldCriterion`

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/criteria/EnterAndHoldCriterion.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/criteria/EnterAndHoldCriterion.java
@@ -23,6 +23,8 @@
  */
 package org.ta4j.core.criteria;
 
+import java.math.BigDecimal;
+
 import org.ta4j.core.AnalysisCriterion;
 import org.ta4j.core.BarSeries;
 import org.ta4j.core.BaseTradingRecord;
@@ -53,6 +55,7 @@ import org.ta4j.core.num.Num;
  */
 public class EnterAndHoldCriterion extends AbstractAnalysisCriterion {
 
+    private final BigDecimal amount;
     private final TradeType tradeType;
     private final AnalysisCriterion criterion;
 
@@ -70,6 +73,7 @@ public class EnterAndHoldCriterion extends AbstractAnalysisCriterion {
      *                                  {@code VersusEnterAndHoldCriterion}
      */
     public EnterAndHoldCriterion(AnalysisCriterion criterion) {
+        this.amount = null;
         this.tradeType = TradeType.BUY;
         this.criterion = criterion;
     }
@@ -84,6 +88,21 @@ public class EnterAndHoldCriterion extends AbstractAnalysisCriterion {
      *                                  {@code VersusEnterAndHoldCriterion}
      */
     public EnterAndHoldCriterion(TradeType tradeType, AnalysisCriterion criterion) {
+        this(tradeType, criterion, null);
+    }
+
+    /**
+     * Constructor.
+     *
+     * @param tradeType the {@link TradeType} used to open the position
+     * @param criterion the {@link AnalysisCriterion criterion} to calculate
+     * @param amount    the amount to be used to hold the entry position; if
+     *                  {@code null} then {@code 1} is used.
+     * @throws IllegalArgumentException if {@code criterion} is an instance of
+     *                                  {@code EnterAndHoldCriterion} or
+     *                                  {@code VersusEnterAndHoldCriterion}
+     */
+    public EnterAndHoldCriterion(TradeType tradeType, AnalysisCriterion criterion, BigDecimal amount) {
         if (criterion instanceof EnterAndHoldCriterion) {
             throw new IllegalArgumentException("Criterion cannot be an instance of EnterAndHoldCriterion.");
         }
@@ -92,12 +111,13 @@ public class EnterAndHoldCriterion extends AbstractAnalysisCriterion {
         }
         this.tradeType = tradeType;
         this.criterion = criterion;
+        this.amount = amount;
     }
 
     @Override
     public Num calculate(BarSeries series, Position position) {
-        int beginIndex = position.getEntry().getIndex();
-        int endIndex = series.getEndIndex();
+        var beginIndex = position.getEntry().getIndex();
+        var endIndex = series.getEndIndex();
         return criterion.calculate(series, createEnterAndHoldTrade(series, beginIndex, endIndex));
     }
 
@@ -106,8 +126,8 @@ public class EnterAndHoldCriterion extends AbstractAnalysisCriterion {
         if (series.isEmpty()) {
             return series.numFactory().one();
         }
-        int beginIndex = tradingRecord.getStartIndex(series);
-        int endIndex = tradingRecord.getEndIndex(series);
+        var beginIndex = tradingRecord.getStartIndex(series);
+        var endIndex = tradingRecord.getEndIndex(series);
         return criterion.calculate(series, createEnterAndHoldTradingRecord(series, beginIndex, endIndex));
     }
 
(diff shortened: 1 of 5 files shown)


ta4j-core/src/main/java/org/ta4j/core/criteria/EnterAndHoldCriterion.java
```diff
@@ -117,16 +137,18 @@ public boolean betterThan(Num criterionValue1, Num criterionValue2) {
     }
 
     private Position createEnterAndHoldTrade(BarSeries series, int beginIndex, int endIndex) {
-        Position position = new Position(tradeType);
-        position.operate(beginIndex, series.getBar(beginIndex).getClosePrice(), series.numFactory().one());
-        position.operate(endIndex, series.getBar(endIndex).getClosePrice(), series.numFactory().one());
+        var position = new Position(tradeType);
+        var entryAmount = amount == null ? series.numFactory().one() : series.numFactory().numOf(amount);
```
TheCookieLab: Please move this null check and default assignment up to the constructor.

ta4j-core/src/main/java/org/ta4j/core/criteria/EnterAndHoldCriterion.java
```diff
@@ -117,16 +137,18 @@ public boolean betterThan(Num criterionValue1, Num criterionValue2) {
...
+        position.operate(endIndex, series.getBar(endIndex).getClosePrice(), entryAmount);
         return position;
     }
 
     private TradingRecord createEnterAndHoldTradingRecord(BarSeries series, int beginIndex, int endIndex) {
-        TradingRecord fakeRecord = new BaseTradingRecord(tradeType);
-        fakeRecord.enter(beginIndex, series.getBar(beginIndex).getClosePrice(), series.numFactory().one());
-        fakeRecord.exit(endIndex, series.getBar(endIndex).getClosePrice(), series.numFactory().one());
+        var fakeRecord = new BaseTradingRecord(tradeType);
+        var entryAmount = amount == null ? series.numFactory().one() : series.numFactory().numOf(amount);
```
TheCookieLab: Same as above.

ta4j-core/src/test/java/org/ta4j/core/criteria/EnterAndHoldCriterionTest.java
```diff
@@ -47,116 +46,116 @@ public EnterAndHoldCriterionTest(NumFactory numFactory) {
...
+        var buyAndHoldPnlPercentage = getCriterion(new ProfitLossPercentageCriterion());
+        assertNumEquals(5, buyAndHoldPnlPercentage.calculate(series, position));
 
-        // sell and hold of ProfitLossPercentageCriterion
-        AnalysisCriterion sellAndHoldPnlPercentage = getCriterion(TradeType.SELL, new ProfitLossPercentageCriterion());
-        assertNumEquals(-5, sellAndHoldPnlPercentage.calculate(series, tradingRecord));
+        // sell and hold of PnlPercentageCriterion
+        var sellAndHoldPnlPercentage = getCriterion(TradeType.SELL, new ProfitLossPercentageCriterion());
+        assertNumEquals(-5, sellAndHoldPnlPercentage.calculate(series, position));
     }
```
TheCookieLab, line 72: There should be additional unit tests that exercise the new `amount` functionality. Among them should be the `null` case and the explicitly set `amount` case.

TheCookieLab (changes requested):

ta4j-core/src/main/java/org/ta4j/core/criteria/EnterAndHoldCriterion.java
nimo23: We cannot put this check in the constructor as we need a `series` to derive its `numFactory`.  Or?

TheCookieLab: Just the `amount`.

nimo23: Yes, and amount is **optional**. Or what you mean? Please provide code how to assign custom num within constructor.

nimo23: >Just the amount

So you mean we should provide another class instance variable like `final boolean hasAmount`? What is the benefit?

```
// adds superfluous intermediate class variable `hasAmount`
var entryAmount = hasAmount ? series.numFactory().one() : series.numFactory().numOf(amount);
```

I can do that, but what is the benefit???

Or do you mean something else. Please be a little more specific in the future (with code details) so that I don't have to think about what you mean several times.

nimo23: Ok, I have another idea.

nimo23: I have added `BigDecimal.ONE` for all default constructors and provide this information as a Javadoc. So there is no need for a check anymore.

ta4j-core/src/main/java/org/ta4j/core/criteria/EnterAndHoldCriterion.java
```diff
@@ -53,29 +56,35 @@
...
      *
      * @param criterion the {@link AnalysisCriterion criterion} to calculate
      * @throws IllegalArgumentException if {@code criterion} is an instance of
      *                                  {@code EnterAndHoldCriterion} or
      *                                  {@code VersusEnterAndHoldCriterion}
      */
     public EnterAndHoldCriterion(AnalysisCriterion criterion) {
+        this.amount = BigDecimal.ONE;
         this.tradeType = TradeType.BUY;
         this.criterion = criterion;
```
TheCookieLab:
```suggestion
        this(TradeType.BUY, criterion);
```

TheCookieLab: My bad, I missed this one earlier. Convenience constructors should funnel into the "main" constructor rather than setting properties directly. This both avoids duplication and protects against possible future divergence. In this case there's additional input validation on the `criterion` param that's in the other constructor but not this. This leads to unwanted inconsistent behavior between the constructors.

TheCookieLab (changes requested):

ta4j-core/src/main/java/org/ta4j/core/criteria/EnterAndHoldCriterion.java
nimo23: > This both avoids duplication and protects against possible future divergence.

I'm trying to understand what you really mean: there are a lot of constructors in ta4j, which also does not avoid duplication by adding such convenient constructors like this one. But I'm not against it, to avoid duplication is good! So what are you specifically suggesting? Is it better to delete all constructors and leave just one with all parameter possibilities?

TheCookieLab: Re-read the sentence before what you quoted.

nimo23: Please be precise and clarify what exactly you mean. If you have language problems, express your opinion by providing explicit code. I don't have time to solve language puzzles. Be precise, give code. Please answer my question or provide detail code and don't distract.

TheCookieLab: The exact, specific change I'm asking for is quite literally right there in the review. Here's a screen-shot of it if you (somehow) missed it.

Screenshot 2024-10-30 152307

If you need me to be more precise than this then I'm afraid I'm not able to help you.

nimo23: Yes, I missed it. Ok. But it was ambiguous for me..the code and your statement..

TheCookieLab (approved):

TheCookieLab:
> Ok, I think it's good now. The only thing what could be improved is to use
> 
> ```
> // use the cache, but with ternary op
> var entryAmount = amount.signum()== 1? series.numFactory().one() : series.numFactory().numOf(amount);
> ```
> 
> instead of
> 
> ```
> // don't use the cache, but without ternary op
> var entryAmount = series.numFactory().numOf(amount);
> ```
> 
> @TheCookieLab What do you think? What's better?

Unless we have a compelling and concrete justification for the former, the latter's simplicity is preferred.

Issue #802, opened 2021-09-19
[BUG] error in calculation of ReturnOverMaxDrawdownCriterion


dev590t:
**I have checked existing issues and wiki**
- [X] I could not find similar [issues](https://github.com/ta4j/ta4j/issues?utf8=%E2%9C%93&q=)
- [X] I could not find a solution in the [wiki](https://ta4j.github.io/ta4j-wiki/) or [faq section](https://ta4j.github.io/ta4j-wiki/FAQ.html)

**Describe the bug**
I have a strategy A that make 11.13866501343252% return, and 0.5954536310601489% Max. Drawdown
ReturnOverMaxDrawdownCriterion give
````
dev_1-bin@ 1.1113866501343252 / 0.005954536310601489 
res0: Double = 186.64537289924766
````
I have a strategy B that make 0.41164363422523% return, and 0.52925610114672155% Max. Drawdown
ReturnOverMaxDrawdownCriterion give
````
dev_1-bin@ 1.0041164363422523 / 0.0052925610114672155 
res1: Double = 189.72222222222223
````

It is evident strategy A is much better thant strategy B
````
// strategy A
dev_1-bin@ (1.1113866501343252 - 1) / 0.005954536310601489 
res2: Double = 18.706183710058472
//strategy B
dev_1-bin@ (1.0041164363422523 - 1) / 0.0052925610114672155 
res3: Double = 0.7777777777777891
````
But ReturnOverMaxDrawdownCriterion indicates strategy B is a little better strategy A
ReturnOverMaxDrawdownCriterion shouldn't use GrossReturnCriterion in its calculation
https://github.com/ta4j/ta4j/blob/369cadd448df3935745f96c5e5c0b712ad0a96c6/ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/ReturnOverMaxDrawdownCriterion.java#L40

**Comment by dev590t** [comment]
In https://github.com/ta4j/ta4j/blob/369cadd448df3935745f96c5e5c0b712ad0a96c6/ta4j-core/src/main/java/org/ta4j/core/analysis/criteria/ReturnOverMaxDrawdownCriterion.java#L60
I think we should replace by
```
            final Num totalProfit = grossReturnCriterion.calculate(series, tradingRecord) - 1;
```

dev590t:
I think a better solution could be add a new indicator GrossReturnPercentCriterion such as
`GrossReturnPercentCriterion = GrossReturnCriterion - 1`

[...]

PR #1194, opened 2024-10-11
fix ReturnCriterion and ReturnOverMaxDrawdown with base

nimo23:
Fixes https://github.com/ta4j/ta4j/issues/802

### Fixed
- Consider `base` when calculating `ReturnCriterion` of a position

### Added
- Added `base` to `ReturnOverMaxDrawdown` to include or exclude the base from the calculation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - Extracted NumFactory as source of numbers with defined precision
 
 ### Fixed
-
+- Consider `base` when calculating `ReturnCriterion` of a position
 
 ### Changed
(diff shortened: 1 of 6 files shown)


TheCookieLab (changes requested):
Thanks for your efforts. After giving this some thought, I prefer we go another route. In a nutshell we need to go back to fundamentals, simplify, and explicitly align with prevailing industry nomenclature. 

If one gains $10 on $100 invested, in financial parlance the **return** is commonly understood to be 10%. It is very unusual to define the return as 110% (or 1.1) so by doing so we are significantly out of sync with the "common tongue".

Let's remove the base entirely from `ReturnCriterion`. If we were writing this class from scratch, would we include an option to include the base in the calculation? I think the answer is clearly no, and the only reason we currently have it is an allowance for the original, idiosyncratic implementation. 

If we must provide a bridge to the past, instead of a boolean constructor toggle that changes class behavior (SRP) let's create a new class that does include the base (ex. `GrossReturnPercentOnInitialInvestmentCriterion`).

TheCookieLab:
For those criteria that "require" the base, convert them to use the new GrossReturnPercentOnInitialInvestmentCriterion.

TheCookieLab:
No worries, appreciate your help up to this point.

Issue #1073, opened 2023-07-25
[BUG] VersusEnterAndHoldCriterion misleading for "+/-"-criteria


nimo23:
**Describe the bug**
Some criteria values can have positive and negative values (e.g. PnlCriterion): For such criteria, the `VersusEnterAndHoldCriterion` compares `X` with its `EnterAndHold`, but it does not handle the sign (+/-) in favour of `X`. 

**1. Example: Given:**
myPnl +4%
enterAndHoldPnl: - 4%
=> actual calculation = myPnl / enterAndHoldPnl = 4 / -4 = -1 (it should be +1, because -1 would mean `X `is not better than its `EnterAndHold`)

**2. Example: Given:**
myPnl -4%
enterAndHoldPnl: +4%
=> actual calculation = myPnl / enterAndHoldPnl = -4 / +4 = -1 (correct, because -1 means `X `is not better than its `EnterAndHold`)

**3. Example: Given:**
myPnl -8%
enterAndHoldPnl: -4%
=> actual calculation = myPnl / enterAndHoldPnl = -8 / -4 = 0.5 (ok?)

**Expected behavior**
If a criterion can have both positive and negative values, then `VersusEnterAndHoldCriterion` should handle it in a special way so that the result of the `VersusEnterAndHoldCriterion` can be interpreted unequivocally. 

Maybe we could use another comparision method (instead of `x.dividedBy(enterAndHold)`, we could use "euclidian distance" (but then we have no sign) or any other representative distance calculation, maybe from package `criterion/statistics`).

phong-phuong:
Correct calculation is: (Strategy Account Balance)/(Buy and Hold Account Balance) - 100%.
In your first example, it would look like:
1.04/0.96 - 1.0 = 1.08333333 - 1.0 = 0.08333333, i.e. The strategy when compared to Enter and Hold, outperforms it by 8.33%.

phong-phuong:
Not quite, the formula will depend on what is being compared, and whether "returns" come into the picture.

Consider a portfolio with a start capital of $100.
Strategy A ends with or `returns` $104 back to the user at the end of a backtest.
Strategy B ends with or `returns` $96 back to the user at the end of a backtest.

In this context of returns, the `return` for Strategy A is 104%, and return for Strategy B is 96% which is what ta4j uses.

In Absolute values:
Strategy A Profit: $4
Strategy B Profit: -$4

In percentages:
Strategy A Profit: 4% or 0.04
[...]

PR #1195, opened 2024-10-11
fix VersusEnterAndHoldCriterion

nimo23:
Fixes https://github.com/ta4j/ta4j/issues/1073

- Consider `base` when calculating `ReturnCriterion` of a position.
- Fixed calculation for `VersusEnterAndHoldCriterion`

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/criteria/VersusEnterAndHoldCriterion.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/criteria/VersusEnterAndHoldCriterion.java
@@ -37,7 +37,23 @@ import org.ta4j.core.num.Num;
  *
  * <p>
  * Compares the value of a provided {@link AnalysisCriterion criterion} with the
- * value of an {@link EnterAndHoldCriterion}.
+ * value of an {@link EnterAndHoldCriterion} by using the following formula:
+ *
+ * <pre>
+ * xVersusEnterAndHold = (x - enterAndHold) / (abs(enterAndHold)
+ * </pre>
+ *
+ * <p>
+ * It can be used for absolute or percentage criteria values, for example:
+ *
+ * <ul>
+ * <lI>ProfitVsEnterAndHold = (400$ - 500$) / abs(500$) = -0.2 (i.e., the
+ * trading profit is 20% worse than the profit that would be made by simply
+ * entering and holding)
+ * <li>ProfitLossPercentageVsEnterAndHold = (0.9 - (-0.3)) / abs(-0.3) = +4
+ * (i.e., the trading profit is 400% better than the profit that would be made
+ * by simply entering and holding)
+ * </ul>
  */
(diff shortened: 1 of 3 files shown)


ta4j-core/src/main/java/org/ta4j/core/criteria/VersusEnterAndHoldCriterion.java
```diff
@@ -35,7 +35,23 @@
...
+ * </pre>
+ *
+ * <p>
+ * It can be used for absolute or percentage criteria values, for example:
+ *
+ * <ul>
+ * <lI>ProfitVsEnterAndHold = (400$ - 500$) / abs(500$) = -0.2 (= the trading
+ * profit is 20% worse than the profit that would be made by simply entering and
+ * holding)
+ * <li>ProfitLossPercentageVsEnterAndHold = (0.9 - (-0.3)) / abs(-0.3) = +4 (=
```
TheCookieLab: Why is 20% less than enter-and-hold represented by -0.2? Shouldn't it be 0.8? 

This is exactly why the same class producing semantically different values via a `addBase` flag introduces needless complexity and defect potential. That the examples in this javadoc mix them up says it all.

Any solution needs to include splitting the `ReturnCriterion` into separate classes. If you are unable or unwilling to do it then thanks for your efforts thus far, I'll add it to my TODO list.

TheCookieLab (changes requested):

ta4j-core/src/main/java/org/ta4j/core/criteria/VersusEnterAndHoldCriterion.java
nimo23: >Why is 20% less than enter-and-hold represented by -0.2? Shouldn't it be 0.8?

No, it is 20 less, because x is just 20% worse than enter and hold. You only care about the `signed difference of x and enterAndHold`. If both are not different then 0 is returned.

By the way, if you were to say "it's 0.8" then firstly that doesn't tell you how much worse oder better x is and secondly, most importantly, it would simply lead to incorrect calculations (even without a `addBase`) because It would ignore `+` and `–` and give an incorrect result for many criteria. Moreover, this fix is ​​stable to `addBase` as it would give the same result whether base is used or not. 

>This is exactly why the same class producing semantically different values via a addBase flag introduces needless complexity and defect potential. That the examples in this javadoc mix them up says it all.

Yes, I am with you. But actually we need to solve the current issue and therefore, the `base` should stay at the moment - no need for breaking changes here. My PR is only about bug fixing. Please re-read https://github.com/ta4j/ta4j/pull/1197#issuecomment-2426962927. 

> If you are unable or unwilling to do it 

No, I am not unwilling to do it. I also want to do it and create another class. But after this bug fix. We need a clean branch and improve that. Don't mess things up. Re-read https://github.com/ta4j/ta4j/pull/1197#issuecomment-2426962927.

Btw, you should be more aware of the consequences of removing the base for other classes or code bases..

Issue #1029, opened 2023-07-13
FixedDecimalIndicator must be renamed to FixedNumIndicator


nimo23:
We should rename `FixedDecimalIndicator` to `FixedNumIndicator`.

Reason:
In ta4j, the name `Decimal` was the previous name of `Num` and the renaming for this indicator was forgotten at that time. The name is also irritating because we have `DecimaNum` but this indicator can also be used with `DoubleNum`. I will do PR.

PR #1196, opened 2024-10-11
renamed-FixedDecimalIndicator-to-FixedNumIndicator

nimo23:
Fixes https://github.com/ta4j/ta4j/issues/1029

- Renamed `FixedDecimalIndicator` with `FixedNumIndicator`


TheCookieLab (approved):
While I agree with this renaming, going forward I'm going to be a bit more cautious about approving breaking changes that aren't solving an immediate and specific problem. We'll need to be a bit more judicious about when/why we force users to update their usages.

Issue #1141, opened 2024-04-05
Improve stability propagation


sgflt:
**Is your feature request related to a problem? Please describe.**
```BaseStrategy``` should not be directly aware about stability. Stability is a function of used rules and indicators.
As user I have no clue about possible instabilities of used indicators and rules, therefore instability of strategy should be calculated at runtime from used rules.

**Describe the solution you'd like**
Drop unstableBars field from BaseStrategy and introduce isStableAt(int index) method to Rule.

F.E.:

```java
public class DownTrendRule extends AbstractRule {

  private static final int UNSTABLE_BARS = 5;

  private final ADXIndicator directionStrengthIndicator;
  private final Rule downtrendRule;


  public DownTrendRule(final BarSeries series) {
    this.directionStrengthIndicator = new ADXIndicator(series, UNSTABLE_BARS);

    final var plusDMIndicator = new PlusDMIndicator(series);
    final var minusDMIndicator = new MinusDMIndicator(series);
    this.downtrendRule = new UnderIndicatorRule(plusDMIndicator, minusDMIndicator);
  }


  @Override
  public boolean isSatisfied(final int index) {
    return this.directionStrengthIndicator.getValue(index).isGreaterThan(numOf(25))
           && this.downtrendRule.isSatisfied(index)
  }
}
```

There is no way how to propagate stability to strategy.

sgflt:
Related question #1142

sgflt:
If I consider folloowing implementation:

```java
public class DownTrendIndicator extends AbstractIndicator<Boolean> {

  private static final int UNSTABLE_BARS = 5;

  private final ADXIndicator directionStrengthIndicator;
  private final Rule downtrendRule;


  public DownTrendIndicator(final BarSeries series) {
    super(series);
    this.directionStrengthIndicator = new ADXIndicator(series, UNSTABLE_BARS);

    final var plusDMIndicator = new PlusDMIndicator(series);
    final var minusDMIndicator = new MinusDMIndicator(series);
    this.downtrendRule = new UnderIndicatorRule(plusDMIndicator, minusDMIndicator);
  }



  @Override
  public Boolean getValue(final int index) {
    return  this.directionStrengthIndicator.getValue(index).isGreaterThan(numOf(25))
            && this.downtrendRule.isSatisfied(index);
  }


  @Override
  public int getUnstableBars() {
    return UNSTABLE_BARS;
  }
}
```

[...]

Issue #947, opened 2023-04-04
Return `NaN` for Indicators while within unstableBars?


nimo23:
Besides `Strategy#unstableBars`, we now also have `Indicator#getUnstableBars` (according to https://github.com/ta4j/ta4j/issues/919).

As @team172011 suggested:

> It is also necessary to determine the behavior of the Indicator.calculates(index) function regarding this issue. At the moment it is inconsistent. If it is not possible to calculate a value (e.g. because of index < size of timeFrame) some indicators return NaN, others zero while others modify timeFrame (make it smaller). I would suggest to return NaN whenever it is not possible to calculate the correct value

I think, this is a good idea. I would also suggest to return `NaN` whenever it is not possible to calculate **the correct value** of an indicator.

Any suggestions or objections on this topic?

_Related issues:_
- https://github.com/ta4j/ta4j/issues/101#issuecomment-346143435
- https://github.com/ta4j/ta4j/issues/101#issue-275653040
- https://github.com/ta4j/ta4j/issues/526#issue-574805289

TheCookieLab:
Agreed, having this as consistent behavior would be ideal. 

How do you see us implementing this - just returning `NaN` directly or should we use the `UnstableIndicator` to do so? I personally am partial to the simplicity of the former approach but am open to arguments for the latter.

Tangentially, we may also want to revisit how `NaN` behaves when compared to other Num values (ex see NumTest.testNaN). Right now it can be unstable when used in a sorting use case.

team172011:
But when returning NaN the user does not have any change using or inspecting values for indices lower than the unstable bars value. Maybe there are cases where the user wants to use or at least the those values

hhashim1:
Users should never be presented with values during warm-up period as it does not provide any value to user and only creates confusion and bad decisions.

hhashim1:
There should not be a wrapper around any indicator to for the purpose of excluding/suppressing data during warm-up period. This should be done by default by every indicator. Any major trading system and other technical analysis systems automatically suppress the warm up data.

TheCookieLab:
@hhashim1 are able/willing to take this on and create a PR?

team172011:
So following @nimo23 and @hhashim1 we should return NaN during warm up period? 

[...]

PR #1201, opened 2024-10-22
Add AverageIndicator

Cypher01:
Add `AverageIndicator`, coming from my spin-off project [ta4j-extended](https://github.com/Cypher01/ta4j-extended)


sgflt:
Isn't it just SMAIndicator?

sgflt:
Ah I see. Average of current state of multiple indicators.

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/AverageIndicator.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+
+/**
+ * Indicator to calculate the average of a list of other indicators.
+ */
+public class AverageIndicator extends CachedIndicator<Num> {
+    private final List<Indicator<Num>> indicators;
+
+    public AverageIndicator(Indicator<Num>... indicators) {
+        super(indicators[0]);
```
TheCookieLab: Please perform some input validation, at minimum the null and empty list cases.

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/AverageIndicator.java
```diff
@@ -0,0 +1,60 @@
...
+
+        this.indicators = Arrays.asList(indicators);
+    }
+
+    @Override
+    protected Num calculate(int index) {
+        Num value = getBarSeries().numFactory().zero();
+
+        for (Indicator<Num> indicator : indicators) {
+            value = value.plus(indicator.getValue(index));
```
TheCookieLab, line 72: What if the user instantiates an `AverageIndicator` with a list of indicators that happens to contain nulls?

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/AverageIndicator.java
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        return value.dividedBy(getBarSeries().numFactory().numOf(indicators.size()));
+    }
+
+    @Override
+    public int getUnstableBars() {
```
TheCookieLab: `calculate` should invoke this function early on to check the current `index` against the unstable bars value. If within the unstable period we should return `NaN`, otherwise proceed normally.

ta4j-core/src/test/java/org/ta4j/core/indicators/helpers/AverageIndicatorTest.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Test
+    public void indicatorShouldRetrieveAverageOfOneAndTwoAndThree() {
+        AverageIndicator averageIndicator = new AverageIndicator(one, two, three);
+
+        for (int i = 1; i < 10; i++) {
+            assertNumEquals(2, averageIndicator.getValue(i));
+        }
+    }
```
TheCookieLab, line 82: Please add tests to exercise the unstable bars scenarios.

TheCookieLab (changes requested):

TheCookieLab:
Thank you for contributing.

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/AverageIndicator.java
Cypher01: Are you sure? Unstable bars sounds like the result is potentially not accurate, but not that it should be `NaN`. If I compare this with e.g. `SMAIndicator`, this one can also calculate the moving average from `index` 0. If I would feed two `SMAIndicator`'s of e.g. `barCount` 10 and 20 into `AverageIndicator`, it could calculate the average from `index` 0, but the value would not be accurate until `index` 20.

TheCookieLab: Why would one want to execute trades (particularly use cases where there's real money on the line) on potentially inaccurate signals? If the end user application is simply analyzing for "fun" or other low/no stakes scenarios then perhaps false positives/negatives are no big deal, but it's a very different story when real money is involved. Since there's no distinction between "play/test" and "serious/production" indicators, all indicators must be the latter by default.

The [emerging convention](https://github.com/ta4j/ta4j/issues/947) is for indicators to return `NaN` within their unstable period to clearly delineate between values that can be trusted (and tradeable) from those that are not. There's already indicators that do this so your `calculate` addition loop logic has to handle this anyway (should be a test case).

Bottom line is you'll need to do _something_ to handle the unstable period scenario. It doesn't necessarily have to be `NaN`, but that is the prevailing convention. Whatever it is, it must unambiguously preserve signal integrity.

Cypher01: The conversation is interesting. As you pointed out, "all indicators must be the latter by default". It's at least not consistent yet. And as I already pointed out, the term unstable bars suggests it is unstable, not that there is no data available.

But OK, I'll implement the changes.

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/AverageIndicator.java
Cypher01: How should this look like? `super()` and `this()` must be first statement in a constructor. This implementation would anyway through an `ArrayIndexOutOfBoundsException`.

Either way, I implemented another constructor accepting a list of indicators and calling `super()` with `getFirst()`. But the situation is the same, just in this other case it will through a `NoSuchElementException`.

A manual validation would do the same, just the runtime exception would be a different one.
```
if (indicators == null || indicators.isEmpty()) {
    throw new IllegalArgumentException("...");
}
```

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/AverageIndicator.java
sgflt: _is for indicators to return NaN within their unstable period to clearly delineate between values that can be trusted (and tradeable) from those that are not_ 

So do we need **_isStable_** method if instability is represented by NaN? https://github.com/ta4j/ta4j/issues/1141

TheCookieLab: Yes, converging on a stability identification and propagation solution will be a high priority item on the road map. I like the `isStable` approach right now but am open to alternatives.

TheCookieLab (approved):

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/AverageIndicator.java
Cypher01: You obviously found a nice solution, thx! :+1:

TheCookieLab:
Alright barring any additional comments I'll be merging this shortly.

TheCookieLab (dismissed):

PR #1202, opened 2024-10-23
Add ChangeIndicator

Cypher01:
Add `ChangeIndicator`, coming from my spin-off project [ta4j-extended](https://github.com/Cypher01/ta4j-extended)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - added `Bar.getSystemZonedBeginTime`: the bar's begin time converted to system time zone
 - added `Bar.getSystemZonedEndTime`: the bar's end time converted to system time zone
 - added `BarSeries.getSeriesPeriodDescriptionInSystemTimeZone`: with times printed in system's default time zone
+- added `ChangeIndicator`
 
 ## 0.17 (released September 9, 2024)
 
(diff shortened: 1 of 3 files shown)


sgflt:
I slightly feel overlap between new BinaryOperation(Num::minus, left, right) + PreviousValueIndicator and this indicator.

sgflt:
Nope, I had created DifferenceIndicator too :D. This is ok.

Cypher01:
Good point, I just realized there is an even simpler implementation.

```java
CombineIndicator.minus(indicator, new PreviousValueIndicator(indicator, barCount))
```

This indicator could still be helpful for its convenient constructors, but the code/calculation is very simple. I'll update the implementation, but I would also be fine if this PR gets declined.

TheCookieLab (changes requested):
There is a lot of overlap in functionality with the existing `ClosePriceDifferenceIndicator`. Rather than a new class can fold these changes into the existing class by:

1. Allowing an arbitrary indicator (rather than the currently hard-coded `ClosePriceIndicator`)
2.  Introducing the `barCount` flexibility
3. Re-name accordingly

Cypher01:
OK, but before I need too many changes, is `ChangeIndicator` fine? Then I'll adapt and rename `ClosePriceDifferenceIndicator`.

TheCookieLab:
> OK, but before I need too many changes, is `ChangeIndicator` fine? Then I'll adapt and rename `ClosePriceDifferenceIndicator`.

Yes, everything else is fine.

TheCookieLab:
@Cypher01 are you still planning on making these changes or should I run with it?

PR #1203, opened 2024-10-23
Add BooleanCombineIndicator

Cypher01:
Add `BooleanCombineIndicator`, coming from my spin-off project [ta4j-extended](https://github.com/Cypher01/ta4j-extended)


sgflt:
This may have slightly different solution. I would wait after resolution of https://github.com/ta4j/ta4j/issues/1208

TheCookieLab:
> This may have slightly different solution. I would wait after resolution of #1208

Any changes in thought process here?

sgflt:
Umm... no. I have migrated a few indicators in event driven branch that prepares data in database and switched to python for AI  experiments.

I was probably referring to change in indicators hierarchy. Each `boolean` returning indicator extends from `BooleanIndicator` and each `Num` returning indicator extends from `NumericIndicator`.

PR #1204, opened 2024-10-23
Add SubstituteIndicator

Cypher01:
Add `SubstituteIndicator`, coming from my spin-off project [ta4j-extended](https://github.com/Cypher01/ta4j-extended)


TheCookieLab (changes requested):
It looks like this could also be a `TransformIndicator` candidate.

Cypher01:
True, same as for `NzIndicator`, I think I'll create a separate PR and close this one in the end.

TheCookieLab:
Please see https://github.com/ta4j/ta4j/pull/1252

PR #1205, opened 2024-10-23
Add NzIndicator

Cypher01:
Add `NzIndicator`, coming from my spin-off project [ta4j-extended](https://github.com/Cypher01/ta4j-extended)


TheCookieLab (changes requested):
I like the idea, however rather than a new indicator class I believe we can fit this into the existing `TransformIndicator` with something like:

`public static TransformIndicator nz(Indicator<Num> indicator, Num replacement) {
        return new TransformIndicator(indicator, val -> val.isNaN() ? replacement : val);
}`

Cypher01:
That is a good idea. I think I'll create a separate PR and close this one in the end.

TheCookieLab:
Please see https://github.com/ta4j/ta4j/pull/1252

PR #1206, opened 2024-10-23
Add KRIIndicator

Cypher01:
Add `KRIIndicator`, coming from my spin-off project [ta4j-extended](https://github.com/Cypher01/ta4j-extended)


Cypher01:
How should an indicator like this one be tested?

I tried to get some guidelines from tests like `MACDIndicatorTest`, but tests like these don't really add any value, because it cannot be validated where the hard-coded expected values come from.

Should I implement a test similar to `MACDIndicatorTest`?

TheCookieLab (dismissed):

TheCookieLab (approved):

PR #1214, opened 2024-11-08
 Use beginTime over endTime

### LINKED ISSUE #1213: Use beginTime over endTime
Bar Creation with End Time: I find the logic of creating bars with an EndTime and assigning the startTime in the bar constructor as endTime - duration a bit confusing. Typically, bar creation is done by specifying the bar's start time, which is also the standard across various platforms, market data providers, brokers, and exchanges. This approach may even constitute a bug, particularly for intraday time frames. Converting my data feed's start time to end time feels computationally inefficient and cumbersome. Currently, I add the duration to my datetime to correctly set the beginTime when creating a bar, which is an extra, unnecessary operation. How are others handling endTime?

The idea is to change the bars to favor begin time over end time. 

- Constructing the candles by passing the beginTime. 
- Print the candle using the beginTime. 
- Remove the endTime Instant from the bar to save memory and return beginTime + duration. 

let me know your comments and I can prepare a PR for this.

ek-ex:
Fixes #1213 

- Using beginTime instead of endTime in the Bar


ek-ex:
I was actually waiting for your confirmation on the last question. I'm bringing back the endTime in the next commit.

ek-ex:
Would you overload `BaseBarBuilder` to provide an option to pass either the `duration` or the `endTime`? If passing the `duration` it will create the Bar with `endTime = beginTime + duration`? BaseBar will still be constructed with `beginTime` and `endTime`.

ek-ex:
OK, lets decouple the `VolumeBarBuilder` changes as it would need changes to the `BarBuilder` interface. I can do those in a following PR.

ek-ex:
In this commit I've kept the `Bar` intact. I also did not changed `BaseBar` to compute the duration on the fly between `beginTime` and `endTime`. The bars are constructed using `beginTime` as opposed to `endTime` which is the main purpose of this PR which aligns with all major brokers and historical data providers.  `endTime` is computed in the `BaseBar` constructor and keeping all the same members in it. 

Please indicate if this is good to merge as the first iteration or where do you want to go from here.

ek-ex:
@nimo23 closed the PR. Its not clear to me that #982 and #983 suggest using beginTime for building the Bars, which was the intention of this PR.

TheCookieLab:
I can see the value of instantiating bars with begin time, but I see this as an addition to, rather than a replace. Users should have the option of creating bars with `beginTime `or `endTime `as there's nothing that says the functionality needs to be mutually exclusive. Given this is a library used by all types of traders, in general we should keep in mind that our use case may not be canonical or even representative of the wider community.

Issue #1209, opened 2024-10-29
Event sourced backtesting


sgflt:
As BFU I wanted to test some strategies. I had to grab BarSeries, code some data loader (CSV, JDBC, online API). After a few days I had some runnable code with copy pasted graphing.

But results are outcome for some static handmade set of parameters. Later I have implemented simulated annealing for parameter estimation.

Real exchange is somehow more complex. There are candle stream, tick stream, news stream, stream of profits for each position. Maybe other useful streams. I have added Apache Camel realtime candle aggregation to multiple timeframes.

And there is the point where I need something better. I need to push that time ordered streams into event sourced backtester that forwards apropriate candles to related time framed series. That forwards tick, profits and other stateful data to "trading context".

For this complexity backtesting is "iterate only for each bar and calculate indicators"  not sufficient, because it covers small portion of real problem.

I usually use Rules that fetch data through Supplier<> from external source and are not even related to series.

```java
  this.liveTrading = Map.of(
        "1m", new LiveTradingBuilder().withStrategyFactory(series -> buildStrategy(series, "1m")).build(),
        "5m", new LiveTradingBuilder().withStrategyFactory(series -> buildStrategy(series, "5m")).build(),
        "15m", new LiveTradingBuilder().withStrategyFactory(series -> buildStrategy(series, "15m")).build(),
        "30m", new LiveTradingBuilder().withStrategyFactory(series -> buildStrategy(series, "30m")).build(),
        "1h", new LiveTradingBuilder().withStrategyFactory(series -> buildStrategy(series, "1h")).build(),
        "4h", new LiveTradingBuilder().withStrategyFactory(series -> buildStrategy(series, "4h")).build()
    );

  private Strategy buildStrategy(final BarSeries series, final String resolution) {
    if (series == null) {
      throw new IllegalArgumentException("Series cannot be null");
    }


    return this.strategyFactoryFactory.createStrategy(
        series,
        this.symbol.symbol(),
        resolution,
        new RuntimeContext(
            this::getProfit,
            this::getMaxProfit,
            this::getTimeInTrade,
            this::getPriceDifference,
            this::getDifference,
            this.positions::size
        )
    );
  }

  public synchronized boolean shouldEnter() {
    return this.liveTrading.values().stream().allMatch(LiveTrading::shouldEnter);
  }

```

So I need event sourced backtesting.

1. Load time based events from data source.
2. Prepare chosen time frame series
3. Assign strategies to corresponding timeframes
4. Publish events
[...]

Issue #906, opened 2022-10-26
Eliminate class instance variables from indicators


nimo23:
We need to eliminate all usages of class instance variables from cachable indicators that store context information that depend on an index. 

This issue is here **to list all those indicators** and afterwards to create dedicated PR's to remove those class instance variables. If anyone knows a better solution, ideas are welcome.

See:
- https://github.com/ta4j/ta4j/issues/904#issuecomment-1292467096
- https://github.com/ta4j/ta4j/issues/902#issuecomment-1289344468

Maybe, we should also include a sentence in the ta4j documentation like:
_"When developing a cached indicator, then do **not** use class instance variable when calculating the indicator value, because ..."_

EDIT 1: Using this top commentary for summary of findings:
- [x] TrailingStopLossRule (currentStopLossLimitActivation)
- [x] ThreeBlackCrowsIndicator (whiteCandleIndex)
- [x] ThreeWhiteSoldiersIndicator (blackCandleIndex)
- [x] DifferencePercentageIndicator (lastNotification)
- [x] SimpleLinearRegressionIndicator (slope, intercept) (not a bug calculateRegressionLine get called in every iteration)
- [x] ParabolicSarIndicator

EDIT 2: Packages that have been reviewed:
- [x] _default_
- [x] utils
- [x] num
- [x] aggregator
- [x] reports
- [x] indicators
- [x] candles
- [x] bollinger
- [x] adx
- [x] aroon
- [x] helpers
- [x] rules
- [x] volume
- [x] statistics
- [x] pivotpoints
- [x] numeric
- [x] keltner
- [x] ichimoku
- [x] criteria
- [x] helpers
- [x] pnl
- [x] analysis
- [x] cost

team172011:
> I think, we can also find a few buggy indicators with the help of https://github.com/ta4j/ta4j/pull/907 by switching the cache off and looking for failed indicator tests.

I think this issue is not directly related to having a cache or not

team172011:
> After finding out the buggy indicators, we should also include dedicated test cases to ensure that no one adds those class instance variables afterwards in any future PR.

every `Num` or primitive as instance variable and that is not marked as `final` would be an candidate for our issue...

team172011:
I added the top commentar with EDIT 1: this is all I could find by going through all packages (EDIT 2)

team172011:
[...]

PR #1219, opened 2024-11-15
Bugfix/#1152 sample variance

### LINKED ISSUE #1152: [BUG] VarianceIndicator calculation seems to be wrong
- [ ] I could not find similar [issues](https://github.com/ta4j/ta4j/is
/tsues?utf8=%E2%9C%93&q=)

**Describe the bug**
A clear and concise description of what the bug is. If possible provide a **[minimal working example](https://stackoverflow.com/help/mcve)**!

Assuming LibreOffices VAR function on test data:

| close price |  VAR of 4 bars  | formula | test assertions |
|--------|--------|--------|--------|
| 1 | -- | -- | 0 |
| 2 | -- | -- | 0.25 |
| 3 | -- | -- | 2.0/3 |
| 4 | 1,66666666666667 | =VAR(A1:A4)  | 1.25 |
| 3 |  0,666666666666667 | =VAR(A2:A5)  | 0.5 | 
| 4 |  0,333333333333333 | =VAR(A3:A6) | 0.25 | 
| 5 | 0,666666666666667 | =VAR(A4:A7)  | 0.5 | 
| 4 | 0,666666666666667 | =VAR(A5:A8)  | 0.5 | 
| 3 | 0,666666666666667 | =VAR(A6:A9)  | 0.5 | 
| 0 | 4,66666666666667 |  =VAR(A7:A10) | 3.5 | 
| 9 | 14 | =VAR(A8:A11)  | 10.5 | 

**Expected behavior**
Test asserts should match LibreOffice calculation or explain assert calculation.

sgflt:
Fixes #1152.

- Variance sample calculation
- Welford's sttreaming algorithm

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicator.java
@@ -303,7 +303,7 @@ public class ConvergenceDivergenceIndicator extends CachedIndicator<Boolean> {
      * @return true, if positive convergent
      */
     private Boolean calculatePositiveConvergence(int index) {
-        CorrelationCoefficientIndicator cc = new CorrelationCoefficientIndicator(ref, other, barCount);
+        var cc = CorrelationCoefficientIndicator.ofSample(ref, other, barCount);
         boolean isConvergent = cc.getValue(index).isGreaterThanOrEqual(minStrength);
 
         Num slope = calculateSlopeRel(index);
@@ -317,7 +317,7 @@ public class ConvergenceDivergenceIndicator extends CachedIndicator<Boolean> {
      * @return true, if negative convergent
      */
     private Boolean calculateNegativeConvergence(int index) {
-        CorrelationCoefficientIndicator cc = new CorrelationCoefficientIndicator(ref, other, barCount);
+        var cc = CorrelationCoefficientIndicator.ofSample(ref, other, barCount);
         boolean isConvergent = cc.getValue(index).isGreaterThanOrEqual(minStrength);
 
         Num slope = calculateSlopeRel(index);
@@ -333,7 +333,7 @@ public class ConvergenceDivergenceIndicator extends CachedIndicator<Boolean> {
      */
     private Boolean calculatePositiveDivergence(int index) {
 
-        CorrelationCoefficientIndicator cc = new CorrelationCoefficientIndicator(ref, other, barCount);
+        var cc = CorrelationCoefficientIndicator.ofSample(ref, other, barCount);
         boolean isDivergent = cc.getValue(index)
                 .isLessThanOrEqual(minStrength.multipliedBy(getBarSeries().numFactory().minusOne()));
 
@@ -352,7 +352,7 @@ public class ConvergenceDivergenceIndicator extends CachedIndicator<Boolean> {
      */
     private Boolean calculateNegativeDivergence(int index) {
 
-        CorrelationCoefficientIndicator cc = new CorrelationCoefficientIndicator(ref, other, barCount);
+        var cc = CorrelationCoefficientIndicator.ofSample(ref, other, barCount);
         boolean isDivergent = cc.getValue(index)
                 .isLessThanOrEqual(minStrength.multipliedBy(getBarSeries().numFactory().numOf(-1)));
 
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/CorrelationCoefficientIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/CorrelationCoefficientIndicator.java
@@ -23,6 +23,8 @@
  */
 package org.ta4j.core.indicators.statistics;
 
+import java.util.function.BiFunction;
+
 import org.ta4j.core.Indicator;
 import org.ta4j.core.indicators.CachedIndicator;
 import org.ta4j.core.num.Num;
@@ -47,19 +49,33 @@ public class CorrelationCoefficientIndicator extends CachedIndicator<Num> {
      * @param indicator2 the second indicator
      * @param barCount   the time frame
      */
-    public CorrelationCoefficientIndicator(Indicator<Num> indicator1, Indicator<Num> indicator2, int barCount) {
+    private CorrelationCoefficientIndicator(final Indicator<Num> indicator1, final Indicator<Num> indicator2,
+            final int barCount, final boolean isSample) {
         super(indicator1);
-        this.variance1 = VarianceIndicator.ofSample(indicator1, barCount);
-        this.variance2 = VarianceIndicator.ofSample(indicator2, barCount);
+        final BiFunction<Indicator<Num>, Integer, VarianceIndicator> varianceProvider = isSample
+                ? VarianceIndicator::ofSample
+                : VarianceIndicator::ofPopulation;
+        this.variance1 = varianceProvider.apply(indicator1, barCount);
+        this.variance2 = varianceProvider.apply(indicator2, barCount);
         this.covariance = new CovarianceIndicator(indicator1, indicator2, barCount);
     }
 
+    public static CorrelationCoefficientIndicator ofSample(final Indicator<Num> indicator1,
+            final Indicator<Num> indicator2, final int barCount) {
+        return new CorrelationCoefficientIndicator(indicator1, indicator2, barCount, true);
+    }
+
+    public static CorrelationCoefficientIndicator ofPopulation(final Indicator<Num> indicator1,
+            final Indicator<Num> indicator2, final int barCount) {
+        return new CorrelationCoefficientIndicator(indicator1, indicator2, barCount, false);
+    }
+
     @Override
-    protected Num calculate(int index) {
-        Num cov = covariance.getValue(index);
-        Num var1 = variance1.getValue(index);
-        Num var2 = variance2.getValue(index);
-        Num multipliedSqrt = var1.multipliedBy(var2).sqrt();
+    protected Num calculate(final int index) {
+        final Num cov = this.covariance.getValue(index);
+        final Num var1 = this.variance1.getValue(index);
+        final Num var2 = this.variance2.getValue(index);
+        final Num multipliedSqrt = var1.multipliedBy(var2).sqrt();
         return cov.dividedBy(multipliedSqrt);
     }
 
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/CorrelationCoefficientIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/CorrelationCoefficientIndicator.java
@@ -49,8 +49,8 @@ public class CorrelationCoefficientIndicator extends CachedIndicator<Num> {
      */
     public CorrelationCoefficientIndicator(Indicator<Num> indicator1, Indicator<Num> indicator2, int barCount) {
         super(indicator1);
-        this.variance1 = new VarianceIndicator(indicator1, barCount);
-        this.variance2 = new VarianceIndicator(indicator2, barCount);
+        this.variance1 = VarianceIndicator.ofSample(indicator1, barCount);
+        this.variance2 = VarianceIndicator.ofSample(indicator2, barCount);
         this.covariance = new CovarianceIndicator(indicator1, indicator2, barCount);
     }
 
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/SigmaIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/SigmaIndicator.java
@@ -47,27 +47,36 @@ public class SigmaIndicator extends CachedIndicator<Num> {
      * @param ref      the indicator
      * @param barCount the time frame
      */
-    public SigmaIndicator(Indicator<Num> ref, int barCount) {
+    private SigmaIndicator(final Indicator<Num> ref, final int barCount, final boolean isSample) {
         super(ref);
         this.ref = ref;
         this.barCount = barCount;
         this.mean = new SMAIndicator(ref, barCount);
-        this.sd = new StandardDeviationIndicator(ref, barCount);
+        this.sd = isSample ? StandardDeviationIndicator.ofSample(ref, barCount)
+                : StandardDeviationIndicator.ofPopulation(ref, barCount);
+    }
+
+    public static SigmaIndicator ofSample(final Indicator<Num> ref, final int barCount) {
+        return new SigmaIndicator(ref, barCount, true);
+    }
+
+    public static SigmaIndicator ofPopulation(final Indicator<Num> ref, final int barCount) {
+        return new SigmaIndicator(ref, barCount, false);
     }
 
     @Override
-    protected Num calculate(int index) {
+    protected Num calculate(final int index) {
         // z-score = (ref - mean) / sd
-        return (ref.getValue(index).minus(mean.getValue(index))).dividedBy(sd.getValue(index));
+        return (this.ref.getValue(index).minus(this.mean.getValue(index))).dividedBy(this.sd.getValue(index));
     }
 
     @Override
     public int getUnstableBars() {
-        return barCount;
+        return this.barCount;
     }
 
     @Override
     public String toString() {
-        return getClass().getSimpleName() + " barCount: " + barCount;
+        return getClass().getSimpleName() + " barCount: " + this.barCount;
     }
 }
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/VarianceIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/VarianceIndicator.java
@@ -28,7 +28,7 @@ import org.ta4j.core.indicators.CachedIndicator;
 import org.ta4j.core.num.Num;
 import org.ta4j.core.num.NumFactory;
 
-public class VarianceIndicator extends CachedIndicator<Num> {
+public final class VarianceIndicator extends CachedIndicator<Num> {
 
     private final Indicator<Num> indicator;
     private final int barCount;
@@ -39,7 +39,7 @@ public class VarianceIndicator extends CachedIndicator<Num> {
     private Num previousSum;
     private Num previousSumOfSquares;
 
-    public VarianceIndicator(final Indicator<Num> indicator, final int barCount, final boolean isSampleVariance) {
+    private VarianceIndicator(final Indicator<Num> indicator, final int barCount, final boolean isSampleVariance) {
         super(indicator);
         this.indicator = indicator;
         this.barCount = Math.max(barCount, 1);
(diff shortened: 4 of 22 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/CorrelationCoefficientIndicator.java
```diff
@@ -47,19 +49,33 @@ public class CorrelationCoefficientIndicator extends CachedIndicator<Num> {
      * @param indicator2 the second indicator
      * @param barCount   the time frame
      */
-    public CorrelationCoefficientIndicator(Indicator<Num> indicator1, Indicator<Num> indicator2, int barCount) {
+    private CorrelationCoefficientIndicator(final Indicator<Num> indicator1, final Indicator<Num> indicator2,
```
TheCookieLab: Why make the constructor private? The new static functions are essentially convenience constructors and while convenience is good, it should not be mandatory. The `TransformIndicator` should be seen as the canonical template for this pattern and it maintains a public constructor despite it's stable of static helpers. Unless there is compelling reason to deviate, the same should apply here.

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/VarianceIndicator.java
```diff
@@ -25,53 +25,122 @@
...
-     *
-     * @param indicator the indicator
-     * @param barCount  the time frame
-     */
-    public VarianceIndicator(Indicator<Num> indicator, int barCount) {
+    private final boolean isSampleVariance;
+
+    // State for sequential calculation
+    private int previousIndex = -1;
+    private Num previousSum;
```
TheCookieLab: There was an [effort to remove class instance variables from cached indicators](https://github.com/ta4j/ta4j/issues/906) so this would be a regression. @nimo23 

Improving performance is needed but will require a global approach. If concurrency is going to play a large role (and I suspect it will) having this state in play will likely make things more difficult.

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/VarianceIndicator.java
```diff
@@ -25,53 +25,122 @@
...
         this.indicator = indicator;
-        this.barCount = barCount;
-        this.sma = new SMAIndicator(indicator, barCount);
+        this.barCount = Math.max(barCount, 1);
+        this.isSampleVariance = isSampleVariance;
+        this.previousSum = numFactory().zero();
+        this.previousSumOfSquares = numFactory().zero();
+    }
+
+    public static VarianceIndicator ofSample(final Indicator<Num> indicator, final int barCount) {
```
TheCookieLab, line 62: I am not convinced this is the implementation pattern we should adopt. I believe I have already voiced my aversion to both private constructors (above) and boolean constructor params (in other PRs and discussions) and this approach would ultimately contain both.

Boolean constructor params are simply less readable than many of the alternatives, and while IDEs that label params can help, we should not rely on them as a crutch for readable code. Brainstorming alternative approaches:

**Use an Enum instead of a boolean**: 
Instead of `new VarianceIndicator(indicator, barCount, true)` 
we'd have `new VarianceIndicator(indicator, barCount, VarianceType.POPULATION)`. 

Someone reading the code in vim or notepad could still reasonably follow along. The enum would be shared among all classes that currently use the boolean flag.

**Separate classes**: This can be implemented in several ways but the naive approach duplicates a lot of code. 

Since the only difference between calculating sample and population is the divisor used, an approach that minimizes code duplication is a "base" `VarianceIndicator` class that contains most of the logic but contains an overridable `getDivisor(int numberOfObservations)` function which it calls when performing the actual variance calculation. 

We then have two classes, `PopulationVarianceIndicator` and `SampleVarianceIndicator` which extend the base class and override the `getDivisor()` function appropriately. If we want to consider the sample variation as the default, `getDivisor()` could return `numberOfObservations - 1`as the default thus eliminating the need for the extra `SampleVarianceIndicator`. 

Other ideas also welcome. Thoughts?

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/CorrelationCoefficientIndicator.java
sgflt: Okay. I usually tend to reduce multiple paths that do the same thing to the minimum. From my perspective it reduces "maintenance cost" of troubleshooting in production as it reduces creativity of programmers.

From usage perspective i preffer narrow paths. 

_Can't use constructor? There is probably factory method._
_Can use constructor? I usually do not look for factories. But they may be more readable and I miss them._

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/VarianceIndicator.java
sgflt: What kind of concurrency are we talking about? 

It is more efficient to scale vertically than horizontally. Adding CPUs to slow algorithm makes it "maybe less slow" but "more expensive" if you pay CPU in the cloud.

I can imagine following scenarios:

- Divide and Conquer per indicator - this may have large overhead
- Divide and Conquer per strategy - already implemented in BacktestExecutor and it is compatible with fast path. (assuming each strategy have own indicator set, hmmm... do you think about sharing indicators between strategies?)

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/VarianceIndicator.java
sgflt: I simply agree. I have chosen enum for simplicity. Usually looking for variance and then I have to decide which variant I want to use.

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/VarianceIndicator.java
TheCookieLab: As you say evaluating multiple strategies concurrently already exists and offers significantly better performance on multi-core CPUs but evaluating a strategy on a very large bar series is still a pain point. 

Sharing indicators could be a possibly be a solution (or rather a component of the solution) however I will need to do some more prototyping before knowing for sure. Assuming it is however, having instance variables within the indicators would complicate things considerably.

Adding new fast path logic to the project is not necessarily forbidden there should be good reason for it (i.e. significant performance gains). We don't know when we'll implement the new performance/concurrency system so having the better performance in the interim may be worth the future tech debt. Do you have any stats on the gains here?

sgflt: I am still doing event driven thing. But stats are clear. Any drop from O(N^2) to O(N) or even in some cases to O(1) is huge for larger windows. RunningTotalndicator is N to 1 case.

Similar Highest and Lowest value indicators:
https://github.com/sgflt/ta4j/blob/52aeb0966207ad1b54141d13815cc2823be13f4c/ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/HighestValueIndicator.java#L65C2-L96C4

I didn't have time to prepare some relevant tests, but if you are interested into indicator sharing, you may look at this example:

https://github.com/sgflt/ta4j/blob/%231209-eventsourced-reactive-backtesting/ta4j-core/src/test/java/org/ta4j/core/backtest/BacktestExecutorTest.java

> having instance variables within the indicators would complicate things considerably

I think the opposite may be true. The state is well encapsualted and simpler to implement. And what I like it is testable. See comments: 
https://github.com/ta4j/ta4j/issues/1209#issuecomment-2482840515
https://github.com/ta4j/ta4j/issues/1209#issuecomment-2546608762

sgflt: You probably meant sharing of indicator between multiple strategies. Hm. It is tricky, because calculation is stateful.

I can imagine following approaches:

- Run Strategy per thread  with separate indicator contexts. It is easy to do, but each timepoint is calculated separately and this wastes CPU time.
- Run single threaded simulation of multiple strategies sharing single indicator context. This does not waste CPU, nor gains multithreading performance. It is small optimization before parallelization.
- **Two phase run. First run fills cache without strategy testing. Second run can execute thread per strategy (thread count = CPU count).** Currently I use only statefulness of each indicator without cache over each timepoint. Two phase approach may be the best for bar simulations as it prepares state without contention, and then all strategies just read prepared states. This behavior is similar to current `CachedIndicator`

IndicatorContext is constraint for threading as it is bound to certain time point.

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/Type.java
```diff
@@ -0,0 +1,32 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package org.ta4j.core.indicators.statistics;
+
+public enum Type {
```
nimo23: Please rename this enum to `VarianceType`. The name `Type` is too general and can also cause naming conflicts when importing.

ta4j-core/src/test/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicatorTest.java
```diff
@@ -105,25 +105,8 @@ public void setUp() {
...
-        testNegativeDivergentStrict();
-
-    }
-
     public void testPositiveConvergent() {
-        assertFalse(isPosCon.getValue(0));
-        assertFalse(isPosCon.getValue(1));
-        assertFalse(isPosCon.getValue(2));
-        assertTrue(isPosCon.getValue(3));
+        // Skip unstable bars (first barCount-1 bars)
```
nimo23: Please do not skip unstable bars for at least one test case to ensure that `NaN` is returned in the unstable state.

ta4j-core/src/test/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicatorTest.java
```diff
@@ -77,17 +76,18 @@ public void setUp() {
         otherPosDiv = new FixedNumIndicator(series, 20, 40, 70, 80, 90, 100, 200, 220, -50, 7);
 
         // for convergence and divergence
+        // Use lower minStrength to match the actual correlation values in the test data
         isPosCon = new ConvergenceDivergenceIndicator(refPosCon, otherPosCon, 3,
-                ConvergenceDivergenceType.positiveConvergent);
+                ConvergenceDivergenceType.positiveConvergent, 0.5, 0.3);
```
nimo23: Why are you changing the values from `0.8 and 0.3` (default) to ` 0.5 and 0.3`?

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/SigmaIndicator.java
```diff
@@ -46,28 +46,38 @@ public class SigmaIndicator extends CachedIndicator<Num> {
...
+    public static SigmaIndicator ofPopulation(final Indicator<Num> ref, final int barCount) {
+        return new SigmaIndicator(ref, barCount, Type.POPULATION);
     }
 
     @Override
-    protected Num calculate(int index) {
+    protected Num calculate(final int index) {
         // z-score = (ref - mean) / sd
-        return (ref.getValue(index).minus(mean.getValue(index))).dividedBy(sd.getValue(index));
+        return (this.ref.getValue(index).minus(this.mean.getValue(index))).dividedBy(this.sd.getValue(index));
```
nimo23: Why not just remove the keyword `this`, which is not required in most cases (see also other classes in this PR).

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/SigmaIndicator.java
```diff
@@ -46,28 +46,38 @@ public class SigmaIndicator extends CachedIndicator<Num> {
...
     }
 
     @Override
     public int getCountOfUnstableBars() {
-        return barCount;
+        return this.barCount;
```
nimo23: Why not just remove the keyword `this`, which is not required in most cases (see also other classes in this PR).

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/Type.java
sgflt: I think VarianceType is too specific, because Type is used also for StandardDeviation.

Would be `SampleType` better?

ta4j-core/src/test/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicatorTest.java
sgflt: It is caused by different `Type` Sample vs Population. Behavior should stay the same, and test asserts may stay at their previous positions with original data.

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/SigmaIndicator.java
sgflt: Is it enforceable with PMD plugin?

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/Type.java
nimo23: >Would be `SampleType` better?

Yes, please rename it to `SampleType`

ta4j-core/src/test/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicatorTest.java
sgflt: This is a boolean indicator so false should be returned?

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/SigmaIndicator.java
nimo23: >Is it enforceable with PMD plugin?

Why enforcable? By default, PMD as well as SonarQube does not enforce `this` in such cases, or? Explicit `this` makes the code unnecessarily bloated and reduces readability.

ta4j-core/src/test/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicatorTest.java
nimo23: >Behavior should stay the same, and test asserts may stay at their previous positions with original data.

Then please explicitly use 0.8 and 0.3.

ta4j-core/src/test/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicatorTest.java
nimo23: >This is a boolean indicator so false should be returned?

Yes.

ta4j-core/src/test/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicatorTest.java
sgflt: It would break true/false sequence or i would have to create a new source data. This is the least breaking change in test.

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/SigmaIndicator.java
sgflt: I always forgot to switch codestyles :). Yes or no. Consistency is what matters and our code base enforces usage of `this`. It would be good enough to set up PMD to enforce rules in ta4j so it will stay consistent without unnecessary round trips to code review.

ta4j-core/src/test/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicatorTest.java
```diff
@@ -77,17 +76,18 @@ public void setUp() {
         otherPosDiv = new FixedNumIndicator(series, 20, 40, 70, 80, 90, 100, 200, 220, -50, 7);
 
         // for convergence and divergence
+        // Use lower minStrength to match the actual correlation values in the test data
```
nimo23, line 79: >Use lower minStrength to match the actual correlation values in the test data

As you changed the `minStrength` from 0.8 (default) to 0.5, does this mean that the adjusted Variance formula results in higher sensitivity in `ConvergenceDivergenceIndicator`? If so, would it be useful to adjust the constructor default values ​​of `ConvergenceDivergenceIndicator`?

sgflt: It seems to be correct, the `(n-1)` divisor is more sensitive than `(n)`.

I have used binary search for determining the closest coefficient: 0.57735

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicator.java
```diff
@@ -143,7 +143,7 @@ public enum ConvergenceDivergenceStrictType {
      * divergence and must be a number between "0.1" and "1.0": <br/>
      * <br/>
      * 0.1: very weak <br/>
-     * 0.8: strong (recommended) <br/>
+     * 0.57735: strong (recommended) <br/>
```
nimo23: What if `CorrelationCoefficientIndicator.ofPopulation` is used? Do we then have a more consistent ratio (like before) without such deviations? Can you try which is better?

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/SampleType.java
```diff
@@ -0,0 +1,32 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package org.ta4j.core.indicators.statistics;
+
+public enum SampleType {
+    SAMPLE, POPULATION;
```
nimo23: Would be nice to include a short javadoc, for example:

```
public enum SampleType {

    /** A subset group under study. */
    SAMPLE, 

    /** The entire group under study. */
    POPULATION;  

}
```

ta4j-core/src/main/java/org/ta4j/core/indicators/helpers/ConvergenceDivergenceIndicator.java
sgflt: simplified_correlation_timeline_4k

I think we should revert recommendation to 0.8. With longer rolling window SAMPLE and POPULATION converges to the same line.

Window of 3 bars is really short and differences are much bigger.

nimo23: @sgflt Yes, please revert to `0.8`.

sgflt: reverted

Issue #1217, opened 2024-11-12
VolumeBarBuilder with beginTime and endTime


nimo23:
**Is your feature request related to a problem? Please describe.**

When we create a `VolumeBar` (from `BaseBar`) using `VolumeBarBuilder`, we add `beginTime` and `endTime` to the builder. The `duration` can vary between bars and should be calculated. The duration-invariant discussed here https://github.com/ta4j/ta4j/pull/983 is wrong for this type of bar.

**Describe the solution you'd like**
To conclude, the `VolumeBarBuilder` should be improved:

Instead of setting:

    private Duration timePeriod;
    private Instant endTime;

it should be:

    private Instant beginTime;
    private Instant endTime;

and its duration (= timePeriod) should be computed by the given `beginTime` and `endTime`.

ek-ex:
`VolumeBarBuilder` inherits from `BarBuilder`  which does not have a method to pass `beginTime`. Do you see a different interface used to `BarBuilder`?

ek-ex:
I was thinking about this problem and wanted to make a suggestion to use java generics. It would look like this: 

a) 
```public interface BarBuilder<T extends Bar> ```

b)
```public class BaseBarBuilder implements BarBuilder<BaseBar>```

c) 
```public class VolumeBarBuilder implements BarBuilder<VolumeBar> ```

and so on.. 

With this we would benefit that BaseBarSeries can now hold any implementation of Bar's and avoid any casting required in the future.

```public class BaseBarSeries<T extends Bar> implements BarSeries<T>```

PR #1222, opened 2024-11-16
improve BarBuilders

nimo23:
Fixes https://github.com/ta4j/ta4j/issues/1217

- improved bar builders (`TickBarBuilder`, `VolumeBarBuilder`) and fixed the calculation for `bar.duration`, `bar.endTime`, `bar.beginTime` when aggregating bars
- renamed `BaseBarBuilder` and `BaseBarBuilderFactory` to `TimeBarBuilder` and `TimeBarBuilderFactory` as this builder builds time bars (= bars aggregated by time period)
- moved `BaseBar` from `bars`-package to `root`-package as it is used by any type of bar (therefore the name `BaseBarBuilder` does not make sense; btw, other type of builders also create BaseBar's)
- improved javadoc for all bar builders and BaseBar
- simplified BaseBar constructor checks
- renamed `BaseBarConvertibleBuilderTest` to `BaseBarSeriesBuilderTest` (since we no longer have the `BaseBarConvertibleBuilder` class and its tests are now only performed by `BaseBarSeriesBuilder`)


ta4j-core/src/main/java/org/ta4j/core/bars/BaseBarBuilder.java
```diff
@@ -50,10 +50,16 @@ public class BaseBarBuilder implements BarBuilder {
     private long trades;
     private BarSeries baseBarSeries;
 
+    /** A builder to build a new {@link BaseBar} with {@link DoubleNumFactory} */
     public BaseBarBuilder() {
         this(DoubleNumFactory.getInstance());
     }
 
+    /**
+     * Constructor.
```
sgflt: I am not a fan of documentation that does not tell anything. 

I know it is constructor by language syntax. 
I know it takes numFactory.

There is no added value, just boiler plate doc.

nimo23: I am also not a fan of boilerplate docs. But ta4j has a Javadoc for all constructors (even if this is usually a self-explanatory case, like here):


```
    /**
     * Constructor.
     *
     * @param numFactory
     */
    public BaseBarBuilder(final NumFactory numFactory) {
        this.numFactory = numFactory;
    }
```

I can delete it here. I don't care. Should we also delete it for other self-describing constructors?

sgflt: I vote for deletion.

nimo23: If we do it, we should do it uniformly in all other places. Deleting it here and leaving it somewhere else is not a good idea either. This can be discussed in a new issue (e.g. "best practice in api documentation standards"). @sgflt If you want, you can open a new topic for discussion.

But in principle, I'm also against unnecessary javadocs (as in this case).

TheCookieLab: If a comment is deemed by all parties to be unnecessary within this change set, it should be removed. That the pattern exists elsewhere in the codebase should not prevent said removal, otherwise we are intentionally adding tech debt.

TheCookieLab (approved):

Issue #916, opened 2022-11-11
remove Num abstraction


nimo23:
Should we remove the `Num`-abstraction and use raw datatypes?

After a few years, we should debate again why there is a need for a `Num`-abstraction:

Important information:
- https://github.com/ta4j/ta4j/issues/113

According to @team172011 in https://github.com/ta4j/ta4j/pull/909#discussion_r1020443581:

> In general I must admit that I have the feeling that the whole Num abstraction is not relevant for the vast majority of users.

I think, this is a good point which should be further discussed here. 

We want to discuss whether there are compelling or beneficial reasons for the existence of the `Num` abstraction layer. If not, we could remove this `Num` abstraction layer from ta4j. 

If we actually delete this `Num`-abstraction layer, however, we should see where `BigDecimal` is absolutely necessary (e.g. Pnl-Criteria or all values ​​that have to do with money).

miracle-kang:
The most problem is scale and rounding mode could not control by user. This may cause the last decimal to be wrong.

I think it‘s better to use BigDecImal

sergproua:
I vote for Num to stay.  There are many decimal type implementations with way better performance than BigDecimal. Here are few https://github.com/tools4j/decimal4j, https://github.com/JavaMoney/jsr354-ri

gelhaimer:
> I vote for Num to stay. There are many decimal type implementations with way better performance than BigDecimal. Here are few https://github.com/tools4j/decimal4j, https://github.com/JavaMoney/jsr354-ri

I've tried to implement Num interface using Decimal4j as a delegate. It broke some indicators like the ADX. All value are NaN.
It probably comes from my implementation, but this is really hard to debug.

The flexibility that Num interface provides seems confortable at first glance, but I think there is a lot of subtilities to go around in order to make custom implementation work.

PR #1226, opened 2024-11-22
Movingaverages

brhurley:
Fixes #.

- 
- 
- 


ta4j-core/src/main/java/org/ta4j/core/indicators/averages/JMAIndicator.java
```diff
@@ -0,0 +1,95 @@
...
+    public JMAIndicator(Indicator<Num> indicator, int barCount, double phase, double power) {
+        super(indicator.getBarSeries());
+        this.indicator = indicator;
+        this.barCount = barCount;
+        this.phase = Math.min(Math.max(phase, -100), 100); // Clamp phase between -100 and 100
+        this.power = Math.max(1, power); // Ensure power is at least 1
+        this.jmaDataMap = new HashMap<>();
+
+        // Compute smoothing factor based on phase
+        beta = 0.45 * (barCount - 1) / (0.45 * (barCount - 1) + 2);
```
nimo23: If calculations are made directly in `double`, then `DecimalNum` is bypassed. Is this desired? Normally, the methods from the `Num` interface are used to calculate these indicators. If this were not the case, there would be no reason for `DoubleNum` and `DecimalNum` to exist. 

The same problem applies to all other indicators in this PR where direct `double` operations were performed. I know that this makes the code very complicated by using `Num` wrapper code, but that's the way it is currently (see https://github.com/ta4j/ta4j/issues/916).

brhurley:
@nimo23 Ready for review

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/JMAIndicator.java
```diff
@@ -0,0 +1,137 @@
...
+
+        NumFactory numFactory = indicator.getBarSeries().numFactory();
+        Num currentPrice = indicator.getValue(index);
+
+        if (index <= 0) {
+            jmaDataMap.put(index, new JmaData(currentPrice, numFactory.zero(), numFactory.zero(), currentPrice));
+            return currentPrice;
+        }
+
+        JmaData previousJMA = jmaDataMap.get(index - 1);
```
sgflt, line 101: This may work for backtesting, but causes memory leak within long running livetrading.

I can't see clean up after window passes further.

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/LSMAIndicator.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @Override
+    protected Num calculate(int index) {
+        if (index < barCount - 1) {
+            return indicator.getValue(index);
+        }
+
+        // Step 1: Calculate AvgPrice
+        Num sumPrice = numFactory.numOf(0);
+        for (int i = 0; i < barCount; i++) {
```
sgflt: Is it possible to reuse RunningTotalndicator?

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/LSMAIndicator.java
```diff
@@ -0,0 +1,130 @@
...
+            sumPrice = sumPrice.plus(indicator.getValue(index - i));
+        }
+        Num avgPrice = sumPrice.dividedBy(numFactory.numOf(barCount));
+
+        // Step 2: Calculate AvgTime
+        Num sumTime = numFactory.numOf(0);
+        for (int i = 1; i <= barCount; i++) {
```
sgflt: Is it possible to reuse RunningTotalndicator?

ta4j-core/src/test/java/org/ta4j/core/indicators/averages/VIDYAIndicatorTest.java
```diff
@@ -0,0 +1,67 @@
...
+        System.out.println("------");
+        System.out.println(vidya);
+        System.out.println("index, close, expected, vidya, diff, passed");
+        for (int i = 0; i < barSeries.getBarCount(); i++) {
+            Num expected = mock.getValue(i);
+            Num value = vidya.getValue(i);
+
+            System.out.println(i + ", " + barSeries.getBar(i).getClosePrice() + ", " + expected + ", " + value + ", "
+                    + expected.minus(value) + ", " + (Math.abs((expected.minus(value)).doubleValue()) < 0.001));
+            // System.out.println(value);
```
sgflt: some unused code

ta4j-core/src/test/java/org/ta4j/core/indicators/averages/VIDYAIndicatorTest.java
```diff
@@ -0,0 +1,67 @@
...
+        VIDYAIndicator vidya = new VIDYAIndicator(new ClosePriceIndicator(barSeries), 9, 20);
+
+        System.out.println("------");
+        System.out.println(vidya);
+        System.out.println("index, close, expected, vidya, diff, passed");
+        for (int i = 0; i < barSeries.getBarCount(); i++) {
+            Num expected = mock.getValue(i);
+            Num value = vidya.getValue(i);
+
+            System.out.println(i + ", " + barSeries.getBar(i).getClosePrice() + ", " + expected + ", " + value + ", "
```
sgflt: use logger instead of stdout

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/WildersMAIndicator.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    public int getUnstableBars() {
+        return barCount * 2;
+    }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + " barCount: " + barCount;
```
sgflt, line 110: I personally use semantic toStrings in format `INDICATOR_SHORTCUT(parameters) => current value`, that simplifies debugging.

In this case it would be `"WildersMA(%s, %d)".formatted(this.indicator, this.barCount)` at least. Formatted for example as `WildersMA(SMA(5), 5)`.

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/WildersMAIndicator.java
```diff
@@ -0,0 +1,111 @@
...
+
+        // Simulate extended historical data for initialization
+        if (index < barCount) {
+            // Pretend there are extra `barCount` points before the first real point
+            // Use the first actual value as a proxy for the "missing" earlier data
+            Num simulatedValue = indicator.getValue(0);
+
+            // Initialize with a simulated Simple Moving Average (SMA)
+            Num sum = numFactory.zero();
+            for (int i = 0; i <= index; i++) {
```
sgflt: Is it possible to reuse RunningTotalIndicator?

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/JMAIndicator.java
brhurley: I wouldn't classify this as a memory leak really.  It's a memory growth issue.   However, I did look at this more and attempted to extract E0, E1 and E2 into their own indicators and that didn't work either since E2 is dependent on JMA which would introduce a circular dependency.   I also looked at limited the jmaDataMap to a fixed size and that doesn't work either since any other reference to a previous JMA would fail. 

If you have any ideas on a different approach here let me know and I'll look into it.

The E2 calculation is the sticking point here since that references E0, E1, previous_E2 and previous_JMA.

e2 = (e0 + phaseRatio * e1 - JMA_prev) * (1 - alpha)^2 + alpha^2 * e2_prev

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/LSMAIndicator.java
brhurley: I added RunningTotalndicator to the sumPrice.

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/LSMAIndicator.java
brhurley: Removed this to an initialization calculation.

ta4j-core/src/test/java/org/ta4j/core/indicators/averages/VIDYAIndicatorTest.java
brhurley: Removed

ta4j-core/src/test/java/org/ta4j/core/indicators/averages/VIDYAIndicatorTest.java
brhurley: Removed

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/WildersMAIndicator.java
brhurley: I don't disagree with you but I was just following the pattern used for all the other indicators.  Would you like me to change the ones that I created then we can do a PR to change all the others?

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/WildersMAIndicator.java
brhurley: Changed to RunningTotalIndicator

brhurley:
Resolved conflict with updated CHANGELOG.md

brhurley:
Updated code for new method in Indicator getCountOfUnstableBars

brhurley:
@sgflt @nimo23 Let me know if there is any additional feedback or if this is ready to commit.

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/LSMAIndicator.java
```diff
@@ -0,0 +1,123 @@
...
+     */
+    public LSMAIndicator(Indicator<Num> indicator, int barCount, int offset) {
+        super(indicator.getBarSeries());
+        this.indicator = indicator;
+        this.barCount = barCount;
+        this.offset = offset;
+        this.barSeries = indicator.getBarSeries();
+        this.numFactory = barSeries.numFactory();
+
+        Num sumTime = numFactory.numOf(0);
```
nimo23: use `Num sumTime = numFactory.zero();`

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/LSMAIndicator.java
```diff
@@ -0,0 +1,123 @@
...
+ */
+public class LSMAIndicator extends CachedIndicator<Num> {
+
+    private final Indicator<Num> indicator;
+    private final int barCount;
+    private final int offset;
+    private final NumFactory numFactory;
+    private final BarSeries barSeries;
+    private final Num avgTime;
+    protected final Logger log = LoggerFactory.getLogger(getClass());
```
nimo23: Remove the `log`. It is already available as a field via `AbstractIndicator` (the parent of `CachedIndicator`).

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/LSMAIndicator.java
```diff
@@ -0,0 +1,123 @@
...
+
+        // Calculate AvgPrice
+        RunningTotalIndicator sumPriceIndicator = new RunningTotalIndicator(indicator, barCount);
+
+        Num sumPrice = sumPriceIndicator.getValue(index);
+        Num avgPrice = sumPrice.dividedBy(numFactory.numOf(barCount));
+
+        // Calculate sx and sy
+        Num sx = numFactory.numOf(0);
+        Num sy = numFactory.numOf(0);
```
nimo23: use `numFactory.zero()` instead of `numFactory.numOf(0)`

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/MCGinleyMAIndicator.java
```diff
@@ -0,0 +1,98 @@
...
+ * to false signals and more reliable for identifying trends in both volatile
+ * and stable markets.
+ *
+ */
+public class MCGinleyMAIndicator extends CachedIndicator<Num> {
+
+    private final Indicator<Num> indicator;
+    private final int barCount;
+    private final NumFactory numFactory;
+    private final BarSeries barSeries;
```
nimo23: No need to store `barSeries` again. It is already available by its parent class via `getBarSeries().`

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/MCGinleyMAIndicator.java
```diff
@@ -0,0 +1,98 @@
...
+ * automatically based on the speed of price movement. This makes it less prone
+ * to false signals and more reliable for identifying trends in both volatile
+ * and stable markets.
+ *
+ */
+public class MCGinleyMAIndicator extends CachedIndicator<Num> {
+
+    private final Indicator<Num> indicator;
+    private final int barCount;
+    private final NumFactory numFactory;
```
nimo23: No need to store `numFactory`. Call it via `getBarSeries().numFactory()`.

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/SGMAIndicator.java
```diff
@@ -0,0 +1,166 @@
...
+ *
+ */
+public class SGMAIndicator extends CachedIndicator<Num> {
+
+    private final int barCount;
+    private final Indicator<Num> indicator; // Input series
+    private final Num polynomialOrder; // The degree of the polynomial as a Num
+    private final List<Num> coefficients; // Pre-computed coefficients for smoothing
+    private final BarSeries barSeries;
+    private final NumFactory numFactory;
```
nimo23: Remove `numFactory` and `barSeries` properties (reason: see above).

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/VIDYAIndicator.java
```diff
@@ -0,0 +1,97 @@
...
+ * results in a highly responsive yet stable moving average that is particularly
+ * useful in dynamic market conditions.
+ *
+ */
+public class VIDYAIndicator extends CachedIndicator<Num> {
+
+    private final Indicator<Num> indicator; // Input price (e.g., close price)
+    private final int cmoPeriod; // Lookback period for cmoPeriod
+    private final int vidyaPeriod; // Lookback period for VIDYA
+    private final NumFactory numFactory;
```
nimo23: Remove `numFactory` property (reason: access is already given via `getBarSeries.numFactory()`)

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/WildersMAIndicator.java
```diff
@@ -0,0 +1,111 @@
...
+    public WildersMAIndicator(Indicator<Num> indicator, int barCount) {
+        super(indicator);
+        this.indicator = indicator;
+        this.barCount = barCount;
+    }
+
+    @Override
+    protected Num calculate(int index) {
+        NumFactory numFactory = getBarSeries().numFactory();
+        Num one = numFactory.numOf(1);
```
nimo23: use `numFactory.one()` (instead of `numFactory.numOf(1)`).

brhurley:
Updated as requested, ran format and tests again.

[INFO] Tests run: 22, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Ta4j Parent 0.18-SNAPSHOT:
[INFO]
[INFO] Ta4j Parent ........................................ SUCCESS [  0.044 s]
[INFO] Ta4j Core .......................................... SUCCESS [ 38.145 s]
[INFO] Ta4j Examples ...................................... SUCCESS [  9.051 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  47.321 s
[INFO] Finished at: 2024-12-04T22:01:01-06:00

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/LSMAIndicator.java
```diff
@@ -92,8 +86,8 @@ protected Num calculate(int index) {
         Num avgPrice = sumPrice.dividedBy(numFactory.numOf(barCount));
 
         // Calculate sx and sy
-        Num sx = numFactory.numOf(0);
-        Num sy = numFactory.numOf(0);
+        Num sx = numFactory.zero();
+        Num sy = numFactory.zero();
```
nimo23: Please write:

```
Num zero = numFactory.zero();
Num sx = zero;
Num sy = zero;
```

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/MCGinleyMAIndicator.java
```diff
@@ -79,7 +73,10 @@ protected Num calculate(int index) {
         Num currentPrice = indicator.getValue(index);
 
         // Speed ratio (smoothing factor)
-        Num speedRatio = numFactory.numOf(barCount).multipliedBy(currentPrice.dividedBy(previousMcGinley).pow(2));
+        Num speedRatio = indicator.getBarSeries()
+                .numFactory()
+                .numOf(barCount)
+                .multipliedBy(currentPrice.dividedBy(previousMcGinley).pow(2));
```
nimo23, line 78: Please write:

```
Num numBars = indicator.getBarSeries().numFactory().numOf(barCount):
Num speedRatio = numBars.multipliedBy(currentPrice.dividedBy(previousMcGinley).pow(2));
```

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/WildersMAIndicator.java
```diff
@@ -0,0 +1,111 @@
...
+        this.indicator = indicator;
+        this.barCount = barCount;
+    }
+
+    @Override
+    protected Num calculate(int index) {
+        NumFactory numFactory = getBarSeries().numFactory();
+        Num one = numFactory.one();
+
+        Num k = one.dividedBy(numFactory.numOf(barCount));
```
nimo23: write:

```
Num numBars = numFactory.numOf(barCount);
Num k = one.dividedBy(numBars);
```

and reuse `numBars` below..

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/WildersMAIndicator.java
```diff
@@ -0,0 +1,111 @@
...
+            // Use the first actual value as a proxy for the "missing" earlier data
+            Num simulatedValue = indicator.getValue(0);
+
+            // Initialize with a simulated Simple Moving Average (SMA)
+            Num sum = sumPriceIndicator.getValue(index);
+
+            // Return average of available points for initialization
+            Num preResult = sum.plus(simulatedValue.multipliedBy(numFactory.numOf(barCount - index - 1)))
+                    .dividedBy(numFactory.numOf(barCount));
+
```
nimo23, line 89: use `numBars` from above and write:

`..dividedBy(numBars);`

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/JMAIndicator.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Override
+    protected Num calculate(int index) {
+
+        NumFactory numFactory = indicator.getBarSeries().numFactory();
+        Num currentPrice = indicator.getValue(index);
+
+        if (index <= 0) {
+            jmaDataMap.put(index, new JmaData(currentPrice, numFactory.zero(), numFactory.zero(), currentPrice));
+            return currentPrice;
```
nimo23, line 98: write:

```
Num zero = numFactory.zero();
jmaDataMap.put(index, new JmaData(currentPrice, zero, zero, currentPrice));
```

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/SMMAIndicator.java
```diff
@@ -0,0 +1,87 @@
...
+        Num previousSMMA = getValue(index - 1);
+
+        // Current price
+        Num currentPrice = indicator.getValue(index);
+
+        // SMMA formula
+        return previousSMMA.multipliedBy(indicator.getBarSeries().numFactory().numOf(barCount - 1))
+                .plus(currentPrice)
+                .dividedBy(indicator.getBarSeries().numFactory().numOf(barCount));
+    }
```
nimo23, line 78: write:

```
var numFactory = indicator.getBarSeries().numFactory();
return previousSMMA.multipliedBy(numFactory.numOf(barCount - 1))
                .plus(currentPrice)
                .dividedBy(numFactory.numOf(barCount));
```

TheCookieLab:
@nimo23 approving the PR yet asking for changes sends a mixed message, both for the author and myself. In the future please refrain from approving unless the code is to your satisfaction _as-is_.

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/LSMAIndicator.java
brhurley: updated

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/MCGinleyMAIndicator.java
brhurley: updated

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/SMMAIndicator.java
brhurley: Updated

brhurley:
@nimo23 I made the changes you requested.   Let me know if there is anything else, however, I'd like to get this committed if possible.

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/WildersMAIndicator.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    protected Num calculate(int index) {
+        NumFactory numFactory = getBarSeries().numFactory();
+        Num one = numFactory.one();
+
+        Num numBars = numFactory.numOf(barCount);
+        Num k = one.dividedBy(numBars);
+        Num prevWMA = indicator.getValue(0);
+
+        RunningTotalIndicator sumPriceIndicator = new RunningTotalIndicator(indicator, barCount);
```
sgflt: As this will work, it cannot benefit from fastPath() optimized performance. Instance field would use fastPath.

nimo23: Yes, 

`RunningTotalIndicator sumPriceIndicator = new RunningTotalIndicator(indicator, barCount);`

should be defined in the constructor. Please do this in **all** classes where `RunningTotalIndicator` is used.

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/JMAIndicator.java
```diff
@@ -0,0 +1,150 @@
...
+                .multipliedBy(numFactory.numOf(barCount - 1))
+                .dividedBy(numFactory.numOf(0.45).multipliedBy(numFactory.numOf(barCount - 1)).plus(numFactory.two()));
+
+        phaseRatio = this.phase.isLessThan(numFactory.numOf(-100)) ? numFactory.numOf(0.5)
+                : (this.phase.isGreaterThan(numFactory.numOf(100)) ? numFactory.numOf(2.5)
+                        : this.phase.dividedBy(numFactory.numOf(100)).plus(numFactory.numOf(1.5)));
+
+        alpha = beta.pow(this.power);
+
+        for (int i = 0; i < indicator.getBarSeries().getBarCount(); i++) {
```
francescopeloi: do all barSeries start from zero? I usually do:
``for (int i = barSeries.getStartIndex(); i <= barSeries.getEndIndex(); i++) {``
the existance of getStartIndex() made me think that they could start from something different than 0, but I am not sure.

nimo23: Please use the following loop for all indicators:

```
@Override

    final int startIndex = Math.max(0, index - barCount + 1);
    for (int i = startIndex; i <= index; i++) {
        // ..
    }

}
```

TheCookieLab: This and the above change should be the last two. I know it's not always possible but reviewers should aim to capture _all_ items within a single review.  There should not be multiple round trips unless subsequent comments are on changes just implemented.

brhurley: I made these changes but I had to go with @francescopeloi approach since I don't have index at that point, anyway I had to use `BarSeries.getBeginIndex()`.   There is no `getStartIndex()`.

brhurley:
Ready for review, again.

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/JMAIndicator.java
```diff
@@ -0,0 +1,155 @@
...
+                .multipliedBy(numFactory.one().minus(beta))
+                .plus(previousJMA.e1.multipliedBy(beta));
+        return e1;
+    }
+
+    /*
+     * Since previous e2 and previous JMA has to be used to calculate e2 I didn't
+     * see a way to abstract this out to a separate indicator which would have made
+     * JMA so much simpler to calculate and we wouldn't need the internal cache.
+     */
```
nimo23, line 132: Just a thought: Could we calculate the **previous values** by using `RecursiveCachedIndicator` instead of using an internal cache?

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/KiJunV2Indicator.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param barCount        the Simple Moving Average time frame
+     * @param polynomialOrder the degree of the polynomial, default 2
+     */
+    public KiJunV2Indicator(Indicator<Num> highPrice, Indicator<Num> lowPrice, int barCount) {
+        super(highPrice.getBarSeries());
+        this.highPrice = highPrice;
+        this.lowPrice = lowPrice;
+        this.barCount = barCount;
+        this.numFactory = highPrice.getBarSeries().numFactory();
+    }
```
nimo23, line 66: As an alternative, you can define the following in the constructor:

```
HighestValueIndicator highestValue = new HighestValueIndicator(HighPriceIndicator(barSeries));

LowestValueIndicator lowestValue = new LowestValueIndicator(LowPriceIndicator(barSeries));
```

and use

```
        // Get the highest high and lowest low within the barCount period
        Num highestHigh = highestValue.calculate(index);
        Num lowestLow = lowestValue.calculate(index);

        // Calculate the midpoint
        return highestHigh.plus(lowestLow).dividedBy(numFactory.two());
    }
```


The problem with your approach (even though it is more memory friendly) is that values ​​might contain "NaN", which could lead to unchecked errors. The HighestValueIndicator/LowestValueIndicator takes care of such things.

_@brhurley Btw, these are really the last points I'll address here. After that, we can merge it. I'll then create a follow-up PR that moves the other average indicators (from the parent package) to this new `average`-package. Thanks again for your efforts!_

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/JMAIndicator.java
brhurley: I tried this...


`       

 this.e0Indicator = new RecursiveCachedIndicator<Num>(indicator) {

            @Override
                if (index <= 0) {
                    return numFactory.zero();
                }

                Num currentPrice = indicator.getValue(index);
                // Get the previous e0 value
                Num previousE0 = getValue(index - 1);

                Num e0 = currentPrice.multipliedBy(numFactory.one().minus(alpha)).plus(previousE0.multipliedBy(alpha));

                return e0;
            }

            @Override
                return 0;
            }
        };

        this.e1Indicator = new RecursiveCachedIndicator<Num>(indicator) {

            @Override
                if (index <= 0) {
                    return numFactory.zero();
                }

                Num currentPrice = indicator.getValue(index);
                // Get the previous e0 value
                Num previousE1 = getValue(index - 1);

                Num e1 = currentPrice.minus(JMAIndicator.this.e0Indicator.getValue(index))
                        .multipliedBy(numFactory.one().minus(beta))
                        .plus(previousE1.multipliedBy(beta));

                return e1;
            }

            @Override
                return 0;
            }
        };

        this.e2Indicator = new RecursiveCachedIndicator<Num>(indicator) {

            @Override
                if (index <= 0) {
                    return numFactory.zero();
                }

                // Get the previous e0 value
                Num previousE2 = getValue(index - 1);

                Num e2 = JMAIndicator.this.e0Indicator.getValue(index)
                        .plus(phaseRatio.multipliedBy(JMAIndicator.this.e1Indicator.getValue(index)))
                        .minus(JMAIndicator.this.getValue(index))
                        .multipliedBy(numFactory.one().minus(alpha).pow(2))
                        .plus(previousE2.multipliedBy(alpha.pow(2)));

                return e2;
            }

            @Override
                return 0;
            }
        };
        
`

Which gives stack overflow error.

Just using the E0 and E1 without the E2 as a RecursiveCachedIndicator, I don't get a stack overflow but the numbers are off by a little.  At least enough not to pass the test.   Do you see anything wrong with this implementation that would cause the stack overflow, or changes in calculation with

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/KiJunV2Indicator.java
brhurley: changed  KiJunV2Indicator.java

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/JMAIndicator.java
nimo23: I think the stackoverflow error occur due to an infinite `calculate()`-loop. I don't see the whole class. **Here some checkpoints:**

- You need to do: `public class JMAIndicator extends RecursiveCachedIndicator<Num>{..}` 
- Using something like `this.intermediateValue = new RecursiveCachedIndicator<Num>(ref) {..}` should be done at **constructor level** (if possible) and then `this.intermediateValue(s)` are stored as class properties (accessible within the method `protected Num calculate(int index)`). The `intermediateValue(s)` represents your previous "internal cache"..

brhurley: Here I put the new JMA in this branch.   This gives the stack overflow.   If I understand your last comment I've done all the above.

https://github.com/brhurley/ta4j/blob/movingaverages-jma/ta4j-core/src/main/java/org/ta4j/core/indicators/averages/JMAIndicator.java

Let me know if you can't see that branch and I can open a draft PR if needed.

brhurley: I still have some of the old code in that class that would obviously be removed.  Just testing to see if I could get the RecursiveCachedIndicator working.

brhurley:
If this is good now what is the next step?

TheCookieLab (approved):

Issue #1230, opened 2024-12-07
[BUG] - MockBarSeriesBuilder and MockBarBuilder using inconsistent time


ek-ex:
**I have checked existing issues and wiki**
- [x] I could not find similar [issues](https://github.com/ta4j/ta4j/issues?utf8=%E2%9C%93&q=)
- [x] I could not find a solution in the [wiki](https://ta4j.github.io/ta4j-wiki/) or [faq section](https://ta4j.github.io/ta4j-wiki/FAQ.html)

**Describe the bug**
There is an inconsistency on the time scale used between MockBarBuilder and MockBarSeriesBuilder. One is using today's date and the other is using beginning of epoch times. 
![Screenshot from 2024-12-06 16-51-21](https://github.com/user-attachments/assets/684d1eca-e681-4b5f-8cbc-05f3c4156bb2)
![Screenshot from 2024-12-06 16-51-46](https://github.com/user-attachments/assets/005f5a48-2272-4902-82b6-b1ae2a2cfde7)

The test RecentSwingHighIndicatorTest uses both. New versions of Maven runs the test in a different order and if you run the test alone the test will fail with this error:
```
[INFO] 
[ERROR] Errors: 
[ERROR]   RecentSwingHighIndicatorTest.testCalculate_OnMovingBarSeries_ReturnsValue:211 » IllegalArgument Cannot add a bar with end time:1975-02-28T00:00:00Z that is <= to series end time: 2024-12-07T00:12:00.897172497Z
[ERROR]   RecentSwingHighIndicatorTest.testCalculate_OnMovingBarSeries_ReturnsValue:211 » IllegalArgument Cannot add a bar with end time:1975-07-31T00:00:00Z that is <= to series end time: 2024-12-07T00:12:00.900613375Z
[INFO] 
[ERROR] Tests run: 1253, Failures: 0, Errors: 2, Skipped: 0
```

The date from MockBarSeriesBuilder is newer than the date from MockBarBuilder

![Screenshot from 2024-12-06 16-47-37](https://github.com/user-attachments/assets/fd9539a9-0314-474f-982e-74ede4dde0c5)

The reason a `mvn clean install` works, is that enough Bars are created so that this is not a problem.
![Screenshot from 2024-12-06 16-55-46](https://github.com/user-attachments/assets/011f4983-a2a8-4ecc-92d6-a10ad452cde2)

So the test is not deterministic. 

**To Reproduce**
[...]

PR #1231, opened 2024-12-07
Fixed MockBarBuilder to use Instant.now for beginTime

nimo23:
Fixes https://github.com/ta4j/ta4j/issues/1230

- Fixed `MockBarBuilder` to use `Instant.now` for beginTime

I replaced `Clock` by using simply `Instant` because we do not need a clock in this case (which wraps an UTC).

Both are equal:

```
private Clock clock = Clock.fixed(Instant.now(), ZoneOffset.UTC);
   
endTime(Clock.offset(clock, timePeriod.multipliedBy(++countOfProducedBars)).instant());
```

Simplified to:

```
private Instant beginTime = Instant.now();
endTime(beginTime.plus(timePeriod.multipliedBy(++countOfProducedBars)));
```

### DIFF
--- a/ta4j-core/src/test/java/org/ta4j/core/mocks/MockBarBuilder.java
+++ b/ta4j-core/src/test/java/org/ta4j/core/mocks/MockBarBuilder.java
@@ -23,18 +23,15 @@
  */
 package org.ta4j.core.mocks;
 
-import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
-import java.time.ZoneOffset;
-
 import org.ta4j.core.bars.BaseBar;
 import org.ta4j.core.bars.BaseBarBuilder;
 import org.ta4j.core.num.NumFactory;
 
 public class MockBarBuilder extends BaseBarBuilder {
 
-    private Clock clock = Clock.fixed(Instant.ofEpochMilli(0), ZoneOffset.UTC);
+    private Instant beginTime = Instant.now();
     private boolean periodSet;
     private boolean endTimeSet;
 
@@ -65,8 +62,9 @@ public class MockBarBuilder extends BaseBarBuilder {
         }
 
         if (!endTimeSet) {
-            endTime(Instant.now(Clock.offset(clock, timePeriod.multipliedBy(++countOfProducedBars))));
+            endTime(beginTime.plus(timePeriod.multipliedBy(++countOfProducedBars)));
         }
         return super.build();
     }
+
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - Fixed `SMAIndicatorTest` to set the endTime of the next bar correctly
 - Fixed `SMAIndicatorMovingSeriesTest` to set the endTime of the next bar correctly
 - Use UTC TimeZone for `AroonOscillatorIndicatorTest`, `PivotPointIndicatorTest`
+- Fixed `MockBarBuilder` to use `Instant.now` for beginTime
 
 ### Changed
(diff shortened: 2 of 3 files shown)


ta4j-core/src/test/java/org/ta4j/core/mocks/MockBarBuilder.java
```diff
@@ -23,18 +23,15 @@
...
-import java.time.ZoneOffset;
-
 import org.ta4j.core.bars.BaseBar;
 import org.ta4j.core.bars.BaseBarBuilder;
 import org.ta4j.core.num.NumFactory;
 
 public class MockBarBuilder extends BaseBarBuilder {
 
-    private Clock clock = Clock.fixed(Instant.ofEpochMilli(0), ZoneOffset.UTC);
+    private Instant beginTime = Instant.now();
```
sgflt, line 34: I would still use predefined Instant.EPOCH at least for debugging purpose. It is harder to debug code that have each run in different time span.

Example:
First run have bars 00:00, 00:01, 00:02. Second run in a nother part of day 12:30, 12:31, 12:32.
These differences may discract from real cause of indicator miscalculation (Renko case?).


From another point of view random time may prove it works in any case. Maybe it should be configurable by tester, what he want to test like `mockbarBilder.withStartAtEpoch()` or `mockbarBuilder.witnStartAtRealTime()`.

nimo23: >Maybe it should be configurable by tester, what he want to test like mockbarBilder.withStartAtEpoch() or mockbarBuilder.witnStartAtRealTime().

That's a good idea. Currently we assume something like `mockbarBuilder.witnStartAtRealTime() ` (where it is possible that a date in the future from `NOW` was set as the `endTime` of a bar, e.g. `lastBar.endTime = 2027-01-01` - even if it is not an error.)

@sgflt Could you provide a new PR with your suggestions? I will then close this PR.

ek-ex: Can I suggest creating a static method that would provide incremental end dates that would be used by both `MockBarBuilder` and `MockBarSeriesBuilder` keeping consistency between both?

nimo23: @ek-ex Sure, create a PR. I will close this PR after a better one is suggested. 

@sgflt are you preparing a PR with your suggestion? I think, something like `mockbarBuilder.witnStartAtRealTime()` is not needed. We should stick with `Instant.EPOCH` (when mocking).

sgflt: Yes, I have replaced now by EPOCH.

sgflt: Wouldn't be better if you pulled my branch to yours? This PR may stay.

nimo23: @sgflt No, please create a new PR - I will then close this one. 

Unfortunately, simply using `Instant.EPOCH` in MockBarBuilder is not enough (you will see the test errors when running the tests after changing to `Instant.EPOCH`).

You also have to consider the 

https://github.com/ta4j/ta4j/blob/607cd8c451da6ddcfe2ca19042e4eed62ccc698f/ta4j-core/src/test/java/org/ta4j/core/mocks/MockBarSeriesBuilder.java#L90 

and

https://github.com/ta4j/ta4j/blob/607cd8c451da6ddcfe2ca19042e4eed62ccc698f/ta4j-core/src/test/java/org/ta4j/core/mocks/MockBarSeriesBuilder.java#L90

Please create a new PR with your suggestion.

sgflt:
=> #1234

PR #1232, opened 2024-12-07
RecentSwingHighIndicatorTest Update failing tests

### LINKED ISSUE #1230: [BUG] - MockBarSeriesBuilder and MockBarBuilder using inconsistent time

ek-ex:
Fixes #1230

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - Fixed `SMAIndicatorTest` to set the endTime of the next bar correctly
 - Fixed `SMAIndicatorMovingSeriesTest` to set the endTime of the next bar correctly
 - Use UTC TimeZone for `AroonOscillatorIndicatorTest`, `PivotPointIndicatorTest`
+- Fixed `RecentSwingHighIndicatorTest` to create bars consistently
 
 ### Changed
 - Updated **jfreechart** dependency in **ta4j-examples** project from 1.5.3 to 1.5.5 to resolve [CVE-2023-52070](https://ossindex.sonatype.org/vulnerability/CVE-2023-6481?component-type=maven&component-name=ch.qos.logback%2Flogback-core)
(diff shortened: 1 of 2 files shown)


ek-ex:
Agreed this is not a solution to the problem, but will guarantee stability to this test. I would not harm having this additional to the other PR.

TheCookieLab (approved):

Issue #1228, opened 2024-11-23
What is the status of DecimalNum?


if6was9:
When I started using TA4J I started with DecimalNum.  Everything was fine for the most part, but indicators started to give nonsensical results in certain cases.

It was a very confusing problem to debug.  I finally thought "Why don't I try DoubleNum and see if the problem goes away?"  Sure enough, it did.

Most of the problem as far as I cared to tell had to do with BigDecimal scale propagation from arguments passed in.  Expected loss of decimal precision was the result, which cascaded into nonsensical output from derived indicators.

I concluded that DecimalNum was just experimental and pre-alpha.  I considered fixing it, because I don't think it would be all that hard, but switching to DoubleNum was easier and entirely sufficient.

Is my experience unusual? Is DecimalNum being actively used?

sgflt:
Any example?

TheCookieLab:
@if6was9 have you revisited this?

if6was9:
This is what I found to be the root of the problem. 

https://github.com/ta4j/ta4j/blob/master/ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java#L270

Clearly, from the code comments the issue is understood by some.  But it is far from obvious for casual/new users and the impacts stay can stay hidden until certain indicators are applied and the (unexpected) scale propagation turns into erroneous values.

The behaviors of these two DecimalNum values are very different:

```
DecimalNum.valueOf(new BigDecimal("1.234567").scale(2));
DecimalNum.valueOf(new BigDecimal("1.234567").scale(2).doubleValue());
```

The impact on indicators varied.  IIRC, SMA was fine, but EMA generated incorrect values since the scale loss was compounded many times over.

I just switched to DoubleNum and all the problems went away.

Not a bug, per se. Just very surprising behavior.

sgflt:
I think the NumFactory should be used in any case to preserve precision.
https://github.com/sgflt/ta4j/blob/52aeb0966207ad1b54141d13815cc2823be13f4c/ta4j-core/src/main/java/org/ta4j/core/num/DecimalNumFactory.java#L104C1-L107C4

Shouldn't we make Num implementations package private to prevent missuse? @nimo23 @TheCookieLab

if6was9:
I worked around this issue long ago, so it’s neither here nor there for me, but I do feel strongly that scale propagation violates the principle of least surprise.  

I’m sure there are some cases where this is desirable, but I can’t think of what they would be. 

I’m going to close the issue.

PR #1246, opened 2025-01-07
#1244 share math context

### LINKED ISSUE #1244: MathContext can be shared for all instances

After doing optimizations of whole codebase I have found there is unnecessary memory allocation for MathContext.
It is immutable  and may be shared for all instances created by single NumFactory.

https://github.com/sgflt/ta4j/commit/da0ab6d1a736e29b3773b190d558d52c4ed049df
https://github.com/sgflt/ta4j/commit/b3984ba65c795a4c29eaf558452499a2e6a8206f

**Additional context**
Both changes together saves about (1980 ms/3050 ms => 35 %) in test with 1370 MarketEvents that are used for strategy optimization with SimulatedAnnealing algorithm.

sgflt:
Fixes #1244.

- reduce memory allocation of MathContexts that are immutable


ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
```diff
@@ -186,8 +188,8 @@ public static DecimalNum valueOf(Number val) {
      * @return the {@code Num} with a precision of {@link #DEFAULT_PRECISION}
      * @throws NumberFormatException if {@code val} is {@code "NaN"}
      */
-    public static DecimalNum valueOf(DoubleNum val) {
-        return valueOf(val.doubleValue());
+    public static DecimalNum valueOf(final Num val) {
```
nimo23, line 196: Why not leave `DoubleNum` as a parameter? The javadoc also says: 

> Returns a {@code DecimalNum} version of the given {@code DoubleNum}.

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
```diff
@@ -196,8 +198,8 @@ public static DecimalNum valueOf(DoubleNum val) {
      * @param val the number
      * @return the {@code Num}
      */
-    public static DecimalNum valueOf(int val) {
-        return new DecimalNum(val);
+    public static DecimalNum valueOf(final int val, final MathContext mathContext) {
```
nimo23, line 206: Why adding `MathContext` as parameter instead of using _private_ `DEFAULT_MATH_CONTEXT`?

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
```diff
@@ -206,8 +208,8 @@ public static DecimalNum valueOf(int val) {
      * @param val the number
      * @return the {@code Num}
      */
-    public static DecimalNum valueOf(long val) {
-        return new DecimalNum(val);
+    public static DecimalNum valueOf(final long val, final MathContext mathContext) {
```
nimo23, line 216: Why adding `MathContext` as parameter instead of using _private_ `DEFAULT_MATH_CONTEXT`?

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
```diff
@@ -216,8 +218,8 @@ public static DecimalNum valueOf(long val) {
      * @param val the number
      * @return the {@code Num}
      */
-    public static DecimalNum valueOf(short val) {
-        return new DecimalNum(val);
+    public static DecimalNum valueOf(final short val, final MathContext mathContext) {
```
nimo23, line 226: Why adding `MathContext` as parameter instead of using _private_ `DEFAULT_MATH_CONTEXT`?

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
```diff
@@ -231,11 +233,11 @@ public static DecimalNum valueOf(short val) {
      *         value of {@code val}.
      * @throws NumberFormatException if {@code val} is {@code Float.NaN}
      */
-    public static DecimalNum valueOf(float val) {
+    public static DecimalNum valueOf(final float val, final MathContext mathContext) {
```
nimo23, line 241: Why adding `MathContext` as parameter instead of using _private_ `DEFAULT_MATH_CONTEXT`?

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
```diff
@@ -249,37 +251,42 @@ public static DecimalNum valueOf(float val) {
      *         value of {@code val}.
      * @throws NumberFormatException if {@code val} is {@code Double.NaN}
      */
-    public static DecimalNum valueOf(double val) {
+    public static DecimalNum valueOf(final double val, final MathContext mathContext) {
```
nimo23, line 259: Why adding `MathContext` as parameter instead of using _private_ `DEFAULT_MATH_CONTEXT`?

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
```diff
@@ -249,37 +251,42 @@ public static DecimalNum valueOf(float val) {
...
+     * precision of {@code precision}.
      *
-     * @param val the number
+     * @param val         the number
+     * @param mathContext the precision
      * @return the {@code Num}
      */
-    public static DecimalNum valueOf(BigDecimal val) {
-        return new DecimalNum(val, val.precision());
+    public static DecimalNum valueOf(final BigDecimal val, final MathContext mathContext) {
```
nimo23, line 274: Why adding `MathContext` as parameter instead of using _private_ `DEFAULT_MATH_CONTEXT`?

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
```diff
@@ -308,23 +315,23 @@ public String getName() {
...
         if (augend.isNaN()) {
             return NaN;
         }
-        BigDecimal bigDecimal = ((DecimalNum) augend).delegate;
-        int precision = mathContext.getPrecision();
-        BigDecimal result = delegate.add(bigDecimal, mathContext);
-        return new DecimalNum(result, precision);
+        final var decimalNum = (DecimalNum) augend;
+        final var sumContext = chooseMathContextWithGreaterPrecision(decimalNum, this);
+        final var result = this.delegate.add(decimalNum.delegate, sumContext);
```
nimo23, line 338: Why do we need `chooseMathContextWithGreaterPrecision(decimalNum, this);` ? Normally, the default precision (`DEFAULT_MATH_CONTEXT`) should always be adopted. Why to deviate from the default here?

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
```diff
@@ -351,14 +358,14 @@ public Num minus(Num subtrahend) {
...
         if (multiplicand.isNaN()) {
             return NaN;
         }
-        BigDecimal bigDecimal = ((DecimalNum) multiplicand).delegate;
-        int precision = mathContext.getPrecision();
-        BigDecimal result = delegate.multiply(bigDecimal, new MathContext(precision, RoundingMode.HALF_UP));
-        return new DecimalNum(result, precision);
+        final var decimalNum = (DecimalNum) multiplicand;
+        final var multiplicationContext = chooseMathContextWithGreaterPrecision(decimalNum, this);
+        final var result = this.delegate.multiply(decimalNum.delegate, multiplicationContext);
```
nimo23, line 372: Why do we need `chooseMathContextWithGreaterPrecision(decimalNum, this);` ? Normally, the default precision (`DEFAULT_MATH_CONTEXT`) should always be adopted. Why to deviate from the default here?

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
```diff
@@ -385,108 +392,104 @@ public Num dividedBy(Num divisor) {
...
         if (divisor.isNaN()) {
             return NaN;
         }
-        BigDecimal bigDecimal = ((DecimalNum) divisor).delegate;
-        int precision = mathContext.getPrecision();
-        BigDecimal result = delegate.remainder(bigDecimal, new MathContext(precision, RoundingMode.HALF_UP));
-        return new DecimalNum(result, precision);
+        final var decimalNum = (DecimalNum) divisor;
+        final var moduloContext = chooseMathContextWithGreaterPrecision(decimalNum, this);
+        final var result = this.delegate.remainder(decimalNum.delegate, moduloContext);
```
nimo23, line 406: Why do we need `chooseMathContextWithGreaterPrecision(decimalNum, this);` ? Normally, the default precision (`DEFAULT_MATH_CONTEXT`) should always be adopted. Why to deviate from the default here?

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
sgflt: Because factory decides what precision it has. Constructor with DEFAULT_MATH_CONTEXT is used mostly in tests, so i left it as is. 

It is possible to migrate to  `DecimalNum(final String val, final MathContext mathContext)` variant and delete constructor without context parameter.

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
sgflt: I would remove this method, because mixing factories is not good thing and this factory methods are covered by `NumFactory.numOf`.

Maybe single constructor that is covered by
```java
   @Override
    public Num numOf(final Number number) {
        return numOf(number.toString());
    }

    @Override
    public Num numOf(final String number) {
        return DecimalNum.valueOf(number, this.mathContext);
    }
```
is sufficient.

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
sgflt: BigDecimal inherits the larger precision. This code is required to handle oprations between constants with precision 1 - 4 and user defined precision.

If we used default precision and user wanted to use smaller precision, resulting precision would be the bigger one which may be unexpected.

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
nimo23: >Maybe single constructor that is covered by

@sgflt ok, but I don't know if we should make this (breaking) change in this PR as this PR is about "**sharing math context**". I think, it would be better to create another PR (after this) with this suggestion. Or is this change required for this PR?

sgflt: It is optional.

TheCookieLab (dismissed):

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
nimo23: >I would remove this method, because mixing factories is not good thing and this factory methods are covered by NumFactory.numOf.

Ok, but then such breaking changes should always be communicated in the `changelog`.

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
nimo23: >It is possible to migrate to DecimalNum(final String val, final MathContext mathContext) variant and delete constructor without context parameter.

Ok, please communicate this in the changelog as well.

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
nimo23: >If we used default precision and user wanted to use smaller precision, resulting precision would be the bigger one which may be unexpected.

I'm not sure about this statement. In general, ta4j defines a **fixed contract on precision** (of `BigDecimal` or `Double`) that must be strictly followed. We shouldn't care what precision the user wants - it should always be the precision defined by ta4j. Even the constants inherit this precision defined by ta4j, which is the correct behavior.

Maybe I don't understand your statement exactly. If so, please explain it with a simple example.

ta4j-core/src/test/java/org/ta4j/core/num/DecimalNumTest.java
```diff
@@ -106,10 +106,10 @@ public void testPowLargeBase() {
...
...167880846302979568155452447609393165219184694556547568563049833746616099084178104706579783956643885189182687944337355411829160776115749158426216371198975733337236000622677975532062859889663984220245029614083499062056412119563695415969672881171577315074977121612997201641494418163124665189863336720090047080384520380024989801688508796247196538755854267152990400356333966457448513513783490093356751626757780404302378718179023586666350805807933966952201429976981679276486187406234131642651827515619233568745746525384050433560149042824676641735456001136242027792933334945829622487781427056361761844835076641985169632599306812063071543550215650354087431074151444189330477914230371357520786575009901912486722028463720021676369386704731297397149687396044449279815084721424308311965770758507105755433087416077040541251678966524883733630056691812581998235914062755215346902830701892895934337167958633033925506277564778413019449268398092637819117919226144175987747886367302787462644684084678362717935369253359416877885443271411157903461506741566545412586156404855949968935042796404405451243990144056514443020139022186704960181336090357132472405537467161045309505652674100300676167988881542412218718624389121"),
```
sgflt, line 109: Precision leaking bug that is well tested.

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
sgflt: I will try to enhance my answer.

User has ability to choose precision in which all calculation within simulation context will be done, correct?

```java
var f = DecimalNumFactory.getInstance(5); // user chooses precision 5 and expects that no calculation will be more precise
```
result is expected to have precision 5:
```java
indicator.getValue().plus(f.ten());
```

But if we provided constant 10 with precision 32, it would lead to leaking this higher precision into overall calculation.
There was exactly this kind of bug that has been well tested :)

See https://github.com/ta4j/ta4j/pull/1246#discussion_r1908181778

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
sgflt: In this PR I would left work as-is, because removal of constuctor is bigger deal. It should be in separate PR that will relate to https://github.com/ta4j/ta4j/issues/1228#issuecomment-2546627972 if we want to hide implementation of numbers.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ Changelog for `ta4j`, roughly following [keepachangelog.com](http://keepachangel
 - Moved `indicators/TripleEMAIndicator` to `indicators/averages`-package
 - Moved `indicators/WMAIndicator` to `indicators/averages`-package
 - Moved `indicators/ZLEMAIndicator` to `indicators/averages`-package
+- Implemented sharing of `MathContext` in `DecimalNum`. For creation of numbers are `NumFactory` implementations the preferred way.
```
nimo23: Please correct english: 

"Implemented sharing of `MathContext` in `DecimalNum`. For creating numbers, `NumFactory` implementations are the preferred way."

ta4j-core/src/main/java/org/ta4j/core/num/DecimalNum.java
```diff
@@ -64,15 +65,21 @@ public final class DecimalNum implements Num {
      *
      * <p>
      * Constructs the most precise {@code Num}, because it converts a {@code String}
-     * to a {@code Num} with a precision of {@link #DEFAULT_PRECISION}; only a
-     * string parameter can accurately represent a value.
+     * to a {@code Num} with a precision at least of {@link #DEFAULT_PRECISION};
```
nimo23: Please correct english:

`"..with a precision of at least {@link #DEFAULT_PRECISION}; .."`

TheCookieLab (approved):

Issue #150, opened 2018-01-07
Confused about Collections.reverse() in CsvTradesLoader.java


daveyx:
Hi,

i am confused about this comment in ta4jexamples.loaders.CsvTradesLoader.loadBitstampSeries() line 85/86
```
// Since the CSV file has the most recent trades at the top of the file, we'll reverse the list to feed the List<Tick> correctly.
Collections.reverse(lines);
```
Looking at the file bitstamp_trades_from_20131125_usd.csv i feel like the most recent trades are at the end of the csv-file, not at the top.

Can you please give me an advice how to remove this confusion in my head?

daveyx:
ok, i missed line 62
`if (beginTime.isAfter(endTime)) `

PR #1251, opened 2025-02-22
#150 Use highest and lowest indicators as instance variables

### LINKED ISSUE #1250: [BUG] Performance degradation when using log4j2
Hi,
`AbstractIndicator` uses `LoggerFactory.getLogger(this.getClass());` to instantiate a `Logger`: when using slf4j-lo4j2 bridge this ends up calling `org.apache.logging.log4j.util.StackLocator.getCallerClass` which is very expensive (jprofiler SS):

Image

The issue can be seen when using the `ParabolicSarIndicator` because the `calculateInternal` method creates new `HighestValueIndicator`/`LowestValueIndicator` indicators in the `calculate` method, thus making it practically unusable.

I would change the `AbstractIndicator` to use a static `Logger`:
`protected static final Logger log = LoggerFactory.getLogger(this.getClass());`

Thanks

sgflt:
- there is no need to create them each execution

Fixes #1250.

- made highest and lowest as instance variables

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/ParabolicSarIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/ParabolicSarIndicator.java
@@ -47,7 +47,9 @@ import org.ta4j.core.num.Num;
 public class ParabolicSarIndicator extends RecursiveCachedIndicator<Num> {
 
     private final LowPriceIndicator lowPriceIndicator;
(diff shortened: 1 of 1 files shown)


TheCookieLab (approved):

TheCookieLab:
@sgflt please run the formatter

TheCookieLab (approved):

PR #1255, opened 2025-04-26
Feature/904 enhance caching layer take 2

TheCookieLab:
Fixes #.

- 
- 
- 


ta4j-core/src/main/java/org/ta4j/core/indicators/CachedIndicator.java
```diff
@@ -23,165 +23,79 @@
...
-                results.remove(0);
-            } else {
-                results.subList(0, nbResultsToRemove).clear();
-            }
+        // 3) else, look up in cache (or compute & cache)
+        T value = cache.get(index);
+        if (value == null) {
+            value = calculate(index);
+            cache.put(index, value);
+        }
```
francescopeloi, line 95:
```suggestion
        T value = cache.computeIfAbsent(index, this::calculate);
```

TheCookieLab: I like the thinking but since we have custom `removeEldestEntry` logic for the LinkedHashMap cache there are some non-obvious gotchas in play. 

LinkedHashMap internally tracks a `modCount` which can be concurrently modified (throwing an exception) when using `computeIfAbsent` but not when manually checking the cache and populating the value (if absent). 

This can happen if `computeIfAbsent` leads to another "nested" `computeIfAbsent` which in turn can update the `modCount`, which can be unexpected from the perspective of the parent `computeIfAbsent`, _especially_ if `removeEldestEntry` triggers positively. This can all happen within the same thread.

I hope this makes sense. TL;DR, making the proposed change results in multiple indicator unit test failures due to ConcurrentModification.

If you have some bandwidth I'd appreciate feedback on this approach vs the "pluggable" cache solution (https://github.com/ta4j/ta4j/pull/907). Is some kind of synthesis approach we should pursue?

francescopeloi: sorry didn't think that what I thought was an innocent change could bring up this mess! Hope I didn't make you lose time, but learn something instead :)

> If you have some bandwidth I'd appreciate feedback on this approach vs the "pluggable" cache solution (https://github.com/ta4j/ta4j/pull/907). Is some kind of synthesis approach we should pursue?

I wish I could say yes to this, but I am swamped at the moment. I will try my best to give a hand, first by reading that long thread.

TheCookieLab: No worries, it was definitely the latter. As for the feedback on the caching approach don't worry about it. It's definitely a prohibitive amount of context to digest if not already initiated.

PR #1257, opened 2025-05-04
Introduce BarPool for memory optimization and bar reuse

jdgriffith:
Implemented a singleton BarPool to reduce memory usage and garbage collection by reusing Bar instances. Updated BaseBar to support reset functionality for reuse and modified relevant components like BaseBarSeries and TimeBarBuilder to integrate with the pool. Added comprehensive tests and documentation for the new feature.

- Add a BaseBar object pool for re-using objects in a high-frequency environment to reduce garbage collection

Since this isn't a "bug", I'm unsure of the template to use here. I'm curious if this would be considered useful by the community at large.

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/BarPool.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BarPool.java
@@ -1,0 +1,99 @@
+package org.ta4j.core;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+import org.ta4j.core.num.Num;
+
+/**
+ * An object pool for {@link Bar} instances to reduce memory usage and garbage collection.
+ * This class maintains a pool of {@link BaseBar} objects that can be reused instead of
+ * creating new instances each time.
+ */
+public class BarPool {
+
+    /** Singleton instance of the BarPool */
+    private static final BarPool INSTANCE = new BarPool();
+
+    /** Map of pools for different time periods */
+    private final Map<Duration, ConcurrentLinkedQueue<BaseBar>> pools = new ConcurrentHashMap<>();
+
+    /** Maximum size for each pool */
+    private static final int MAX_POOL_SIZE = 10_000;
+
+    /**
+     * Private constructor to enforce singleton pattern
+     */
+    private BarPool() {
+    }
+
+    /**
+     * Gets the singleton instance of the BarPool
+     * 
+     * @return the singleton instance
+     */
+    public static BarPool getInstance() {
+        return INSTANCE;
+    }
+
+    /**
+     * Gets a Bar from the pool or creates a new one if none is available
+     * 
+     * @param timePeriod the time period
+     * @param endTime    the end time of the bar period
+     * @param openPrice  the open price of the bar period
+     * @param highPrice  the highest price of the bar period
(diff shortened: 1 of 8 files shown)


ta4j-core/src/main/java/org/ta4j/core/BarPool.java
```diff
@@ -0,0 +1,99 @@
...
+ * This class maintains a pool of {@link BaseBar} objects that can be reused instead of
+ * creating new instances each time.
+ */
+public class BarPool {
+
+    /** Singleton instance of the BarPool */
+    private static final BarPool INSTANCE = new BarPool();
+
+    /** Map of pools for different time periods */
+    private final Map<Duration, ConcurrentLinkedQueue<BaseBar>> pools = new ConcurrentHashMap<>();
```
nimo23, line 22: Imagine two different assets with the same duration. Asset 1 adds bar x1 to the cache, and Asset 2 (with the same duration) uses bar x1 from the cache (= pool) even though it doesn't match that asset. Can this happen with the current implementation? Would it be better to create a BarSeries pool instead?

sgflt:
What problem does solve this PR?

BarSeries is "bar pool" itself if you test multiple strategies in single run.

For live trading there is zero chance to get the same bar twice.

ta4j-core/src/main/java/org/ta4j/core/BarPool.java
jdgriffith: Good question! 

Since bars are agnostic of symbols (assets), I'm not sure it matters 🤷🏻‍♂️ The biggest thing to ensure is the reset of the objects is comprehensive, so there isn't any leaking of data from the previous use of the object. Since the most frequent objects being created and consumed are `Bar`, I decided to start there.

sgflt:
If I get it correctly, you are connected to multiple brokers at the same time and consume events for single asset from multiple sources?

jdgriffith:
Nice question! 

I'm consuming trades from only one data provider, Alpaca SIP. Since I am building bars for most/all equities, there is quite a bit of memory usage and garbage collection from the Bar object creation (no reuse). I currently build out 15-second OHLCV bars/aggregates for each symbol and then track that in a finite BarSeries for analysis. FWIW ta4j has been very effective so far in this endeavour! I'm merely trying to get as much performance as I can from different spots in the flow.

What I'm proposing here is a normal pattern in the high-frequency Java ecosystem of trading with the intention of mitigating GC events and high memory usage spikes.

sgflt:
Now I get it. You want to preallocate constant amount of Bars as a "workspace" and then fill data into any available bar?


I what form are source data coming? Raw strings, json?

I mean the Bar overhead may be very very invisible compared to json parsing or even indicator calculations as they produce a lot of Num objects and Bug Decimals.

Could you profile the impact of change?

jdgriffith:
Source data coming is JSON and I'm using Chronicle Wire with an object pool for JSON objects to prevent GC and memory growth there, too ;) 

Let me see if I can do a custom setup for profiling since this change hasn't been published to maven yet.

TheCookieLab:
> Source data coming is JSON and I'm using Chronicle Wire with an object pool for JSON objects to prevent GC and memory growth there, too ;)
> 
> Let me see if I can do a custom setup for profiling since this change hasn't been published to maven yet.

Agreed, seems promising and I look forward to seeing the profile quantifying the performance gains.

TheCookieLab:
Closing due to inactivity. Feel free to re-open whenever you want to circle back to this.

PR #1258, opened 2025-05-04
Add HFTBar (high-frequency trading) to use atomic types for thread safety

jdgriffith:
Add HFTBar implementation for high-frequency trading use cases

Introduced a thread-safe HFTBar class optimized for multi-threaded environments and high-frequency trading. Included a builder (HFTBarBuilder) and factory (HFTBarBuilderFactory) for creating instances, along with comprehensive test coverage. Adjusted BaseBar to remove thread-safe overhead, distinguishing it from HFTBar.

- Add a new HFTBar (high-frequency) implementation

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/HFTBar.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/HFTBar.java
@@ -1,0 +1,360 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017-2024 Ta4j Organization & respective
+ * authors (see AUTHORS)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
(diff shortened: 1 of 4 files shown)


jdgriffith:
The use case is for a stream of trades coming into an application and being processed by multiple threads. For instance, in my application, I process 15-second aggregates (bars) from trades and then feed those "bars" into ta4j. Granted, I'm doing this for the entire (mostly) equities market via real-time Alpaca data. Ideally, I would like to utilize ta4j in the most performant way for this. Right now, I'm using the approach I added here in custom code to facilitate the least amount of contention possible.

Ideally, there would be less contention this way than using `synchronized` for functions but my understanding could be limited on the topic. This may be a use case that most users of ta4j do not encounter.

sgflt:
Hmm, I would rather introduce specialized HFTBar instead of modifiing current implementation.

jdgriffith:
I added a small performance test and determined there isn't much benefit to this approach. I'm closing this PR. Thanks for all the review and guidance! 

For anyone who might find this in the future, the addition of atomic types does appear to cause more contention and slow-downs. For my use case, I'm going to back track to merely using `synchronized` for specific methods.

image

PR #1260, opened 2025-05-14
RSIIndicator Fix getCountOfUnstableBars

### LINKED ISSUE #1256: RSI indicator should have the unstableBars() equals to barCount
https://github.com/ta4j/ta4j/blob/1101dbe059cda92d7dd1f86e755b0466782911d5/ta4j-core/src/main/java/org/ta4j/core/indicators/RSIIndicator.java#L71

TheCookieLab:
Fixes #1256

- Updated RSIIndicator `getCountOfUnstableBars` to return `barCount` value instead of 0
- Updated RSIIndicator `calculate` to return NaN during unstable period 
- Updated StochasticOscillatorKIndicator constructor to use generic params
- Updated StochasticRSIIndicator to use StochasticOscillatorKIndicator instead of duplicating the logic
- Updated TestUtils assertIndicatorEquals and assertIndicatorNotEquals to handle NaN values


ta4j-core/src/main/java/org/ta4j/core/indicators/RSIIndicator.java
```diff
@@ -50,10 +53,14 @@ public RSIIndicator(Indicator<Num> indicator, int barCount) {
         super(indicator);
         this.averageGainIndicator = new MMAIndicator(new GainIndicator(indicator), barCount);
         this.averageLossIndicator = new MMAIndicator(new LossIndicator(indicator), barCount);
+        this.barCount = barCount;
     }
 
     @Override
     protected Num calculate(int index) {
+        if (index < getCountOfUnstableBars()) {
```
francescopeloi, line 61: Looking at this code, I was wondering if this is a check that should be done on any indicator we have? And if so, should be centralized somewhere

TheCookieLab: This harkens back to https://github.com/ta4j/ta4j/issues/1141 which discusses the nuances and gotchas to such an approach. If we assume for a moment that yes, we do want to centralize this check somewhere then what's the approach?

CachedIndicator.getValue() seems to be a good candidate spot, however it's complicated by the fact that its return type is generic and we have existing subclasses that return Boolean and other non-Num values. To get around this we would need a new NumericCachedIndicator extending CachedIndicator so we only perform this check when the return type can be NaN. Indicators like RSIIndicator would then switch over to extend NumericCachedIndicator. Thoughts on this approach? Alternatives also welcome.

francescopeloi: I had a thought about the problem before reading your idea, and I also came down to the same your `NumericIndicator` idea. Having said that, probably it's a bit of a scope creep for this PR :laughing:

TheCookieLab: I had thought there were more indicators checking and returning NaN during the unstable period but after doing a search it's currently just a sparse handful. I suppose we could also take a more incremental step and return 0 instead of NaN, or even not perform a check in `calculate` at all. Thoughts?

Issue #1272, opened 2025-06-01
ProfitLossPercentageCriterion doubts


francescopeloi:
**I have checked existing issues and wiki**
- [x] I could not find similar [issues](https://github.com/ta4j/ta4j/issues?utf8=%E2%9C%93&q=)
- [x] I could not find a solution in the [wiki](https://ta4j.github.io/ta4j-wiki/) or [faq section](https://ta4j.github.io/ta4j-wiki/FAQ.html)

ProfitLossPercentageCriterion is doing:

![formula](https://latex.codecogs.com/png.image?\bg_white%20ProfitLossPercentageCriterion=\sum_{i=1}^{n}\frac{profit_i}{entryValue_i}\times100)

but percentages are not additive: 12 % + 3 % on 2 positions does not equal 15 % portfolio return.

I think we should be doing instead:

![formula](https://latex.codecogs.com/png.image?\bg_white%20ProfitLossPercentageCriterion=\frac{\sum_{i=1}^{n}profit_i}{\sum_{i=1}^{n}entryValue_i}\times100)

I am not sure if this metric was built to measure something else, that I am missing.

Examples:
| Trade | Cost | Profit | % (profit / cost) |
|-------|------|--------|-------------------|
| 1     | 100  | +10    | +10 % |
| 2     | 100  |  –5    |  –5 % |

Old additive criterion: +10 % + (–5 %) = **+5 %**  
Proposed capital weighted: ( +10 – 5 ) / 200 × 100 = **+2.5 %**

| Trade | Cost | Profit | % (profit / cost) |
|-------|------|--------|-------------------|
| 1     | 10   | +1     | +10 % |
| 2     | 1000 | +20    | +2 % |

Old additive criterion: +10 % + 2 % = **+12 %**  
Proposed capital weighted: ( +1 + 20 ) / 1010 × 100 ≈ **+2.08 %**

PR #1275, opened 2025-06-03
Refactor all PnL criteria to have Net and Gross versions

francescopeloi:
More fixes for #1272

After I found another wrong variable name naming gross profit when instead it was net profit (AverageProfitCriterion), I thought that the current naming convention is not working, so I did:
- centralised pnl logic
- created classes to calculate net and gross pnl, so it's very clear from the class name, instead of one class calculating both (with confusions)
- kept the original pnl classes for backward compatibility (I can delete them if we don't want them)


francescopeloi:
this requires some test fixing that I am doing, also I am going to follow this approach for all PnL criteria, so also:
- AverageLoss
- AverageProfit
- ProfitLossPercentage
- ProfitLossRatio
- Return

francescopeloi:
will open a new PR

PR #1291, opened 2025-06-27
Fix aggregation of bars

### LINKED ISSUE #1290: [BUG] VolumeBarBuilder and TickBarBuilder: aggregate all cumulative fields
**Describe the bug**
The `VolumeBarBuilder` aggregates by `volume` and correctly creates a new bar after `volumeThreshold` is reached. However, while sampling, it should also aggregate its `amount` and `trades`. The same applies to the `TickBarBuilder`.

Instead

https://github.com/ta4j/ta4j/blob/a5eb46ffdfcfa78a11906bb3d66b509c4918b421/ta4j-core/src/main/java/org/ta4j/core/bars/VolumeBarBuilder.java#L186

it should aggregate:

`this.amount = this.amount.plus(amount);`

**Expected behavior**
Cumulative fields (volume, amount, trades) should always be aggregated while sampling.

sgflt:
Fixes #1290.

- sum up trades and amount

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/bars/TickBarBuilder.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/bars/TickBarBuilder.java
@@ -141,25 +141,25 @@ public class TickBarBuilder implements BarBuilder {
 
     @Override
     public BarBuilder closePrice(final Num tickPrice) {
-        this.closePrice = tickPrice;
-        if (this.openPrice == null) {
-            this.openPrice = tickPrice;
+        closePrice = tickPrice;
+        if (openPrice == null) {
+            openPrice = tickPrice;
         }
 
-        this.highPrice = this.highPrice.max(tickPrice);
-        this.lowPrice = this.lowPrice.min(tickPrice);
+        highPrice = highPrice.max(tickPrice);
+        lowPrice = lowPrice.min(tickPrice);
 
         return this;
     }
 
     @Override
     public BarBuilder closePrice(final Number closePrice) {
-        return closePrice(this.numFactory.numOf(closePrice));
+        return closePrice(numFactory.numOf(closePrice));
     }
 
     @Override
     public BarBuilder closePrice(final String closePrice) {
-        return closePrice(this.numFactory.numOf(closePrice));
+        return closePrice(numFactory.numOf(closePrice));
     }
 
     @Override
@@ -182,7 +182,7 @@ public class TickBarBuilder implements BarBuilder {
 
     @Override
     public BarBuilder amount(final Num amount) {
-        this.amount = this.amount.plus(amount);
+        this.amount = this.amount == null ? amount : this.amount.plus(amount);
         return this;
     }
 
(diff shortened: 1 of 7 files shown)


ta4j-core/src/main/java/org/ta4j/core/bars/TickBarBuilder.java
```diff
@@ -242,6 +242,8 @@ private void reset() {
         this.highPrice = zero;
         this.lowPrice = this.numFactory.numOf(Integer.MAX_VALUE);
         this.closePrice = null;
+        this.amount = this.numFactory.zero();
```
nimo23: you can use 

`this.amount = zero;`

ta4j-core/src/main/java/org/ta4j/core/bars/VolumeBarBuilder.java
```diff
@@ -235,6 +235,7 @@ public void add() {
             var volumeRemainder = this.numFactory.zero();
             if (this.volume.isGreaterThan(this.volumeThreshold)) {
                 volumeRemainder = this.volume.minus(this.volumeThreshold);
+                // cap currently built bar, volume is then restored to volumeRemainder
                 this.volume = this.volumeThreshold;
```
nimo23: What about the remainder of `amount` and `trades`? The remainder of `amount` and `trades` should also be added to the next bar and subtracted from the current bar. It should also be handled in`TickBarBuilder`. 

And also add appropriate test cases (i.e. test whether the remainder for `amount` and `trades` is handled correctly).

sgflt: Hm. It would be simpler to just test whether volume exceeds threshold (assuming remiander is << smaller than threshold), use it as whole and then reset all variables to initial state, because there is no such amount threshold that can be used for remainder calculation. It's all or nothing.

nimo23: > ..because there is no such amount threshold that can be used for remainder calculation. It's all or nothing.

I thought about the formula `amount = closePrice * volume` which can be used to calculate the relative remainder for amount. For example:

```
if (this.volume.isGreaterThan(this.volumeThreshold)) {
volumeRemainder = this.volume.minus(this.volumeThreshold);
// cap currently built bar, volume is then restored to volumeRemainder
this.volume = this.volumeThreshold;

amountRemainder = closePrice * volumeRemainder
this.amount = amount - amountRemainder;
}
```

But I don't know if this makes sense. It would probably only be an approximation..what do you think?

sgflt: Who is supposed to fill amount as independent unit? Does make sense remove the amount field and calculate amount ad-hoc in getter? It would solve this dilema and delay calculation to probably more correct place.

sgflt: As reference https://docs.alpaca.markets/docs/real-time-stock-pricing-data, they return price and volume. Amount is redundant information.

nimo23: >Does make sense remove the amount field and calculate amount ad-hoc in getter? It would solve this dilema and delay calculation to probably more correct place.

@sgflt Please calculate the `amount` within the `add()`-method of `VolumeBarBuilder`, `TickBarBuilder` and `TimeBarBuilder` **only if** `amount` was `null` (i.e. if the user does not set it explicitly). 

Some market providers also provide the `amount` field, so the user should be allowed to explicitly set it. 

And: After merging this PR, I will create another bar type (`DollarBar` with an `amountThreshold` instead of `volumeThreshold`) for which the `amount` field must be set.

TheCookieLab (approved):

PR #1296, opened 2025-06-29
add CriterionIndicator

nimo23:
Fixes https://github.com/ta4j/ta4j/issues/742

- Add CriterionIndicator

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/CriterionIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/CriterionIndicator.java
@@ -77,6 +77,9 @@ public final class CriterionIndicator extends CachedIndicator<Boolean> {
             // consider only the positions made till index
             if (entry.getIndex() <= index) {
                 subTradingRecord.enter(entry.getIndex(), entry.getNetPrice(), entry.getAmount());
+            } else {
+                // no need to iterate positions made after the index
+                break;
             }
             if (exit != null && exit.getIndex() <= index) {
                 subTradingRecord.exit(exit.getIndex(), exit.getNetPrice(), exit.getAmount());
--- a/ta4j-core/src/test/java/org/ta4j/core/indicators/CriterionIndicatorTest.java
+++ b/ta4j-core/src/test/java/org/ta4j/core/indicators/CriterionIndicatorTest.java
@@ -82,9 +82,9 @@ public class CriterionIndicatorTest extends AbstractIndicatorTest<Indicator<Num>
         criterionIndicator = new CriterionIndicator(series, tradingRecord, numberOfWinningPositionsCriterion,
                 requiredCriterionValue);
 
-        // till index 0, no winning position were made
+        // till index 0, no winning position was made
         assertFalse(criterionIndicator.getValue(0));
-        // till index 1, no winning position were made
+        // till index 1, no winning position was made
         assertFalse(criterionIndicator.getValue(1));
         // till index 2, at least one winning position was made
         assertTrue(criterionIndicator.getValue(2));


ta4j-core/src/main/java/org/ta4j/core/indicators/CriterionIndicator.java
```diff
@@ -0,0 +1,96 @@
...
+            if (entry.getIndex() <= index) {
+                subTradingRecord.enter(entry.getIndex(), entry.getNetPrice(), entry.getAmount());
+            }
+            if (exit != null && exit.getIndex() <= index) {
+                subTradingRecord.exit(exit.getIndex(), exit.getNetPrice(), exit.getAmount());
+            }
+        }
+
+        var calculatedCriterionValue = criterion.calculate(getBarSeries(), subTradingRecord);
+        return criterion.betterThan(calculatedCriterionValue, requiredCriterionValue)
```
francescopeloi, line 90: why not returning the criterion value directly, instead of making this bit opinionated? i.e. let the caller decide what to do with the value

ta4j-core/src/main/java/org/ta4j/core/indicators/CriterionIndicator.java
```diff
@@ -0,0 +1,96 @@
...
+                tradingRecord.getTransactionCostModel(), tradingRecord.getHoldingCostModel());
+
+        for (var position : tradingRecord.getPositions()) {
+            var entry = position.getEntry();
+            var exit = position.getExit();
+
+            // consider only those positions made till index
+            if (entry.getIndex() <= index) {
+                subTradingRecord.enter(entry.getIndex(), entry.getNetPrice(), entry.getAmount());
+            }
```
francescopeloi, line 83: if you are in the `else` I think you could `break`, to avoid scanning the whole trading record. This assumes the trading record is ordered, which I believe it is? Same in the else below.

ta4j-core/src/main/java/org/ta4j/core/indicators/CriterionIndicator.java
nimo23: In this solution, it doesn't make sense to return the value as is. If the user wants to get the value, they should simply use the `Criterion` rather than the `CriterionIndicator`. A Boolean indicator combined with a criterion is more convenient for the user. What can the user do with the value other than compare it with `betterThan`? By the way, returning the value instead would require the user to then instantiate a new instance of the `Citerion` and call `betterThan`.

ta4j-core/src/main/java/org/ta4j/core/indicators/CriterionIndicator.java
nimo23: >This assumes the trading record is ordered, which I believe it is?

Yes, _normally_ it should be ordered. I added a `break` statement to this code.

sgflt:
I am late to party... but for purpose of data mining I have introduced `RuntimeContext`. It is a space for data , that does not come from Ta4j, but from other sources like broker, local database and so on.

I just store events that are useful (f.e. XTB returned profit of position) and it was stored within RuntimeContext. StrategyFactory accepts `IndicatorContexts` - place where indicators for all timerames are held, `RuntimeContext` - user defined implementation that may store anything. 

Strategy then has just `isSatisfied()` method without parameters and may lookup through string keys or typesafe resolvers into `RuntimeContext` for measured value.

I have created for demonstration usage of AI service:
https://github.com/sgflt/ta4k/blob/master/ta4j-examples/src/main/kotlin/ta4jexamples/strategies/ai/AITradingStrategyFactory.kt

PR #1299, opened 2025-07-02
added AmountBarBuilder

nimo23:
Fixes #.

- added `AmountBarBuilder`

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/bars/AmountBarBuilder.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/bars/AmountBarBuilder.java
@@ -1,0 +1,285 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017-2025 Ta4j Organization & respective
+ * authors (see AUTHORS)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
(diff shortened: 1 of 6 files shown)


ta4j-core/src/main/java/org/ta4j/core/bars/AmountBarBuilderFactory.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param amountThreshold   the threshold at which a new bar should be created
+     * @param setAmountByVolume if {@code true} the {@code amount} is set by
+     *                          {@code volume * closePrice}, otherwise
+     *                          {@code amount} must be explicitly set
+     */
+    public AmountBarBuilderFactory(final int amountThreshold, final boolean setAmountByVolume) {
```
sgflt, line 44: As in other classes, wouldn't be enum mode more descriptive than boolean?

```java
enum AccumulationMode {
  MANUAL, 
  AUTO,
  ;
}

Comment (nimo23) [reply ta4j-core/src/main/java/org/ta4j/core/bars/AmountBarBuilderFactory.java:44]: I think `setAmountByVolume` is more descriptive than `AccumulationMode.AUTO`: what does `AUTO` mean without reading a javadoc? Reading `setAmountByVolume` is more descriptive.

Furthermore, the `setAmountByVolume` argument always has a status of `true` or `false` - so I'll **keep it simple** and use a `boolean` value instead of an `enum`.

File: ta4j-core/src/test/java/org/ta4j/core/bars/AmountBarBuilderTest.java [hunk 2]
```diff
@@ -0,0 +1,271 @@
...
+import java.time.Duration;
+import java.time.Instant;
+
+import org.junit.Test;
+import org.ta4j.core.BaseBarSeriesBuilder;
+import org.ta4j.core.num.DecimalNumFactory;
+
+public class AmountBarBuilderTest {
+
+    @Test(expected = IllegalArgumentException.class)
```
sgflt: I would preffer JUnit5 for new tests.

ta4j-core/src/main/java/org/ta4j/core/bars/AmountBarBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+    public Bar build() {
+        return new BaseBar(timePeriod, beginTime, endTime, openPrice, highPrice, lowPrice, closePrice, volume, amount,
+                trades);
+    }
+
+    @Override
+    public void add() {
+
+        if (setAmountByVolume) {
+            var calculatedAmount = closePrice.multipliedBy(distinctVolume);
```
sgflt: I would use final for variables, that are assigned only once. It reduces amount of  surprise moments.

ta4j-core/src/test/java/org/ta4j/core/bars/AmountBarBuilderTest.java
```diff
@@ -30,14 +30,15 @@
...
+import org.junit.jupiter.api.Assertions;
 import org.ta4j.core.BaseBarSeriesBuilder;
 import org.ta4j.core.num.DecimalNumFactory;
 
 public class AmountBarBuilderTest {
 
-    @Test(expected = IllegalArgumentException.class)
-    @SuppressWarnings("unused")
+    @Test
     public void createBarsWithSetAmountByVolume() {
```
sgflt: Aaaand JUnit5 supports package private test methods and classes, we can remove public modifiers.

sgflt (approved):

TheCookieLab (approved):

Issue #1216, opened 2024-11-10
Allow mutation of the last Bar


ek-ex:
This is a spinoff issue of #1211 to discuss mutability of the bars solely. 

For context the brokers and data providers can be updating bars real time even when the bar is still open, multiples times per second in many cases. Additionally, the trading bots have one thread dedicated of fetching live updates (the data thread) and other multiple threads to compute indicators, execute trades, manage internal state, and other actions. So the live thread can be updating the Bars while the trades are being executed and indicators computed. Having hard synchronization is not desired because you don't want to block the live thread. If there is a data change, the live thread should be able to update the internal state and move to the next update.

Currently, in order to keep the BarSeries updated one must call the `BarSeries.add(Bar, replace:true)` which internally calls `bars.set(this.bars.size() - 1, bar)`. Replacing the Bar in the ArrayList is problematic:

- Can have ConcurrentModificationException if the list is being updated while being updated. 
- Any preexisting information in the last Bar will be discarded, like Trades or Amount as result of a live trade in this bar. 
- Is computationally more expensive creating new bars and replacing in the list versus updating individual members in the Bar.

A better alternative would be to allow Bar mutation when the bar is still open. Maybe the BarSeries can allow mutating the last bar of the series when the date of the last bar equals the date of the bar requested. Something else? 

Happy to contribute a PR if the maintainers agree on the approach.

ek-ex:
I discovered that the `BaseBar` already allows somewhat a form of mutation so this would partially solve my issues with having to create a new bar and replacing it in the list. I just need to call this method 4 times, in the exact order for OHLC. 

https://github.com/ta4j/ta4j/blob/42d93197fb6851fe23fbd863c1843fdb9844f47e/ta4j-core/src/main/java/org/ta4j/core/BaseBar.java#L164-L176

The missing piece now is that I dont have a mechanism to update the volume in the bar. Maybe the solution to this issues is to allow the mutation of the volume in a similar way to something like this:

```java
@Override
public void addVolume(Num tradeVolume) {
    this.volume = this.volume.plus(tradeVolume);
}

@Override
public void setVolume(Num tradeVolume) {
    this.volume = tradeVolume;
}
```

francescopeloi:
How would you mutate it? From a quick look, close price aside, it sets the prices only if they weren't set before (the null check)

ek-ex:
[...]

PR #1302, opened 2025-07-09
Add convenient methods to BarSeries

nimo23:
Fixes https://github.com/ta4j/ta4j/issues/1216

- added convenient methods to `BarSeries.java` and `BarSeriesUtils.java`
- Moved `BarSeriesUtils.java` to the root folder (so that it is next to `BarSeries.java`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,16 @@ See also: the [contribution policy](.github/CONTRIBUTING.md) and [Code of Conduc
 &nbsp;
 &nbsp;
 
+### Powered by
+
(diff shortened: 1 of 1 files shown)


TheCookieLab:
Keeping this open. The live-feed bar update use case from #1216 is valid and we should support a deterministic append/update path.

I likely will re-implement this from seed rather than merge the branch directly. Planned divergences:
1. Keep BarSeriesUtils in org.ta4j.core.utils (no package move).
2. Keep core additions: addLastBar behavior and replaceBarsIfChanged helper.
3. Remove wall-clock dependent branching (no Instant.now gating); decisions should be series/bar-time based.
4. Rework tests around append/update/replace semantics and verify copy(name, max) behavior if that API is kept.

Net: moving forward on the idea, but with a narrower and more stable API footprint.

PR #1326, opened 2025-09-27
correct calculation of unstable bars of SMA indicator

tab269:
- the SMA indicator (calculating on another indicator) should return the unstable bars of the indicator the sma is calculating on plus its own unstable bars

Fixes #.

- 
- 
- 

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/averages/SMAIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/averages/SMAIndicator.java
@@ -37,7 +37,8 @@ import org.ta4j.core.num.Num;
 public class SMAIndicator extends CachedIndicator<Num> {
 
     private final int barCount;
-    private RunningTotalIndicator previousSum;
+    private final Indicator<Num> indicator;
+    private final RunningTotalIndicator previousSum;
 
     /**
      * Constructor.
@@ -48,6 +49,7 @@ public class SMAIndicator extends CachedIndicator<Num> {
     public SMAIndicator(Indicator<Num> indicator, int barCount) {
         super(indicator);
         this.previousSum = new RunningTotalIndicator(indicator, barCount);
+        this.indicator = indicator;
         this.barCount = barCount;
     }
 
@@ -65,7 +67,7 @@ public class SMAIndicator extends CachedIndicator<Num> {
     /** @return {@link #barCount} */
     @Override
     public int getCountOfUnstableBars() {
-        return barCount;
+        return indicator.getCountOfUnstableBars() + barCount;
     }
 
     @Override
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/averages/SMAIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/averages/SMAIndicator.java
@@ -67,7 +67,7 @@ public class SMAIndicator extends CachedIndicator<Num> {
     /** @return {@link #barCount} */
     @Override
     public int getCountOfUnstableBars() {
-        return indicator.getCountOfUnstableBars() + barCount;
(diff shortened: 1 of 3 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/averages/SMAIndicator.java
```diff
@@ -65,7 +67,7 @@ private Num partialSum(int index) {
     /** @return {@link #barCount} */
     @Override
     public int getCountOfUnstableBars() {
-        return barCount;
+        return indicator.getCountOfUnstableBars() + barCount;
```
francescopeloi: should it be a sum, or would it be better to be a `Integer.max(indicator.getCountOfUnstableBars(), barCount)`?

tab269: Thank you @francescopeloi for making me to rethink my proposed solution. But after some thinking I am still convinced it should be a sum (but reduced by one).
What would we gain with the max function?
Here is the example I first came over the problem:
Take a STOCHASTIC (e.g. 14 lookback bars = 13 unstable bars) indicator and do an SMA (e.g. 3 lookbar bars) of it. The SMA has some unstable bars because of its own calculations (2 in this case). But since it is operating on the output bars of the STOCHASTIC indicator, the calculation should not start with unstable bars of this guy, so the earliest bar the SMA can start calculating the average on is the first stable bar of STOCHASTIC (0..12 = 13 are unstable so no.13 is the first stable one). Here we go with the sum. The resulting bar series (being output of the chain of STOCHASTIC and SMA) should be stable from bar no. 13 + 2 = 15.
It might be different for other indicators, I have not thought it through, yet, but for the SMA I am pretty sure.
Would you vote for me writing a test for that?

tab269:
OK, tests make it clearer and often correct imperfect thinking ;)
It is slightly different from my originally proposed solution: Since we include the current bar it the calculation, it is not part of the unstable bars. So the implementation for `getCountOfUnstableBars()` is `indicator.getCountOfUnstableBars() + barCount - 1`.

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/SMAIndicator.java
francescopeloi: My question was more because I had never fully understood what to use. Your example explained it to me well, thank you @tab269! `sum` seems legit in this case since the unstable bars of the inner indicator are carried forward into the outer one. I guess if an indicator inside is doing something like (indicator1 OR indicator2), where they don’t actually stack, then `max` would probably be more appropriate.

francescopeloi:
> Since `getCountOfUnstableBars()` is describing negative logic I would suggest to introduce a method called `getIndexOfFirstStableBar()` (describes positive logic). What do you think?

I have the feeling that something like that would be useful in some of my use cases, but I'd need to double check. Maybe discussion for a different issue though.

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/SMAIndicator.java
tab269: @francescopeloi yes, max might be appropriate for some indicators. But I guess we need to test this in detail first ;)

tab269 (commented):
resolved merge conflict

ta4j-core/src/main/java/org/ta4j/core/indicators/averages/SMAIndicator.java
TheCookieLab: @francescopeloi are your concerns resolved here?

francescopeloi: yes

TheCookieLab (approved):

PR #1350, opened 2025-10-19
Add TrueStrengthIndex, SchaffTrendCycle, and ConnorsRSI indicators

TheCookieLab:
- **EMA indicators now return NaN during unstable period**: `EMAIndicator`, `MMAIndicator`, and all indicators extending `AbstractEMAIndicator` now return `NaN` for indices within the unstable period (indices < `beginIndex + getCountOfUnstableBars()`). Previously, these indicators would return calculated values during the unstable period. 
- `DifferencePercentageIndicator` has been deprecated in favor of `PercentageChangeIndicator`, which now provides all the same functionality plus additional features.
- Added `TrueStrengthIndexIndicator`, `SchaffTrendCycleIndicator`, and `ConnorsRSIIndicator` to expand oscillator coverage
- Added `PercentRankIndicator` helper indicator to calculate the percentile rank of a value within a rolling window
- Added `DifferenceIndicator` helper indicator to calculate the difference between current and previous indicator values
- Added `StreakIndicator` helper indicator to track consecutive up or down movements in indicator values
- Added `StochasticIndicator` as a generic stochastic calculation indicator

### DIFF
--- a/.github/workflows/release-scheduler.yml
+++ b/.github/workflows/release-scheduler.yml
@@ -12,7 +12,7 @@ name: AI Semantic Release Scheduler
 
 concurrency:
   group: semantic-release-ai-${{ github.repository }}
-  cancel-in-progress: false
+  cancel-in-progress: true
 
 on:
   schedule:
@@ -460,7 +460,7 @@ jobs:
       }}
     steps:
       - name: Trigger publish workflow
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
@@ -495,4 +495,4 @@ jobs:
           echo "  bump:    ${{ needs.analyze.outputs.bump }}"
           echo "  version: ${{ needs.analyze.outputs.version }}"
           echo ""
-          echo "(Dry run enabled - skipping release trigger)"
+          echo "(Dry run enabled - propagating dryRun=true to release.yml)"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,6 +51,76 @@ jobs:
           echo "audit:dry_run_normalized=$dry_run"
           echo "dryRun=$dry_run" >> $GITHUB_OUTPUT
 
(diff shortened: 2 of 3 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/ConnorsRSIIndicator.java
```diff
@@ -0,0 +1,229 @@
...
+            int valid = 0;
+            int lessThanCount = 0;
+            for (int i = startIndex; i <= endIndex; i++) {
+                Num candidate = indicator.getValue(i);
+                if (isNaN(candidate)) {
+                    continue;
+                }
+                valid++;
+                if (candidate.isLessThan(current)) {
+                    lessThanCount++;
```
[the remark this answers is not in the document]
TheCookieLab: 👍

ta4j-core/src/main/java/org/ta4j/core/indicators/TrueStrengthIndexIndicator.java
```diff
@@ -0,0 +1,225 @@
...
+            return current.minus(previous);
+        }
+
+        @Override
+        public int getCountOfUnstableBars() {
+            return 1;
+        }
+    }
+
+    private static final class AbsoluteValueIndicator extends CachedIndicator<Num> {
```
francescopeloi: would these inner indicators be valuable as independent indicators?

TheCookieLab: Yes, worth another look at the very least. Originally the thought was we needed slightly opinionated versions of these inner indicators but turns out not to be so. In fact some of these can likely be replaced with existing altogether.

TheCookieLab: All inner indicators have either been extracted or replaced by existing. This actually forced a (long overdue) reckoning with how NaNs are propagated in the EMA based indicators.

Issue #1370, opened 2025-11-12
Feature Request: Add Sharpe Ratio and Sortino Ratio Criteria


naveenairani:
**I have checked existing issues and wiki**
- [x] I could not find similar [issues](https://github.com/ta4j/ta4j/issues?utf8=✓&q=)
- [x] I could not find a solution in the [wiki](https://ta4j.github.io/ta4j-wiki/) or [faq section](https://ta4j.github.io/ta4j-wiki/FAQ.html)

Currently, ta4j lacks built-in Sharpe Ratio and Sortino Ratio criteria. Users must implement these risk-adjusted metrics manually, leading to inconsistent calculations, error-prone implementations, and maintenance burden.

**Solution**
Add standardized `SharpeRatioCriterion` and `SortinoRatioCriterion` classes with:
- Configurable annual risk-free rate
- Automatic period detection from BarSeries  
- Proper compounding for different frequencies
- Bounded, reasonable output values

**Alternatives**
Manual implementation, but this causes calculation errors (especially with risk-free rate handling and annualization) and makes strategy comparisons unreliable.

**Additional context**
These are essential portfolio theory metrics needed for professional strategy comparison, risk management, and portfolio optimization. Standard implementation would ensure accuracy and comparability across all ta4j users.

naveenairani:
Thank you @francescopeloi for sharing your implementation! I'd be happy to help test these criteria in our environment and provide feedback once the draft PR is ready. Looking forward to reviewing the implementation.

francescopeloi:
@naveenairani sharpe ratio has been released

PR #1371, opened 2025-11-12
Sortino and Sharpe ratio

francescopeloi:
AI generated version of Sortino and Sharpe ratio. These are not ready as I haven't reviewed what they do exactly and there are no tests nor javadocs, but putting them out there as they've been asked in #1370 

- I don't recall why I did a per bar version of Sortino
- they have been made in a functional style which probably makes it harder to understand what they do 
- I recall that there were some subtle choices to make implementation wise that might make them different from the canonical implementation. Don't ask me what because I don't remember :smile:

### DIFF
--- a/ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
@@ -115,7 +115,10 @@ public class SharpeRatioCriterion extends AbstractAnalysisCriterion {
         if (seconds == 0) {
             return numFactory.zero();
         }
-        var deltaYears = seconds / SECONDS_PER_YEAR;
+        var seconds = Duration.between(endPrev, endNow).toNanos() / 1_000_000_000.0;
+        Num deltaYearsNum = numFactory.numOf(seconds).dividedBy(numFactory.numOf(SECONDS_PER_YEAR));
+        Num per = riskFreeRate.plus(numFactory.one()).pow(deltaYearsNum).minus(numFactory.one());
+        return per;
         var annual = riskFreeRate.doubleValue();
         var per = Math.pow(1.0 + annual, deltaYears) - 1.0;
         return numFactory.numOf(per);


ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
```diff
@@ -0,0 +1,175 @@
...
+    @Override
+    public Num calculate(BarSeries series, TradingRecord tradingRecord) {
+        var numFactory = series.numFactory();
+        var zero = numFactory.zero();
+        var closedPositions = tradingRecord.getPositions().stream().filter(Position::isClosed).toList();
+        if (closedPositions.isEmpty()) {
+            return zero;
+        }
+
+        var cashFlow = new CashFlow(series, tradingRecord);
```
TheCookieLab, line 84: Can this `calculate` call the position-based one rather than duplicating much of the same logic?

ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
```diff
@@ -0,0 +1,175 @@
...
+
+    private static final double SECONDS_PER_YEAR = 365.2425d * 24 * 3600;
+    private final Num riskFreeRate;
+
+    public SharpeRatioCriterion(Num riskFreeRateValue) {
+        this.riskFreeRate = riskFreeRateValue;
+    }
+
+    public SharpeRatioCriterion() {
+        this.riskFreeRate = DecimalNumFactory.getInstance().zero();
```
TheCookieLab, line 46: Hmm directly coupling to DecimalNumFactory here seems like it's just asking for subtle & obscure bugs down the line. Ideally we consistently enforce the same num factory (from the bar series) everywhere and not provide opportunity for divergence.

ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
```diff
@@ -0,0 +1,175 @@
...
+
+    private Num periodRiskFree(BarSeries series, int i) {
+        var numFactory = series.numFactory();
+        var endNow = series.getBar(i).getEndTime();
+        var endPrev = series.getBar(i - 1).getEndTime();
+        var seconds = Math.max(0, Duration.between(endPrev, endNow).getSeconds());
+        if (seconds == 0) {
+            return numFactory.zero();
+        }
+        var deltaYears = seconds / SECONDS_PER_YEAR;
```
TheCookieLab:
```suggestion
        var seconds = Duration.between(endPrev, endNow).toNanos() / 1_000_000_000.0;
        Num deltaYearsNum = numFactory.numOf(seconds).dividedBy(numFactory.numOf(SECONDS_PER_YEAR));
        Num per = riskFreeRate.plus(numFactory.one()).pow(deltaYearsNum).minus(numFactory.one());
        return per;
```
Slight tweak, on the chance someone is using sub-second bars this won't truncate (ex. 500ms to 0). Also rather going from Num to double to Num, prefer doing all the math via Num.

ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
```diff
@@ -0,0 +1,175 @@
...
+                .reduce(Stats.empty(zero), (acc, x) -> acc.add(x, numFactory), (a, b) -> a.merge(b, numFactory));
+
+        if (stats.count() < 2) {
+            return zero;
+        }
+        var stdev = stats.sampleVariance(numFactory).sqrt();
+        return stdev.isZero() ? zero : stats.mean().dividedBy(stdev);
+    }
+
+    private Num excessReturn(BarSeries series, CashFlow cashFlow, int i) {
```
TheCookieLab, line 102: We should probably call out in the javadoc if this Sharpe is meant to be annualized or not. The risk free calculation is annualized, however the return distribution isn't so there's ambiguity on the intentions. Actually it'd be great if we supported both, something like. 

```
new SharpeRatioCriterion(Annualization.ANNUALIZED)
new SharpeRatioCriterion(Annualization.PERIOD)
```

ta4j-core/src/main/java/org/ta4j/core/criteria/SortinoRatioByBarCriterion.java
```diff
@@ -0,0 +1,137 @@
...
+
+    private Num average(List<Num> data, BarSeries series) {
+        var sum = data.stream().reduce(series.numFactory().zero(), Num::plus);
+        return sum.dividedBy(series.numFactory().numOf(data.size()));
+    }
+
+    private Num downsideDeviation(List<Num> data, BarSeries series) {
+        var numFactory = series.numFactory();
+        var negatives = data.stream()
+                .filter(r -> r.isLessThanOrEqual(targetReturn)) // include zeros
```
TheCookieLab, line 124: should we use strictly less than here?

ta4j-core/src/main/java/org/ta4j/core/criteria/SortinoRatioByBarCriterion.java
```diff
@@ -0,0 +1,137 @@
...
+                    while (current != null && i > current.getExit().getIndex()) {
+                        idx += 1;
+                        current = idx < positions.size() ? positions.get(idx) : null;
+                    }
+                    var prev = series.getBar(i - 1).getClosePrice();
+                    var now = series.getBar(i).getClosePrice();
+                    var ratio = (current != null && i > current.getEntry().getIndex())
+                            ? current.getEntry().isBuy() ? now.dividedBy(prev) : prev.dividedBy(now)
+                            : one;
+                    pair.returns().add(ratio.minus(one));
```
TheCookieLab, line 107: This mutates the `pair` no? This `reduce` as a whole is pretty gnarly. Any way we can simplify?

ta4j-core/src/main/java/org/ta4j/core/criteria/SortinoRatioByBarCriterion.java
```diff
@@ -0,0 +1,137 @@
...
+
+    private Num downsideDeviation(List<Num> data, BarSeries series) {
+        var numFactory = series.numFactory();
+        var negatives = data.stream()
+                .filter(r -> r.isLessThanOrEqual(targetReturn)) // include zeros
+                .toList();
+        if (negatives.isEmpty())
+            return numFactory.zero();
+        var arr = negatives.stream().mapToDouble(n -> targetReturn.minus(n).doubleValue()).toArray();
+        return numFactory.numOf(new StandardDeviation(true).evaluate(arr));
```
TheCookieLab, line 129: Should we use the population SD (false) here?

ta4j-core/src/main/java/org/ta4j/core/criteria/SortinoRatioByBarCriterion.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private Num downsideDeviation(List<Num> data, BarSeries series) {
+        var numFactory = series.numFactory();
+        var negatives = data.stream()
+                .filter(r -> r.isLessThanOrEqual(targetReturn)) // include zeros
+                .toList();
+        if (negatives.isEmpty())
+            return numFactory.zero();
+        var arr = negatives.stream().mapToDouble(n -> targetReturn.minus(n).doubleValue()).toArray();
```
TheCookieLab, line 128: I'm going to challenge you to keep all this math in Num world 😝

TheCookieLab (changes requested):
Looks promising. Please add unit tests to demonstrate correctness and serve as a bulwark against future regressions.

PR #1406, opened 2026-01-09
Sharpe ratio

francescopeloi:
- adds SharpeRatioCriterion 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
   a dedicated series (`ConcurrentBarSeries`/builder), realtime bar model (`RealtimeBar`/`BaseRealtimeBar`), and
   streaming-bar ingestion helpers to enable candle reconciliation and side/liquidity-aware trade aggregation.
 - **Sharpe Ratio**: Added `SharpeRatioCriterion` and its related calculation classes
+- Added **ThreeInsideUpIndicator** and **ThreeInsideDownIndicator**
+- Added **MorningStarIndicator** and **EveningStarIndicator**
+- Added **BullishKickerIndicator** and **BearishKickerIndicator**
+- Added **PiercingIndicator** and **DarkCloudIndicator**
 
 ### Changed
 - **Bar builders null handling**: Bar builders now skip null-valued bars entirely instead of inserting placeholder/null bars, leaving gaps when inputs are missing or invalid.
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/candles/BearishKickerIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/candles/BearishKickerIndicator.java
@@ -1,0 +1,88 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017-2025 Ta4j Organization & respective
+ * authors (see AUTHORS)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package org.ta4j.core.indicators.candles;
+
+import org.ta4j.core.Bar;
+import org.ta4j.core.BarSeries;
+import org.ta4j.core.indicators.CachedIndicator;
+import org.ta4j.core.indicators.trend.UpTrendIndicator;
+import org.ta4j.core.num.Num;
+
+/**
+ * Bearish kicker candle indicator.
+ *
+ * @see <a href="https://www.investopedia.com/terms/k/kickerpattern.asp">
+ *      https://www.investopedia.com/terms/k/kickerpattern.asp</a>
+ * @since 0.22.2
+ */
+public class BearishKickerIndicator extends CachedIndicator<Boolean> {
+
+    private final UpTrendIndicator trendIndicator;
+    private final RealBodyIndicator realBodyIndicator;
+    private final Num bigBodyThresholdPercentage;
+
+    /**
+     * Constructor.
+     *
+     * @param series the bar series
+     */
+    public BearishKickerIndicator(final BarSeries series) {
+        this(series, series.numFactory().numOf(0.03));
+    }
+
+    /**
+     * Constructor.
+     *
+     * @param series                     the bar series
+     * @param bigBodyThresholdPercentage percentage to determine whether a candle
+     *                                   has a big body or not
+     */
+    public BearishKickerIndicator(final BarSeries series, final Num bigBodyThresholdPercentage) {
+        super(series);
+        this.trendIndicator = new UpTrendIndicator(series);
+        this.realBodyIndicator = new RealBodyIndicator(series);
+        this.bigBodyThresholdPercentage = bigBodyThresholdPercentage;
+    }
+
+    @Override
+    protected Boolean calculate(int index) {
+        if (index < getCountOfUnstableBars()) {
+            return false;
+        }
+        Bar firstBar = getBarSeries().getBar(index - 1);
+        Bar secondBar = getBarSeries().getBar(index);
+        Num firstBarPercentage = this.realBodyIndicator.getValue(index - 1).abs().dividedBy(firstBar.getOpenPrice());
+        Num secondBarPercentage = this.realBodyIndicator.getValue(index).abs().dividedBy(secondBar.getOpenPrice());
+
+        return firstBar.isBullish() && firstBarPercentage.isGreaterThanOrEqual(bigBodyThresholdPercentage)
+                && secondBar.isBearish() && secondBarPercentage.isGreaterThanOrEqual(bigBodyThresholdPercentage)
+                && secondBar.getOpenPrice().isLessThan(firstBar.getClosePrice())
+                && secondBar.getClosePrice().isLessThan(firstBar.getOpenPrice()) && this.trendIndicator.getValue(index);
+    }
+
+    @Override
+    public int getCountOfUnstableBars() {
+        return 1;
+    }
+}
--- a/ta4j-core/src/main/java/org/ta4j/core/indicators/candles/BullishKickerIndicator.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/indicators/candles/BullishKickerIndicator.java
@@ -1,0 +1,89 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017-2025 Ta4j Organization & respective
+ * authors (see AUTHORS)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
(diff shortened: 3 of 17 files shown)


ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    private Num calculateSharpe(BarSeries series, CashFlow cashFlow, int anchorIndex, int start, int end) {
+        var numFactory = series.numFactory();
+        var zero = numFactory.zero();
+        if (end - start + 1 < 2) {
+            return zero;
+        }
+
+        var pairs = indexPairs(series, anchorIndex, start, end);
```
TheCookieLab: this is borderline, but as general convention let's prefer explicit typing for function returns

ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
```diff
@@ -0,0 +1,363 @@
...
+ * {@code periodsPerYear} is estimated from observed time deltas (count of
+ * positive deltas divided by the sum of deltas in years).
+ *
+ * @since 0.22.1
+ *
+ */
+public class SharpeRatioCriterion extends AbstractAnalysisCriterion {
+
+    private static final double SECONDS_PER_YEAR = 365.2425d * 24 * 3600;
+    private static final WeekFields ISO_WEEK_FIELDS = WeekFields.of(Locale.ROOT);
```
TheCookieLab:
```suggestion
    private static final WeekFields ISO_WEEK_FIELDS = WeekFields.ISO;
```

ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
```diff
@@ -0,0 +1,363 @@
...
+    public enum Annualization {
+        PERIOD, ANNUALIZED
+    }
+
+    private final Num annualRiskFreeRate;
+    private final Sampling sampling;
+    private final Annualization annualization;
+    private final ZoneId groupingZoneId;
+
+    public SharpeRatioCriterion(Num annualRiskFreeRate, Sampling sampling, Annualization annualization,
```
TheCookieLab:
```suggestion
    public SharpeRatioCriterion(Num annualRiskFreeRate, Sampling sampling, Annualization annualization,
            ZoneId groupingZoneId) {
        this.annualRiskFreeRate = Objects.requireNonNull(annualRiskFreeRate, "annualRiskFreeRate must not be null");
        this.sampling = Objects.requireNonNull(sampling, "sampling must not be null");
        this.annualization = Objects.requireNonNull(annualization, "annualization must not be null");
        this.groupingZoneId = Objects.requireNonNull(groupingZoneId, "groupingZoneId must not be null");
    }

    public SharpeRatioCriterion(Num annualRiskFreeRate) {
        this(annualRiskFreeRate, Sampling.PER_BAR, Annualization.ANNUALIZED, ZoneOffset.UTC);
    }
```

TheCookieLab: Could also make it a no-arg constructor defaulting annualRiskFreeRate to 0.

ta4j-core/src/test/java/org/ta4j/core/criteria/SharpeRatioCriterionTest.java
```diff
@@ -0,0 +1,531 @@
...
+                return true;
+            }
+            var now = series.getBar(i).getEndTime().atZone(ZoneOffset.UTC);
+            var next = series.getBar(i + 1).getEndTime().atZone(ZoneOffset.UTC);
+            return !YearMonth.from(now).equals(YearMonth.from(next));
+        }).toArray();
+    }
+
+    private static BarSeries compressSeries(BarSeries source, int[] indices, String name) {
+        var series = new BaseBarSeriesBuilder().withName(name).build();
```
TheCookieLab:
```suggestion
        var series = new BaseBarSeriesBuilder().withName(name).withNumFactory(source.numFactory()).build();
```

ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
```diff
@@ -0,0 +1,363 @@
...
+
+    @Override
+    public Num calculate(BarSeries series, TradingRecord tradingRecord) {
+        var numFactory = series.numFactory();
+        var zero = numFactory.zero();
+        if (tradingRecord == null) {
+            return zero;
+        }
+
+        var hasClosedPositions = tradingRecord.getPositions().stream().anyMatch(Position::isClosed);
```
TheCookieLab: Since we only check that the trading record contains at least one closed position, there could be an open position as well. The `public Num calculate(BarSeries series, Position position)` explicitly returns zero for an open position but here we include an open position's mark-to-market value as part of the equity curve. 

To maintain consistency and allow flexibility, let's add an `OpenPositionHandling` enum + constructor param with MARK_TO_MARKET and REJECT (or IGNORE) as possible values. MARK_TO_MARKET should probably be the default. The position-based calculate would then need to also handle the MARK_TO_MARKET case and the tradingRecord version to handle REJECT/IGNORE. Thoughts?

ta4j-core/src/test/java/org/ta4j/core/criteria/SharpeRatioCriterionTest.java
```diff
@@ -0,0 +1,531 @@
...
+                return true;
+            }
+            var now = series.getBar(i).getEndTime().atZone(ZoneOffset.UTC);
+            var next = series.getBar(i + 1).getEndTime().atZone(ZoneOffset.UTC);
+            return !sameIsoWeek(now, next);
+        }).toArray();
+    }
+
+    private static boolean sameIsoWeek(ZonedDateTime a, ZonedDateTime b) {
+        var weekFields = WeekFields.of(Locale.ROOT);
```
TheCookieLab:
```suggestion
        var weekFields = WeekFields.ISO;
```

ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
```diff
@@ -0,0 +1,363 @@
...
+        var numFactory = series.numFactory();
+        var one = numFactory.one();
+        var eReturn = cashFlow.getValue(currentIndex).dividedBy(cashFlow.getValue(previousIndex)).minus(one);
+        return eReturn.minus(periodRiskFree(series, previousIndex, currentIndex));
+    }
+
+    private Num periodRiskFree(BarSeries series, int previousIndex, int currentIndex) {
+        var numFactory = series.numFactory();
+        var deltaYears = deltaYears(series, previousIndex, currentIndex);
+        if (deltaYears <= 0.0) {
```
TheCookieLab:
```suggestion
                if (annualRiskFreeRate.isZero() || deltaYears <= 0.0) {
            return numFactory.zero();
        }

        final Num one = numFactory.one();
        final Num annualRiskFreeRateConverted = numFactory.produces(annualRiskFreeRate) ? annualRiskFreeRate
                : numFactory.numOf(annualRiskFreeRate.doubleValue());

        final Num base = one.plus(annualRiskFreeRateConverted);
        final Num t = numFactory.numOf(deltaYears);

        return base.pow(t).minus(one);
```

TheCookieLab: may need to clean the suggested up a bit, consider it pseudo-code.

ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
francescopeloi: yeah that makes sense to me, I'll do it.

francescopeloi: @TheCookieLab after reviewing the code more:

I think the code currently is somewhat consistent: it returns 0 if there's only one position (or none), and that only position is open. If there's more than one, and there's an open position, it then uses mark to market.

This logic was applied on the `Position` `calculate()` because, if the position is open, CashFlow throws an exception. Having said this: the `Position` `calculate()` is probably useless in Sharpe Ratio, as calculating a Sharpe Ratio with one position only is non-sense. I wonder if we should just throw an exception in this case and not over-complicate things. Or just always return 0.

On the `TradingRecord` `calculate()`: in here I could add some code to either consider open positions (with mark to market) or not, or leave the current code as it is, that uses already mark to market if there are open positions.

Let me know what you think.

ta4j-core/src/main/java/org/ta4j/core/analysis/ExcessReturns.java
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * Creates an excess return calculator with invested interval detection from a
+     * trading record.
+     *
+     * @param series             the bar series providing time deltas and num
+     *                           factory
+     * @param annualRiskFreeRate the annual risk-free rate (e.g. 0.05 for 5%)
+     * @param cashReturnPolicy   the policy for flat equity intervals
+     * @param tradingRecord      the trading record used to detect invested
```
TheCookieLab, line 84: the javadoc calls out tradingRecord as nullable which is in conflict with the Objects.requireNonNull in the base constructor.

ta4j-core/src/main/java/org/ta4j/core/analysis/frequency/SamplingFrequencyIndexes.java
```diff
@@ -0,0 +1,140 @@
...
+     * Creates a grouping helper for the chosen aggregation mode and time zone.
+     *
+     * @param samplingFrequency the sampling granularity
+     * @param groupingZoneId    the time zone used to interpret bar end times
+     * @since 0.22.2
+     */
+    public SamplingFrequencyIndexes(SamplingFrequency samplingFrequency, ZoneId groupingZoneId) {
+        this.samplingFrequency = Objects.requireNonNull(samplingFrequency, "samplingFrequency must not be null");
+        this.groupingZoneId = Objects.requireNonNull(groupingZoneId, "groupingZoneId must not be null");
+        ;
```
TheCookieLab: extra semi-colon here

ta4j-core/src/main/java/org/ta4j/core/analysis/frequency/SamplingFrequencyIndexes.java
```diff
@@ -0,0 +1,140 @@
...
+     *
+     * @param series      the bar series
+     * @param anchorIndex the starting anchor index for the first sampled pair
+     * @param start       the first index eligible for sampling
+     * @param end         the last index eligible for sampling
+     * @return a stream of index pairs describing each sampling interval
+     * @since 0.22.2
+     */
+    public Stream<IndexPair> sample(BarSeries series, int anchorIndex, int start, int end) {
+        if (samplingFrequency == SamplingFrequency.BAR) {
```
TheCookieLab, line 83: Add some index validations here, like `if (start > end || end - start + 1 < 2) {
    return zero;
}` or similar.

ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
```diff
@@ -0,0 +1,272 @@
...
+        var sharpePerPeriod = summary.mean().dividedBy(stdev);
+
+        if (annualization == Annualization.PERIOD) {
+            return sharpePerPeriod;
+        }
+
+        return summary.annualizationFactor(numFactory).map(sharpePerPeriod::multipliedBy).orElse(sharpePerPeriod);
+    }
+
+    private static boolean hasInsufficientPositions(TradingRecord tradingRecord,
```
TheCookieLab: Do we need this check? Consider a strategy with a single position or even a single open position (i.e a buy and hold). These would be registered as insufficient positions no? I think 231-233 already protects against the insufficient data scenario so this check may be both redundant and incorrectly flagging valid scenarios.

ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
```diff
@@ -0,0 +1,272 @@
...
+
+    @Override
+    public Num calculate(BarSeries series, TradingRecord tradingRecord) {
+        var numFactory = series.numFactory();
+        var zero = numFactory.zero();
+        if (hasInsufficientPositions(tradingRecord, openPositionHandling)) {
+            return zero;
+        }
+
+        var start = series.getBeginIndex() + 1;
```
TheCookieLab: Should this use the TradingRecord's start & end rather than that of the bar series?

francescopeloi: I think we should be using bar series indexes, as the strategy has been tested on the whole bar series and not only when we started a position. Using the trading record would cut off some bars at the beginning and some at the end, inflating the Sharpe ratio, and it'll also change the sample size (less bars), so making other strategies run on the same bar series incomparable. We are actually handling out-of-market of markets periods via the CashReturnPolicy.

francescopeloi: after reading also your other comment, I think I need to think about this more

ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
francescopeloi: Fair comment, I removed the check. I was getting confused between number of trades and number of samples. We could have a good number of samples even if one open trade only, so a trade count check doesn't make much sense. For this reason I also re-instantiated the position-calculate.

ta4j-core/src/main/java/org/ta4j/core/criteria/SharpeRatioCriterion.java
francescopeloi: ok, I've answered the other comment. I think my answer above still stands.

Issue #1410, opened 2026-01-14
[BUG] `series.getSubSeries(startIndex, endIndex)` returns a series with maxBars=Integer.MAX_VALUE


drse:
Seeing this on `0.22.0`: When taking a sub series of an existing series, the returned series has `maxBars` set to the default `Integer.MAX_VALUE` rather than the maxBars of the original series. I believe it should also copy the maxBars in addition to the name and numFactory in `BaseBarSeries.java:147`

PR #1413, opened 2026-01-15
edge case missed on #1410

francescopeloi:
edge case missed on #1410


ta4j-core/src/main/java/org/ta4j/core/BaseBarSeries.java
```diff
@@ -92,8 +94,6 @@ public class BaseBarSeries implements BarSeries {
         this.barBuilderFactory = Objects.requireNonNull(barBuilderFactory);
         if (bars.isEmpty()) {
             // Bar list empty
-            this.seriesBeginIndex = -1;
```
francescopeloi, line 95: just some cleaning, these are already assigned above

PR #1420, opened 2026-01-18
Candle indicators

btezergil:
- Added three inside up/down candlestick indicators
- Added morning/evening star candlestick indicators
- Added bullish/bearish kicker candlestick indicators
- Added piercing/dark cloud candlestick indicators

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - **Release token preflight**: Fail fast when `GH_TA4J_REPO_TOKEN` lacks write permission (warn-only in dry-run mode).
 - **Release scheduler enablement**: Gate scheduled runs on `RELEASE_SCHEDULER_ENABLED` (defaults to disabled when unset).
 - **Factory selection from bars**: Derive the NumFactory from the first available bar price instead of assuming a specific price is always present.
+- **CashFlow**: Added a realized-only calculation mode alongside the default mark-to-market cash flow curve.
 
 ### Fixed
 - **TimeBarBuilder**: Preserve in-progress bars when trade ingestion skips across multiple time periods.
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,7 @@
         <maven.surefire.version>3.5.4</maven.surefire.version>
 
         <!-- Dependencies -->
-        <assertj.version>3.27.6</assertj.version>
+        <assertj.version>3.27.7</assertj.version>
         <commons-math3.version>3.6.1</commons-math3.version>
         <gson.version>2.13.2</gson.version>
         <jfreechart.version>1.5.6</jfreechart.version>
--- a/ta4j-core/src/main/java/org/ta4j/core/analysis/CashFlow.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/analysis/CashFlow.java
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
+import java.util.Objects;
 import org.ta4j.core.BarSeries;
 import org.ta4j.core.Indicator;
 import org.ta4j.core.Position;
@@ -35,7 +36,7 @@ import org.ta4j.core.num.Num;
 
 /**
  * Allows to follow the money cash flow involved by a list of positions over a
- * bar series.
+ * bar series, either marked to market or using realized values only.
  */
 public class CashFlow implements Indicator<Num> {
 
@@ -45,6 +46,9 @@ public class CashFlow implements Indicator<Num> {
     /** The (accrued) cash flow sequence (without trading costs). */
     private final List<Num> values;
 
+    /** The equity curve calculation mode. */
+    private final EquityCurveMode equityCurveMode;
+
     /**
      * Constructor for cash flows of a closed position.
      *
@@ -52,7 +56,21 @@ public class CashFlow implements Indicator<Num> {
      * @param position  a single position
      */
     public CashFlow(BarSeries barSeries, Position position) {
-        this.barSeries = barSeries;
+        this(barSeries, position, EquityCurveMode.MARK_TO_MARKET);
+    }
+
+    /**
+     * Constructor for cash flows of a closed position.
+     *
+     * @param barSeries       the bar series
+     * @param position        a single position
+     * @param equityCurveMode the calculation mode
+     *
+     * @since 0.22.2
+     */
+    public CashFlow(BarSeries barSeries, Position position, EquityCurveMode equityCurveMode) {
+        this.barSeries = Objects.requireNonNull(barSeries);
+        this.equityCurveMode = Objects.requireNonNull(equityCurveMode);
         values = new ArrayList<>(Collections.singletonList(barSeries.numFactory().one()));
 
         calculate(position);
@@ -66,7 +84,20 @@ public class CashFlow implements Indicator<Num> {
      * @param tradingRecord the trading record
      */
     public CashFlow(BarSeries barSeries, TradingRecord tradingRecord) {
-        this(barSeries, tradingRecord, tradingRecord.getEndIndex(barSeries));
+        this(barSeries, tradingRecord, tradingRecord.getEndIndex(barSeries), EquityCurveMode.MARK_TO_MARKET);
+    }
+
+    /**
+     * Constructor.
+     *
+     * @param barSeries       the bar series
+     * @param tradingRecord   the trading record
+     * @param equityCurveMode the calculation mode
+     *
+     * @since 0.22.2
+     */
+    public CashFlow(BarSeries barSeries, TradingRecord tradingRecord, EquityCurveMode equityCurveMode) {
+        this(barSeries, tradingRecord, tradingRecord.getEndIndex(barSeries), equityCurveMode);
     }
 
     /**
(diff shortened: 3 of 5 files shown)


ta4j-core/src/main/java/org/ta4j/core/indicators/candles/BearishKickerIndicator.java
```diff
@@ -0,0 +1,77 @@
...
+    public BearishKickerIndicator(final BarSeries series) {
+        super(series);
+        this.trendIndicator = new UpTrendIndicator(series);
+        this.realBodyIndicator = new RealBodyIndicator(series);
+        this.bigBodyThresholdPercentage = series.numFactory().numOf(0.03);
+    }
+
+    @Override
+    protected Boolean calculate(int index) {
+        if (index < 1) {
```
TheCookieLab: Let's move this threshold to the getCountOfUnstableBars(). See RSIIndicator for an example. This comment applies universally to all new classes in this PR.

ta4j-examples/.factorypath
```diff
@@ -0,0 +1,17 @@
+<factorypath>
```
TheCookieLab: .gitignore this

ta4j-core/src/main/java/org/ta4j/core/indicators/candles/BearishKickerIndicator.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Constructor.
+     *
+     * @param series the bar series
+     */
+    public BearishKickerIndicator(final BarSeries series) {
+        super(series);
+        this.trendIndicator = new UpTrendIndicator(series);
+        this.realBodyIndicator = new RealBodyIndicator(series);
+        this.bigBodyThresholdPercentage = series.numFactory().numOf(0.03);
```
TheCookieLab: Rather than hardcode this to 0.03, let's make it configurable via optional constructor param. If not provided we can use 0.03 as the default. This comment also applies to the other classes.

ta4j-core/src/main/java/org/ta4j/core/indicators/candles/MorningStarIndicator.java
```diff
@@ -0,0 +1,83 @@
...
+        Bar secondBar = getBarSeries().getBar(index - 1);
+        Bar thirdBar = getBarSeries().getBar(index);
+        Num firstBarPercentage = this.realBodyIndicator.getValue(index - 2).abs().dividedBy(firstBar.getOpenPrice());
+        Num secondBarPercentage = this.realBodyIndicator.getValue(index - 1).abs().dividedBy(secondBar.getOpenPrice());
+        Num thirdBarPercentage = this.realBodyIndicator.getValue(index).abs().dividedBy(thirdBar.getOpenPrice());
+
+        return firstBar.isBearish() && firstBarPercentage.isGreaterThanOrEqual(bigBodyThresholdPercentage)
+                && secondBar.getOpenPrice().isLessThan(firstBar.getClosePrice())
+                && secondBarPercentage.isLessThanOrEqual(smallBodyThresholdPercentage)
+                && thirdBar.getClosePrice().isLessThan(firstBar.getOpenPrice()) && thirdBar.isBullish()
```
TheCookieLab: I'm certainly no expert on candle indicators but for a bullish reversal should the 3rd bar's close price be less than the 1st bar's open? It seems it should be the other way around, i.e. the 3rd bar's close should be higher than the 1st bar's open.

ta4j-core/src/main/java/org/ta4j/core/indicators/candles/EveningStarIndicator.java
```diff
@@ -0,0 +1,83 @@
...
+        Bar secondBar = getBarSeries().getBar(index - 1);
+        Bar thirdBar = getBarSeries().getBar(index);
+        Num firstBarPercentage = this.realBodyIndicator.getValue(index - 2).abs().dividedBy(firstBar.getOpenPrice());
+        Num secondBarPercentage = this.realBodyIndicator.getValue(index - 1).abs().dividedBy(secondBar.getOpenPrice());
+        Num thirdBarPercentage = this.realBodyIndicator.getValue(index).abs().dividedBy(thirdBar.getOpenPrice());
+
+        return firstBar.isBullish() && firstBarPercentage.isGreaterThanOrEqual(bigBodyThresholdPercentage)
+                && secondBar.getOpenPrice().isGreaterThan(firstBar.getClosePrice())
+                && secondBarPercentage.isLessThanOrEqual(smallBodyThresholdPercentage)
+                && thirdBar.getClosePrice().isGreaterThan(firstBar.getOpenPrice()) && thirdBar.isBearish()
```
TheCookieLab: Similar to the comment below for the MorningStarIndicator, during a bearish reversal shouldn't the 3rd bar close comparable or lower than the 1st bar? What you use specifically as the reference point - open vs close (or even midpoint) can be up for debate.

ta4j-core/src/test/java/org/ta4j/core/indicators/candles/EveningStarIndicatorTest.java
```diff
@@ -0,0 +1,80 @@
...
+        List<Bar> bars = new ArrayList<Bar>(30);
+        for (int i = 0; i < 17; ++i) {
+            bars.add(
+                    new MockBarBuilder(numFactory).openPrice(i).closePrice(i + 6).highPrice(i + 8).lowPrice(i).build());
+        }
+
+        return bars;
+    }
+
+    @Test
```
TheCookieLab, line 66: please add more tests for negative and edge-case scenarios. Unit tests not only demonstrate the code is doing what it's intended to, but just as importantly serves as a shield against regressions during future changes.

ta4j-core/src/main/java/org/ta4j/core/indicators/candles/ThreeInsideDownIndicator.java
```diff
@@ -0,0 +1,69 @@
...
+        this.trendIndicator = new UpTrendIndicator(series);
+    }
+
+    @Override
+    protected Boolean calculate(int index) {
+        if (index < 2) {
+            // Three inside down is a 3-candle pattern
+            return false;
+        }
+        BearishHaramiIndicator harami = new BearishHaramiIndicator(getBarSeries());
```
TheCookieLab: Let's make this a class property (final) rather than instantiate every time in the hot path.

ta4j-core/src/main/java/org/ta4j/core/indicators/candles/ThreeInsideUpIndicator.java
```diff
@@ -0,0 +1,69 @@
...
+        this.trendIndicator = new DownTrendIndicator(series);
+    }
+
+    @Override
+    protected Boolean calculate(int index) {
+        if (index < 2) {
+            // Three inside up is a 3-candle pattern
+            return false;
+        }
+        BullishHaramiIndicator harami = new BullishHaramiIndicator(getBarSeries());
```
TheCookieLab: same as above

ta4j-core/src/main/java/org/ta4j/core/indicators/candles/PiercingIndicator.java
```diff
@@ -0,0 +1,81 @@
...
+ * Piercing candle indicator.
+ *
+ * @see <a href="https://www.investopedia.com/terms/piercing-pattern.asp">
+ *      https://www.investopedia.com/terms/piercing-pattern.asp</a>
+ */
+public class PiercingIndicator extends CachedIndicator<Boolean> {
+
+    private final DownTrendIndicator trendIndicator;
+    private final RealBodyIndicator realBodyIndicator;
+    private Num bigBodyThresholdPercentage;
```
TheCookieLab: Can be final.

ta4j-core/src/main/java/org/ta4j/core/indicators/candles/BearishKickerIndicator.java
```diff
@@ -0,0 +1,77 @@
...
+import org.ta4j.core.indicators.CachedIndicator;
+import org.ta4j.core.indicators.trend.UpTrendIndicator;
+import org.ta4j.core.num.Num;
+
+/**
+ * Bearish kicker candle indicator.
+ *
+ * @see <a href="https://www.investopedia.com/terms/k/kickerpattern.asp">
+ *      https://www.investopedia.com/terms/k/kickerpattern.asp</a>
+ */
```
TheCookieLab, line 38: Please tag these new classes with @since x.xx

ta4j-core/src/main/java/org/ta4j/core/indicators/candles/MorningStarIndicator.java
btezergil: I've made some research before implementing this, and checked some public sites again. While multiple say that it should be higher than the middle point, others also add that it should be less than the first candle's open. I think the close point is not the main focus here. All content basically agree on the candle needing to be bullish + having a big body. 

I think the best course of action is to remove that requirement altogether, or change it so that it's higher than the middle point. The closing point of the third candle seems to be up for debate. We can discuss this further, I'm open to change this.

ta4j-core/src/test/java/org/ta4j/core/indicators/candles/EveningStarIndicatorTest.java
btezergil: I've pushed all other review items except this one. I will add some extra tests - do you need the tests only for morning/evening star or for all classes?

TheCookieLab:
All classes.

PR #1427, opened 2026-01-27
Equity curve mode control and open position handling

francescopeloi:
- **Equity curve mode control and open position handling**: Added constructors to select mark-to-market or realized calculation and last open position handling for CashFlow, CumulativePnL, Returns, and as a consequence to drawdown and Sharpe ratio criteria.
- ReturnOverMaxDrawdownCriterion left out on purpose for now as it's a bit more complicated


TheCookieLab:
Do not proceed with these changes as it has significant overlap with the trading record changes I'm currently working on. I'll use this PR as reference and pull in the core pieces.

PR #1443, opened 2026-02-03
Add custom ATR constructors to stop loss & stop gain rules

### LINKED ISSUE #1441: Allow passing ATRIndicator into AverageTrueRangeStopGainRule
- [x ] I could not find similar [issues](https://github.com/ta4j/ta4j/issues?utf8=%E2%9C%93&q=)
- [x ] I could not find a solution in the [wiki](https://ta4j.github.io/ta4j-wiki/) or [faq section](https://ta4j.github.io/ta4j-wiki/FAQ.html)

When I run backtest using AverageTrueRangeStopGainRule it is recreated multiple times. 
The problem is that performance is very poor when I use it, as it manages (initialises) ATRIndicator internally, which in turn creates MMAIndicator which implements RecursiveCachedIndicator.
 
For other indicators implementing RecursiveCachedIndicator I use indicator pool that allows to cache them across backtesting rounds.

I would like to have an ability to use ATRIndicator from my pool with AverageTrueRangeStopGainRule.

Simple solution that i see:
Add new AverageTrueRangeStopGainRule (and similar rules) constructor that should allow to pass an ATRIndicator instance explicitly (In my case from cached indicator pool).

or alternatively add virtual method into AverageTrueRangeStopGainRule that creates ATRIndicator instance. So caller could extend AverageTrueRangeStopGainRule and override the method.

nsk90:
Fixes #1441.

- Using custom ATR indicator is useful for performance gain as it uses RecursiveCachedIndicator ancestor inside, and might be cached and reused.

### DIFF
--- a/ta4j-core/src/test/java/org/ta4j/core/rules/AverageTrueRangeStopGainRuleTest.java
+++ b/ta4j-core/src/test/java/org/ta4j/core/rules/AverageTrueRangeStopGainRuleTest.java
@@ -3,20 +3,24 @@
  */
 package org.ta4j.core.rules;
 
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
-
-import java.time.Duration;
-import java.time.Instant;
-
 import org.junit.Before;
 import org.junit.Test;
 import org.ta4j.core.BarSeries;
 import org.ta4j.core.BaseTradingRecord;
 import org.ta4j.core.Trade;
+import org.ta4j.core.indicators.ATRIndicator;
+import org.ta4j.core.indicators.helpers.ClosePriceIndicator;
 import org.ta4j.core.indicators.helpers.FixedIndicator;
 import org.ta4j.core.mocks.MockBarSeriesBuilder;
 import org.ta4j.core.num.Num;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.List;
+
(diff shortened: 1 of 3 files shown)


ta4j-core/src/main/java/org/ta4j/core/rules/AverageTrueRangeStopLossRule.java
```diff
@@ -57,10 +59,22 @@ public AverageTrueRangeStopLossRule(BarSeries series, int atrBarCount, Number at
...
+    /**
+     * Constructor with custom reference price and ATR indicators.
+     *
+     * @param referencePrice the reference price indicator
+     * @param atrIndicator   ATR indicator
+     * @param atrCoefficient the coefficient to multiply ATR
+     */
+    public AverageTrueRangeStopLossRule(final Indicator<Num> referencePrice,
+            final ATRIndicator atrIndicator, final Number atrCoefficient) {
+        this.referencePrice = requireNonNull(referencePrice);
```
TheCookieLab, line 74: please add unit tests covering all the code you changed and make sure they pass both before and after the changes. Doubly so if there is a behavior change (like on these lines).

nsk90: [done] Add unit-test
[done] Check passing after change
[done] Also check that new unit-tests work before changes in production code.

TheCookieLab (approved):

PR #1445, opened 2026-02-05
Update Indicator Unstable Counts

### LINKED ISSUE #947: Return `NaN` for Indicators while within unstableBars?

TheCookieLab:
Fixes #947

- Standardized unstable-bar aggregation and NaN gating across indicators to ensure consistent warm-up behavior.

- [X] added an entry with related ticket number(s) to the unreleased section of `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,7 @@
 - **TimeBarBuilder**: Preserve in-progress bars when trade ingestion skips across multiple time periods.
 - **Release workflow notifications**: Fix discussion comment posting in workflows (unescaped template literals).
 - **Release health workflow**: Ensure discussion notifications are posted even when summary generation fails and avoid `github-script` core redeclaration errors.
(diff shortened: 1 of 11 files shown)


TheCookieLab:
@coderabbitai full review

PR #1497, opened 2026-04-07
Add IndicatorFamilyManager analysis API

TheCookieLab:
Fixes #.

- Adds `IndicatorFamilyManager` and `IndicatorFamilyResult` in `org.ta4j.core.indicators` so callers can group behaviorally similar indicators by passing a `BarSeries`, a deterministic `Map<String, Indicator<Num>>`, and an optional similarity threshold.
- Keeps the new public API intentionally small by removing the unreleased manifest/config/engine/catalog surface and using absolute population correlation internally for pair scoring.
- Adds a realistic S&P 500 weekly example that analyzes 27 mixed indicators at thresholds `0.80`, `0.90`, and `0.97`, plus black-box tests for the manager and demo flow.

- [x] I have run `mvn -B clean license:format formatter:format test install` to format and test my changes per the [contributing guidelines](.github/CONTRIBUTING.md)

## Summary by CodeRabbit

* **New Features**
  * Added indicator family analysis capability to group similar indicators and compute pairwise similarity scores.
  * Similarity threshold tuning allows you to discover tighter or broader indicator substitutes.
  * New example demonstrating indicator family analysis on S&P 500 data.

* **Improvements**
  * Enhanced correlation coefficient calculations with improved null-safety and stability handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ta4j/ta4j/pull/1497)


TheCookieLab:
@coderabbitai full review

TheCookieLab:
@coderabbitai full review

TheCookieLab (commented):
Inline findings only.

— ChatGPT

ta4j-core/src/main/java/org/ta4j/core/indicators/IndicatorFamilyManager.java
```diff
@@ -0,0 +1,537 @@
...
+ * windows better match the analysis horizon.
+ * <p>
+ * Custom metric factories may be supplied when a different signed similarity
+ * stream is more appropriate. The returned metric indicator must use the same
+ * {@link BarSeries} and must return values in {@code [-1, 1]}; invalid values
+ * are skipped.
+ *
+ * @since 0.22.7
+ */
+public final class IndicatorFamilyManager {
```
TheCookieLab, line 43: [Major] Move this whole-series analysis service out of the indicator package

Location: `IndicatorFamilyManager`
Violated principle/invariant: One concept should have one obvious home; public indicator-package types should follow the indicator composition model.
What happens: This type does not implement `Indicator` or expose `getValue(index)`. It accepts a catalog, scans full history, manages an executor, clusters pairs, and returns a report. That is an analysis service, while `org.ta4j.core.indicators` is documented for mathematical indicator transformations and the repository already has `org.ta4j.core.analysis` for analysis workflows.
Operator/downstream consequence: The first released package becomes permanent API debt: discovery and imports imply this composes like an indicator when it cannot, and a later move to the analysis surface becomes a migration.
Evidence: The primary journey is `new IndicatorFamilyManager(series).analyze(map)`, not an indicator value-at-index journey; the companion `IndicatorFamilyResult` is likewise an analysis report.
Recommended remediation: Keep the reduced API, but move the public entry point and result to an analysis-owned package such as `org.ta4j.core.analysis.indicators` (and name the entry point for the analysis it performs). Keep pair requests, clustering, and executor mechanics package-private.
Regression proof: Add an external-style compile fixture using the final analysis package and update the example, package documentation, and changelog together.

ta4j-core/src/main/java/org/ta4j/core/indicators/IndicatorFamilyManager.java
```diff
@@ -0,0 +1,537 @@
...
+                SampleType.POPULATION);
+    }
+
+    private LinkedHashMap<String, Indicator<Num>> validateIndicators(Map<String, Indicator<Num>> indicators) {
+        if (indicators == null || indicators.isEmpty()) {
+            throw new IllegalArgumentException("indicators must not be empty");
+        }
+
+        LinkedHashMap<String, Indicator<Num>> orderedIndicators = new LinkedHashMap<>(indicators.size());
+        for (Map.Entry<String, Indicator<Num>> entry : indicators.entrySet()) {
```
TheCookieLab, line 220: [Major] Canonicalize input order instead of relying on an unverifiable `Map` precondition

Location: `validateIndicators`
Violated principle/invariant: Deterministic family IDs, pair ordering, and tie-breaking must be reproducible from the logical input.
What happens: This copies whatever order `Map.entrySet()` happens to expose. That order then controls pair order, union roots, `family-###` IDs, family order, and representative tie-breaking. The signature accepts any `Map`, so the documented “deterministic iteration order” requirement cannot be checked and common inputs such as `HashMap`/`Map.of(...)` do not contractually provide it.
Operator/downstream consequence: The same named indicator set and scores can produce different serialized/report output across construction paths or JVM runs, defeating the PRD's deterministic-rerun objective.
Evidence: Every downstream list and ID is derived from this copied order without a canonical sort.
Recommended remediation: Canonicalize by stable indicator key before pair construction. If caller order is intentionally a consequential input, represent and persist it with an explicitly ordered domain input rather than a generic `Map` precondition.
Regression proof: Analyze the same named indicators supplied in opposite insertion orders and assert identical family IDs, mappings, representatives, and pair ordering.

ta4j-core/src/main/java/org/ta4j/core/indicators/IndicatorFamilyManager.java
```diff
@@ -0,0 +1,537 @@
...
+                similarityMetric.getCountOfUnstableBars());
+        int endIndex = similarityMetric.getBarSeries().getEndIndex();
+        NumFactory numFactory = similarityMetric.getBarSeries().numFactory();
+        Num absoluteTotal = numFactory.zero();
+        Num signedTotal = numFactory.zero();
+        Num latestSignedSimilarity = numFactory.zero();
+        Num minimumSignedSimilarity = null;
+        Num maximumSignedSimilarity = null;
+        int samples = 0;
+        for (int index = startIndex; index <= endIndex; index++) {
```
TheCookieLab, line 359: [Critical] Do not increment an inclusive `int` loop past `Integer.MAX_VALUE`

Location: pair-similarity scan
Violated principle/invariant: A legal terminal index must terminate; index arithmetic must not wrap.
What happens: When `endIndex == Integer.MAX_VALUE`, the iteration that processes that index executes `index++`, wraps to `Integer.MIN_VALUE`, and still satisfies `index <= endIndex`. The analysis then continues indefinitely through invalid indexes. The newly added inclusive loops in `VarianceIndicator` and `CovarianceIndicator` have the same terminal-overflow defect.
Operator/downstream consequence: A bounded rolling series can eventually hang an analysis despite retaining little data.
Evidence: Java `int` increment wraps; this loop has no terminal break before the increment.
Recommended remediation: Use a terminal-safe loop form (for example, process then `if (index == endIndex) break`) or iterate in `long` with checked conversion. Apply the same fix to every touched inclusive index loop.
Regression proof: Exercise a proxy series/metric whose one-element visible range begins and ends at `Integer.MAX_VALUE`; the manager and both statistics indicators must return exactly once.

ta4j-core/src/main/java/org/ta4j/core/indicators/IndicatorFamilyManager.java
```diff
@@ -0,0 +1,537 @@
...
+            absoluteTotal = absoluteTotal.plus(signedSimilarity.abs());
+            signedTotal = signedTotal.plus(signedSimilarity);
+            latestSignedSimilarity = signedSimilarity;
+            minimumSignedSimilarity = minimumSignedSimilarity == null ? signedSimilarity
+                    : minimumSignedSimilarity.min(signedSimilarity);
+            maximumSignedSimilarity = maximumSignedSimilarity == null ? signedSimilarity
+                    : maximumSignedSimilarity.max(signedSimilarity);
+            samples++;
+        }
+        if (samples == 0) {
```
TheCookieLab, line 376: [Major] Preserve “unavailable” instead of fabricating a zero similarity

Location: zero-sample branch in `estimatePairSimilarity`
Violated principle/invariant: Undefined/unavailable statistics must not be represented as an observed zero.
What happens: A short history, two constant indicators, or any metric that yields no valid samples returns `similarity=0` and zero-valued signed diagnostics. `sampleCount` reveals the absence, but every numerical field still claims a measured result. At the valid threshold `0`, this branch causes an unavailable pair to satisfy `0 >= 0` and be merged into a family.
Operator/downstream consequence: Missing evidence can be ranked, reported, and clustered as real independence; at threshold zero it directly creates a false family.
Evidence: The existing zero-stable-sample test explicitly codifies all-zero metrics rather than an unavailable state.
Recommended remediation: Return the repository's unavailable numeric value (`NaN.NaN`) or an explicit status-bearing pair result, permit that state only when `sampleCount == 0`, and exclude unavailable pairs from clustering/ranking.
Regression proof: With no valid observations and threshold `0`, assert that the pair is unavailable and is not merged; append enough valid data and assert a later analysis recovers to a defined score.

ta4j-core/src/main/java/org/ta4j/core/indicators/IndicatorFamilyManager.java
```diff
@@ -0,0 +1,537 @@
...
+            List<PairAnalysis> pairAnalyses, Num similarityThreshold, NumFactory numFactory) {
+        UnionFind unionFind = new UnionFind(indicatorNames.size());
+        Num[][] similarityByIndex = new Num[indicatorNames.size()][indicatorNames.size()];
+        for (int index = 0; index < indicatorNames.size(); index++) {
+            similarityByIndex[index][index] = numFactory.one();
+        }
+        for (PairAnalysis pair : pairAnalyses) {
+            similarityByIndex[pair.leftIndex()][pair.rightIndex()] = pair.similarity();
+            similarityByIndex[pair.rightIndex()][pair.leftIndex()] = pair.similarity();
+            if (pair.similarity().isGreaterThanOrEqual(similarityThreshold)) {
```
TheCookieLab, line 407: [Major] Enforce the threshold at the family level instead of single-link chaining

Location: `clusterIntoFamilies`
Violated principle/invariant: A family presented as a set of substitutes must satisfy the documented cohesion meaning consistently across all members.
What happens: Union-find implements connected-component/single-link clustering. If A–B is `0.95` and B–C is `0.95`, A and C are placed in one family even when A–C is `0.40` at a `0.90` threshold. The current test deliberately expects that family, whose reported average (`0.766…`) and minimum (`0.40`) are both below the grouping threshold.
Operator/downstream consequence: A genuinely distinct indicator can be discarded as a duplicate merely because it is connected through a bridge, undermining the feature's purpose of preventing false confluence. It also conflicts with the example/release language describing higher-threshold families as tight substitutes.
Evidence: `exposesFamilyRepresentativeAndCohesionMetrics` is an exact counterexample.
Recommended remediation: Use a clustering rule that enforces the advertised family invariant (for example, complete-link so every pair clears the threshold, or a clearly defined average-link criterion). If single-link is intentional, expose that consequential policy explicitly and rename/document the threshold as an edge-connectivity threshold rather than substitute similarity.
Regression proof: Keep the `0.95/0.95/0.40` bridge fixture and require A and C to land in different threshold-coherent families under the default policy.

ta4j-core/src/main/java/org/ta4j/core/indicators/IndicatorFamilyManager.java
```diff
@@ -0,0 +1,537 @@
...
+
+    private static Num clamp(Num value, NumFactory numFactory) {
+        return value.max(numFactory.zero()).min(numFactory.one());
+    }
+
+    private static Num clampSigned(Num value, NumFactory numFactory) {
+        return value.max(numFactory.minusOne()).min(numFactory.one());
+    }
+
+    private static int pairCount(int indicatorCount) {
```
TheCookieLab, line 491: [Major] Bound pairwise work before allocating quadratic state

Location: `pairCount` / pair materialization
Violated principle/invariant: Expensive pair counts and allocations must be checked before materialization, with overflow-safe arithmetic and an operator-visible bound.
What happens: The manager eagerly allocates every `PairRequest`, every pair result/future, and later an `N x N` `Num[][]`, with no catalog-size guard. The `int` multiplication already overflows at 46,342 indicators: `pairCount` becomes zero while the nested loops still attempt to append 1,073,767,311 pairs.
Operator/downstream consequence: A large or accidentally generated registry can hang or exhaust the heap instead of failing predictably; much smaller catalogs can still create prohibitive dense matrices.
Evidence: `indicatorCount * (indicatorCount - 1)` is evaluated in `int`, and all downstream storage is quadratic.
Recommended remediation: Compute the count with checked `long` arithmetic, define a documented maximum pair/catalog budget, reject before invoking metric factories or allocating, and avoid duplicate dense representations where possible (lazy/batched pair evaluation plus bounded storage).
Regression proof: Supply a registry at the overflow boundary and assert a specific fail-fast exception before any metric-factory call; add a realistic worst-case memory/time benchmark at the supported maximum.

ta4j-core/src/main/java/org/ta4j/core/indicators/IndicatorFamilyResult.java
```diff
@@ -0,0 +1,174 @@
...
+ */
+public record IndicatorFamilyResult(Num similarityThreshold, int stableIndex, Map<String, String> familyByIndicator,
+        List<Family> families, List<PairSimilarity> pairSimilarities) {
+
+    /**
+     * Creates a validated result.
+     *
+     * @since 0.22.7
+     */
+    public IndicatorFamilyResult {
```
TheCookieLab, line 34: [Major] Make the public result impossible to construct inconsistently

Location: `IndicatorFamilyResult` canonical constructor and nested public records
Violated principle/invariant: All result fields must describe one coherent population; producer-owned diagnostics should not expose invalid public states.
What happens: Callers can construct a result whose mapping disagrees with `families`, whose family IDs/members are duplicated or missing, whose pair list is incomplete/duplicated/unknown (or even `null`, silently converted to empty), and whose statistics are mathematically contradictory. For example, `minimumInternalSimilarity > averageInternalSimilarity`, or `abs(signedAverageSimilarity) > similarity`, currently passes validation.
Operator/downstream consequence: Consumers cannot trust `familyByIndicator`, family cohesion, and pair diagnostics to agree, so selection, reporting, persistence, and equality can produce contradictory answers from one supposedly immutable value.
Evidence: Validation is field-local; no top-level coverage/uniqueness checks or cross-statistic laws are enforced.
Recommended remediation: Prefer a producer-owned public final result with package-private construction and public immutable accessors. If public construction is required, validate exact membership/mapping coverage, unique IDs/names/pairs, complete pair cardinality, numeric-factory coherence, and all aggregate/min/max/sample-count relationships.
Regression proof: Add direct-construction tests for a mismatched mapping, duplicate membership/pair, missing pair, `minimum > average`, signed average outside min/max, and `abs(signedAverage) > similarity`; each must fail.

ta4j-core/src/main/java/org/ta4j/core/indicators/statistics/VarianceIndicator.java
```diff
@@ -79,11 +77,11 @@ public static VarianceIndicator ofPopulation(Indicator<Num> indicator, int barCo
 
     @Override
     protected Num calculate(int index) {
-        final int startIndex = Math.max(0, index - barCount + 1);
+        final int startIndex = Math.max(Math.max(0, getBarSeries().getBeginIndex()), index - barCount + 1);
```
TheCookieLab, line 80: [Major] Advance the stability boundary when retained history starts later

Location: begin-index-aware rolling calculation (also applies to `CovarianceIndicator`)
Violated principle/invariant: `getCountOfUnstableBars()` must identify the first index whose full requested window is stable.
What happens: Calculation now clamps the window start to `BarSeries.getBeginIndex()`, but the unstable count remains `sourceUnstable + barCount - 1`. For a retained series beginning at `150` with a 120-bar window, the manager reports/starts at `150`; correlations from `151` through `268` can be numerically defined but use only 2–119 retained observations. The first full retained window is `269`.
Operator/downstream consequence: `IndicatorFamilyManager` averages partial-window correlations as stable evidence and reports a `stableIndex` that does not match the represented window.
Evidence: The rolling-series test currently expects `stableIndex == beginIndex`, so it proves only that removed bars are not reread, not that the configured window is complete.
Recommended remediation: Derive the effective boundary from both provenance sources, e.g. checked arithmetic equivalent to `max(series.beginIndex, sourceUnstable) + barCount - 1`; for covariance use the maximum source boundary. Let `CorrelationCoefficientIndicator` inherit the corrected child boundary.
Regression proof: With nonzero `beginIndex` and a small fixed window, assert exact `begin + window - 1`, assert unavailable at `stable - 1`, and assert the first full-window value at `stable` against an independent oracle.

PR #1504, opened 2026-04-20
feat: add OnChange event system for BarSeries (#1503)

### LINKED ISSUE #1503: Having "OnChnage" event on BarSeires
- [X ] I could not find similar [issues](https://github.com/ta4j/ta4j/issues?utf8=%E2%9C%93&q=)
- [X] I could not find a solution in the [wiki](https://ta4j.github.io/ta4j-wiki/) or [faq section](https://ta4j.github.io/ta4j-wiki/FAQ.html)

Its a bigger change, but may be sinpler enough to incorporate.. While in live market, when you are dealing with indicators, say you want to watch an indicator every minute, you will end up calculating it every minute.

The idea is to implement producer/subscriber model between BarSeries and Indicators, so whenever a Bar series is updated, it pushes the update to every indicator using the same series and and every indicator immediately starts reflecting the new value. This will save a lot of time wasted in backtesting and scalping..
**Describe alternatives you've considered**
None
**Additional context**
None

anandaanv:
Fixes #1503.

- Added `BarSeriesListener` interface with `onBarAdded(int index, Bar bar)` and `onBarReplaced(int index, Bar bar)` callbacks
- Added default `addListener`/`removeListener` methods to `BarSeries` interface (backward compatible)
- `BaseBarSeries` fires listener notifications on `addBar()` using `WeakReference` list (GC-safe, thread-safe via `CopyOnWriteArrayList`)
- `CachedIndicator` auto-subscribes as listener in constructor and eagerly caches values on bar events — enabling incremental O(1) per-bar computation instead of full recalculation

This enables a producer/subscriber model where indicators pre-compute values when bars arrive, which is critical for:
- Live trading with real-time tick updates
- Backtesting performance (avoids redundant indicator recalculation)
- Custom components (ZigZag, market structure) as BarSeries subscribers

- [x] I have run `mvn -B clean license:format formatter:format test install` to format and test my changes per the [contributing guidelines](.github/CONTRIBUTING.md)
- [X] I have added an entry with applicable ticket number(s) to the appropriate unreleased section of `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Unreleased
 
 ### Added
-- **BarSeries listener (OnChange) event system**: `BarSeries` now supports push-based `BarSeriesListener` notifications for bar append and replace events, enabling incremental indicator computation. `CachedIndicator` auto-subscribes to its series and pre-computes values when new bars arrive. Listeners use weak references to avoid memory leaks. Includes `readResolve` serialization support and comprehensive test coverage (`#1503`).
+- **BarSeries listener (OnChange) event system**: `BarSeries` now supports push-based `BarSeriesListener` notifications for bar append and replace events, enabling incremental indicator computation.  Includes `readResolve` serialization support and comprehensive test coverage (`#1503`).
 
 ### Changed
 - **Agent guidance stays accurate across repo and personal workflows**: `scripts/agents_for_target.sh` guidance now clearly describes path-scoped `AGENTS.md` discovery from the current repo/workspace root, so contributors can use it for file-targeted lookup without assuming it also covers personal PR/comment workflow guidance.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+### Added
+- **BarSeries listener (OnChange) event system**: `BarSeries` now supports push-based `BarSeriesListener` notifications for bar append and replace events, enabling incremental indicator computation. `CachedIndicator` auto-subscribes to its series and pre-computes values when new bars arrive. Listeners use weak references to avoid memory leaks. Includes `readResolve` serialization support and comprehensive test coverage (`#1503`).
+
 ### Changed
 - **Agent guidance stays accurate across repo and personal workflows**: `scripts/agents_for_target.sh` guidance now clearly describes path-scoped `AGENTS.md` discovery from the current repo/workspace root, so contributors can use it for file-targeted lookup without assuming it also covers personal PR/comment workflow guidance.
 
--- a/ta4j-core/src/main/java/org/ta4j/core/BarSeriesListener.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BarSeriesListener.java
@@ -1,0 +1,35 @@
+/*
+ * SPDX-License-Identifier: MIT
+ */
+package org.ta4j.core;
+
+/**
+ * Listener for {@link BarSeries} change events. Enables push-based
+ * (producer/subscriber) updates so indicators can pre-compute values
+ * incrementally when new bars arrive.
+ *
+ * @since 0.22.7
+ * @see BarSeries#addListener(BarSeriesListener)
+ */
+public interface BarSeriesListener {
+
+    /**
+     * Called after a new bar has been appended to the series.
+     *
+     * @param index the index of the newly added bar
+     * @param bar   the bar that was added
+     * @since 0.22.7
+     */
+    void onBarAdded(int index, Bar bar);
+
+    /**
+     * Called after the last bar has been replaced (e.g. live tick update).
+     *
+     * @param index the index of the replaced bar
+     * @param bar   the replacement bar
+     * @since 0.22.7
+     */
+    default void onBarReplaced(int index, Bar bar) {
+        onBarAdded(index, bar);
+    }
+}
--- a/ta4j-core/src/main/java/org/ta4j/core/BaseBarSeries.java
+++ b/ta4j-core/src/main/java/org/ta4j/core/BaseBarSeries.java
@@ -11,10 +11,12 @@ import org.ta4j.core.num.Num;
 import org.ta4j.core.num.NumFactory;
 
 import java.io.Serial;
+import java.lang.ref.WeakReference;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 /**
  * Base implementation of a {@link BarSeries}.
@@ -62,6 +64,13 @@ public class BaseBarSeries implements BarSeries {
      * The number of removed bars.
      */
     private int removedBarsCount = 0;
+
+    /**
+     * Weak-referenced listeners notified on bar additions/replacements.
+     *
+     * @since 0.22.7
+     */
+    private final transient List<WeakReference<BarSeriesListener>> listeners = new CopyOnWriteArrayList<>();
 
     /**
      * Convenience constructor for BaseBarSeries minimizing upfront parameter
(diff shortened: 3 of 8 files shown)


ta4j-core/src/main/java/org/ta4j/core/BaseBarSeries.java
```diff
@@ -280,6 +289,13 @@ public void addBar(final Bar bar, final boolean replace) {
         }
         this.seriesEndIndex++;
         removeExceedingBars();
+
+        // Notify listeners
+        if (replace) {
```
TheCookieLab, line 294: This replacement notification is unreachable for the normal non-empty replace path because that branch sets the last bar and returns earlier. That means `addBar(bar, true)` never fires `onBarReplaced` for the live-update case this PR is adding. Please move the replacement notification into that branch and add focused tests for non-empty replace, empty-series `replace=true`, and normal append.

ta4j-core/src/main/java/org/ta4j/core/BaseBarSeries.java
```diff
@@ -63,6 +65,13 @@ public class BaseBarSeries implements BarSeries {
      */
     private int removedBarsCount = 0;
 
+    /**
+     * Weak-referenced listeners notified on bar additions/replacements.
+     *
+     * @since 0.22.7
+     */
+    private final transient List<WeakReference<BarSeriesListener>> listeners = new CopyOnWriteArrayList<>();
```
TheCookieLab, line 73: Because this field is `final transient`, a deserialized series gets `listeners == null`, and `addListener` then becomes a permanent no-op. Please make the listener storage reinitializable after deserialization, for example through a lazy accessor or serialization hook, and cover it with a round-trip test.

ta4j-core/src/main/java/org/ta4j/core/BarSeriesListener.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Listener for {@link BarSeries} change events. Enables push-based
+ * (producer/subscriber) updates so indicators can pre-compute values
+ * incrementally when new bars arrive.
+ *
+ * @since 0.22.7
+ * @see BarSeries#addListener(BarSeriesListener)
+ */
+public interface BarSeriesListener {
```
TheCookieLab, line 14: This is new public core API, but there is no focused coverage for the listener contract. Please add tests for add/remove/dedup/null behavior, append versus replace callbacks, max-bar-count interaction, deserialization, and the CachedIndicator subscription behavior.

ta4j-core/src/main/java/org/ta4j/core/indicators/CachedIndicator.java
```diff
@@ -115,6 +118,23 @@ protected CachedIndicator(Indicator<?> indicator) {
...
+     * Pre-computes and caches the indicator value when a new bar arrives. Delegates
+     * to {@link #getValue(int)} which handles cache management.
+     *
+     * @since 0.22.7
+     */
+    @Override
+    public void onBarAdded(final int index, final Bar bar) {
+        // Trigger lazy computation when accessed via getValue(index)
+        // The try-catch handles indicators not yet ready with sufficient data
+        try {
```
TheCookieLab, line 131: Catching every `Exception` here masks real calculation regressions during bar append. If there is a specific warm-up case to tolerate, let's make that narrow and tested; otherwise this should not silently swallow indicator failures.

TheCookieLab (changes requested):
Requesting changes for now. The event-system direction makes sense, but the current implementation misses the main replace path and adds public core behavior without tests. Please also add a changelog entry for the new listener API before this merges. Details inline.

ta4j-core/src/main/java/org/ta4j/core/BaseBarSeries.java
anandaanv: Fixed — moved `notifyBarReplaced()` inside the early-return replace branch (before the `return`), so it now fires for non-empty replace calls. The append path at the end only calls `notifyBarAdded()`. Added focused tests: `onBarReplaced_firesOnNonEmptyReplace`, `addBar_replaceOnEmpty_addsInsteadOfReplace`, `addBar_appendThenReplaceThenAppend`.

ta4j-core/src/main/java/org/ta4j/core/BaseBarSeries.java
anandaanv: Fixed — changed field from `final transient` to `transient` and added a `readResolve()` hook that reinitializes the list after deserialization. Removed all `listeners == null` guards since `readResolve` guarantees initialization. Added round-trip serialization tests: `listenerWorksAfterDeserialization` and `removeListener_worksAfterDeserialization`.

ta4j-core/src/main/java/org/ta4j/core/BarSeriesListener.java
anandaanv: Added `BarSeriesListenerTest` with 18 focused tests covering: add/remove/dedup/null listener behavior, append vs replace callbacks, empty-series replace-as-append, max-bar-count interaction, deserialization round-trip (add + remove after deserialize), CachedIndicator auto-subscription, and weak reference GC cleanup.

ta4j-core/src/main/java/org/ta4j/core/indicators/CachedIndicator.java
anandaanv: Fixed — removed the try-catch entirely. `onBarAdded` now calls `getValue(index)` directly, letting any calculation errors propagate to the caller rather than being silently swallowed.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## Unreleased
 
+### Added
+- **BarSeries listener (OnChange) event system**: `BarSeries` now supports push-based `BarSeriesListener` notifications for bar append and replace events, enabling incremental indicator computation. `CachedIndicator` auto-subscribes to its series and pre-computes values when new bars arrive. Listeners use weak references to avoid memory leaks. Includes `readResolve` serialization support and comprehensive test coverage (`#1503`).
```
TheCookieLab:
```suggestion
- **BarSeries listener (OnChange) event system**: `BarSeries` now supports push-based `BarSeriesListener` notifications for bar append and replace events, enabling incremental indicator computation.  Includes `readResolve` serialization support and comprehensive test coverage (`#1503`).
```

We no longer do this right?

TheCookieLab (approved):

TheCookieLab:
@anandaanv please run the formatter and commit those changes.

TheCookieLab:
Closing this PR without merge.

The implementation work here was thoughtful, and the review feedback was addressed. The reason I am closing this is not code quality. It is that, after the broader discussion on #1503 and the final shape of this branch, I do not think this belongs in ta4j core in its current form.

The deciding points for me are:

- The original value proposition was automatic indicator updates / eager incremental computation. That part was later backed out because it introduced deadlock risk, and `CachedIndicator` already covers the usual cached path lazily via `getValue()`.
- I am not persuaded that adding `BarSeries` callbacks that ultimately lead back to `getValue()` buys enough over normal application-managed access to justify new public core API.
- More importantly, I think the architectural split here matters: ta4j should own series mutation primitives and indicator semantics, while applications should own market-data ingestion, bar-close clocking, persistence, and downstream orchestration. This PR moves too much of the latter into core.
- What remains is mainly a new listener surface on `BarSeries`. That is a long-lived compatibility commitment, and I do not think we have a strong enough ta4j-owned use case yet to justify it.
- The contract is also uneven across series types: `BaseBarSeries` supports listeners, while `ConcurrentBarSeries` disables them. I am not comfortable baking that split into core without a stronger library-level need.
- More broadly, I agree with the concern raised on #1503 that ta4j should stay focused on technical-analysis primitives and avoid absorbing application-level event orchestration unless the need is clearly library-owned.

I do think the underlying use case may still be valid. If someone wants to reopen this direction, the better path is a narrower proposal:
- start from one concrete ta4j-owned consumer that cannot reasonably live at the application layer,
- prefer a narrow mutation-result primitive over a generic listener surface if that is enough,
- require at least two real consumers before adding a broad public listener API,
- and avoid designs that depend on eager cross-indicator callbacks under concurrent mutation.

So I am closing this PR without merge and leaving #1503 open for further design discussion.

PR #1557, opened 2026-07-07
CF-148 Add portfolio foundation

TheCookieLab:
Fixes CF-148.

- Add org.ta4j.core.portfolio contracts for static target-weight portfolio backtests with aligned asset series, target allocations, rebalance policies, execution snapshots, and value-series export.
- Add deterministic portfolio execution coverage for alignment, allocation validation, transaction costs, cash sleeves, and result ordering.
- Add the runnable StaticPortfolioBacktest example and concise user-facing references.

- [x] I have run `mvn -B clean license:format formatter:format verify install` (or `scripts/run-full-build-quiet.sh`) to confirm this change does not break the build and keeps code style/license formatting consistent.
- [x] I have added an entry to `CHANGELOG.md` for this change.


TheCookieLab:
@coderabbitai full review

TheCookieLab:
@coderabbitai full review

ta4j-core/src/main/java/org/ta4j/core/portfolio/PortfolioExecutor.java
```diff
@@ -0,0 +1,255 @@
...
+        Num transactionCost = series.numFactory().zero();
+        Num turnover = series.numFactory().zero();
+
+        for (PortfolioAsset asset : series.assets()) {
+            Num price = prices.get(asset);
+            Num currentUnits = nextHoldings.get(asset);
+            Num currentNotional = price.multipliedBy(currentUnits);
+            Num targetNotional = portfolioValue.multipliedBy(targetWeight(asset));
+            Num deltaNotional = targetNotional.minus(currentNotional);
+            if (deltaNotional.isNegative()) {
```
TheCookieLab, line 128: Functional correctness: this targets asset notionals from the pre-cost `portfolioValue`, then subtracts transaction costs from cash afterward. That means allocations with an explicit cash sleeve (`totalWeight < 1`) are not actually restored to their configured target weights once costs are applied. Example: a 60%/30% allocation with a 1% transaction cost buys 900 gross, pays 9 in costs, leaves portfolio value 991, and realizes ~90.8% invested / ~9.2% cash instead of 90% / 10%.

Since `PortfolioAllocation` explicitly advertises cash as the residual target, the rebalance math should either solve target notionals against post-cost value or proportionally scale gross buys so the cash sleeve absorbs its intended share after costs. If this drift is intentional, it needs to be called out in the executor/allocation docs and covered by a test.

ta4j-core/src/main/java/org/ta4j/core/portfolio/PortfolioExecutor.java
```diff
@@ -0,0 +1,255 @@
...
+            }
+
+            Num price = prices.get(asset);
+            Num amount = desiredGross.dividedBy(price);
+            Num cost = transactionCostModel.calculate(price, amount);
+            Num requiredCash = desiredGross.plus(cost);
+            if (requiredCash.isGreaterThan(nextCash)) {
+                Num affordableGross = nextCash.minus(cost);
+                if (affordableGross.isNegativeOrZero()) {
+                    continue;
```
TheCookieLab: This affordability fallback subtracts `cost` calculated for the original `desiredGross`, then recalculates cost after reducing the gross. That happens to work for simple proportional costs, but it is not safe for arbitrary `CostModel` implementations, especially fixed/minimum/step costs. It can skip an otherwise affordable trade or accept a trade only because the second cost happens to fit.

Given that the API accepts any ta4j `CostModel`, consider isolating this into an `affordableGross(...)` helper that solves for `gross + cost(gross) <= cash` for the supplied model, or narrow the documented support to proportional trade-cost models for this first portfolio slice.

ta4j-core/src/main/java/org/ta4j/core/portfolio/PortfolioExecutionResult.java
```diff
@@ -0,0 +1,136 @@
...
+     * @since 0.22.9
+     */
+    public Map<PortfolioAsset, Num> finalWeights() {
+        PortfolioSnapshot finalSnapshot = finalSnapshot();
+        return finalSnapshot.prices()
+                .keySet()
+                .stream()
+                .collect(java.util.stream.Collectors.toUnmodifiableMap(asset -> asset, finalSnapshot::assetWeight));
+    }
+}
```
TheCookieLab, line 139: This loses the portfolio's deterministic asset order because `Collectors.toUnmodifiableMap(...)` does not preserve encounter order. Most of the new package is careful to expose assets in deterministic portfolio order, so `finalWeights()` should probably return weights in `series.assets()` order as well.

A small `LinkedHashMap` loop followed by `Map.copyOf` would still not guarantee iteration order; use `Collections.unmodifiableMap(linkedHashMap)` or another order-preserving immutable wrapper if the public result-map iteration order is part of the intended contract.

TheCookieLab (commented):
Reviewed for portfolio accounting correctness, ta4j convention adherence, documentation quality, and general API/dev-practice concerns. CI is green on the current head, and the overall slice is coherent, but I left a few comments where the public contract can drift from behavior in cost/cash cases or where deterministic ordering is advertised but not consistently preserved.
- reviewed and posted by ChatGPT

TheCookieLab (commented):
— ChatGPT

ta4j-core/src/main/java/org/ta4j/core/portfolio/AlignedPortfolioSeries.java
```diff
@@ -0,0 +1,261 @@
...
+        Map<PortfolioAsset, BarSeries> sourceSeriesByAsset = new LinkedHashMap<>();
+        for (PortfolioSeries portfolioSeries : sourceSeries) {
+            sourceSeriesByAsset.put(portfolioSeries.asset(), portfolioSeries.series());
+        }
+        return Map.copyOf(sourceSeriesByAsset);
+    }
+
+    private static Map<Instant, Integer> indexesByEndTime(PortfolioAsset asset, BarSeries source) {
+        Map<Instant, Integer> indexesByEndTime = new HashMap<>();
+        for (int index = source.getBeginIndex(); index <= source.getEndIndex(); index++) {
```
TheCookieLab, line 249: [Major] Make terminal retained-index iteration overflow-safe

A legal one-bar source series whose retained begin/end index is `Integer.MAX_VALUE` fails during alignment here. The terminal bar is processed once, then `index++` wraps to `Integer.MIN_VALUE`; `index <= source.getEndIndex()` is still true, so the next iteration calls `getBar` with an invalid index.

Use a terminal-safe loop that processes `endIndex` and breaks before incrementing past it, or iterate by snapshot offset instead of an inclusive `int` counter. Add a regression test that aligns two one-bar series at the same end time with one built via `withBeginIndex(Integer.MAX_VALUE)`, then assert construction completes and its source index remains `Integer.MAX_VALUE`. The constructor-first replacement in stacked PR #1583 contains the same loop and needs the same correction.

TheCookieLab:
## API direction: one obvious portfolio workflow before the first release

Reviewed the foundation at `501e67053e40e919501bae84b81067f4cce5a22b`, including the replacement series/manager/result in stacked #1583. The useful core is already here: explicit target weights, cash, strict alignment, costs, and inspectable results. I would make that small workflow effortless rather than add another abstraction layer.

### Converge the existing stack, rather than publish two successive APIs

#1583's `PortfolioSeries` → `PortfolioAllocation` → `PortfolioSeriesManager.run(...)` → immutable result is the better supported direction. In particular, putting series/cost configuration on the manager and allocation/capital on `run(...)` makes comparisons natural and resembles the existing single-series workflow. Bring that constructor-first core into the first releasable change, remove the superseded **unreleased** asset/pair/alignment/executor entrypoints, and keep the examples on that one surface. Do not add a compatibility facade for an API that has not shipped. Preserve genuinely released APIs.

Reconcile the currently conflicted foundation with master and carry the inline accounting/snapshot/export fixes into the surviving implementation; renaming the executor does not fix them. This is not a request to finish every optimizer, Yahoo loader, chart, spreadsheet, or AI report before the static-allocation core can ship. Those consumers should remain optional and consume the same result.

### A one-risk-asset-plus-cash portfolio should use the same workflow

Please reconsider the `at least two assets` restriction in both drafts. A 60% equity / 40% cash benchmark is a portfolio experiment with the same allocation, rebalance, fee, and reporting semantics as the two-risk-asset version. Sending that user to the strategy/trading-record API changes the experiment; adding a dummy second series can even change the strict intersection of dates. Supporting one nonempty asset series is the natural degenerate case, not a separate feature framework. Include that baseline in the accounting tests.

### Make initial investment and scheduled rebalancing unambiguous

Today `RebalancePolicy.onIndexes(Set.of(3))` leaves the portfolio entirely in cash through indexes 0–2: the predicate controls initial investment as well as subsequent rebalances. The name/Javadoc does not make that distinction clear. Keep #1583's obvious buy-and-hold default, and explicitly define whether a supplied schedule replaces the initial investment or only schedules later rebalances. Preserve an explicit delayed-investment path rather than silently forcing bar zero for every custom policy. Show the distinction in a small example/test. Calendar examples should use aligned timestamps, not imply that every N retained bars means a calendar month.

### Make an unexpected result explain itself

When a rebalance is intentionally skipped or partial, expose a compact outcome/reason and achieved-versus-target allocation; costs and turnover alone cannot explain why the requested portfolio was not restored. Keep this proportional to the actual partial-execution behavior—no order-management or execution-provider framework. Cash should also be easy to inspect alongside asset weights.

The getting-started example should fit on one screen and state its accounting assumptions next to `run(...)`: fractional long-only holdings, trades/valuation at aligned closes, one common quote currency with any required adjustments supplied upstream, residual cash, and no implicit forward fill. Give the aligned date range/count and a compact result summary by default; make full snapshot history explicit. Include one performance-analysis example that agrees with the result's net total return, including initial fees. Name gross traded notional versus a turnover ratio distinctly, and label fractional versus percentage returns.

**First-release acceptance:** the same short API handles buy-and-hold, a cash sleeve, a scheduled rebalance, and a one-asset benchmark; repeated runs over the owned snapshot are reproducible; target/cash/cost accounting reconciles; and exported performance agrees with the result. That would make this feel like ta4j rather than a separate mini-platform.

ta4j-core/src/main/java/org/ta4j/core/portfolio/PortfolioExecutor.java
```diff
@@ -0,0 +1,353 @@
...
+    private Num rebalanceCost(Map<PortfolioAsset, Num> holdings, Map<PortfolioAsset, Num> prices,
+            Num targetPortfolioValue) {
+        Num totalCost = series.numFactory().zero();
+
+        for (PortfolioAsset asset : series.assets()) {
+            Num price = prices.get(asset);
+            Num currentNotional = price.multipliedBy(holdings.get(asset));
+            Num targetNotional = targetPortfolioValue.multipliedBy(targetWeight(asset));
+            Num gross = targetNotional.minus(currentNotional).abs();
+            totalCost = totalCost.plus(transactionCost(price, gross));
```
TheCookieLab: [P1] Do not turn solver residue into a new fixed-fee rebalance

`postCostPortfolioValue()` stops within `valueTolerance`, but this cost calculation and the execution loops classify every nonzero notional difference as a trade. The residual from the previous solve therefore triggers full fixed/minimum fees on the next bar even when prices have not changed.

Concrete fixture: initial cash 1000, both closes constantly 100, weights 0.6/0.3 (10% cash), `FixedTransactionCostModel(5)`, and `everyBar()`. Following the current arithmetic gives portfolio values approximately **990 → 980 → 970**: two fees are charged on each unchanged bar. After the initial solve, holdings differ from the target only by solver residue; charging those differences makes the next solve shrink the target again and creates real sales to fund unnecessary fees.

Use one consistent, scale-aware no-trade tolerance when determining the active trades, estimating their costs, and executing them. Merely rounding the reported weights will not fix the accounting. Add a multi-bar flat-price regression: after the initial allocation, holdings/value remain unchanged and subsequent turnover/cost are zero within the declared tolerance. Cover both numeric backends and carry the correction into #1583's manager.

ta4j-core/src/main/java/org/ta4j/core/portfolio/PortfolioExecutor.java
```diff
@@ -0,0 +1,353 @@
...
+        for (PortfolioAsset asset : series.assets()) {
+            Num price = prices.get(asset);
+            Num currentUnits = nextHoldings.get(asset);
+            Num currentNotional = price.multipliedBy(currentUnits);
+            Num targetNotional = targetNotionals.get(asset);
+            Num deltaNotional = targetNotional.minus(currentNotional);
+            if (deltaNotional.isNegative()) {
+                Num gross = affordableSellGross(price, deltaNotional.abs(), nextCash);
+                if (gross.isZero() || gross.isNegative()) {
+                    continue;
```
TheCookieLab: [P2] Reconsider skipped sells after other sells supply the required cash

This permanent `continue` makes asset presentation order an undocumented execution priority. A sale whose fee needs cash is skipped before a later cash-generating sale, although executing the latter first would make the complete rebalance affordable without ever allowing negative cash.

The existing `fixedTransactionCostsSkipSellThatWouldSpendCashNegative` fixture demonstrates it: initial cash 1000; ALPHA 100→101, BETA 100→100; weights 0.01/0.99; fixed fee 50. At the second bar the planned sales are approximately 1.0891 and 98.9109. ALPHA-first skips the small sale and ends with **48.9109 cash, 50 fees, NAV 850.09**. Reversing only series encounter order allows both sales and ends with approximately **zero cash, 100 fees, NAV 800.09**, restoring the target. The current test asserts the first ordering's partial result but does not check the permutation.

Plan/sequence cash-generating sells before cash-consuming sells, or revisit deferred sells after funding becomes available. If a rebalance genuinely cannot complete, give it an explicit partial/skipped outcome rather than silently treating it as target restoration. Add a permutation regression asserting identical per-asset holdings, cash, and fees for this collectively feasible trade set. Preserve display ordering separately.

ta4j-core/src/main/java/org/ta4j/core/portfolio/PortfolioSeries.java
```diff
@@ -0,0 +1,72 @@
...
+    public BarSeries series() {
+        return snapshotSeries(series);
+    }
+
+    static BarSeries snapshotSeries(BarSeries barSeries) {
+        BarSeries source = Objects.requireNonNull(barSeries, "barSeries");
+        return new BaseBarSeriesBuilder().withName(source.getName())
+                .withNumFactory(source.numFactory())
+                .withBeginIndex(source.getBeginIndex())
+                .withBars(source.getBarData())
```
TheCookieLab: [P1] Snapshot the bar contents, not just the list containing them

`getBarData()` and `BaseBarSeriesBuilder.withBars(...)` copy the list but retain the same mutable `Bar` instances. Consequently, `source.addPrice(...)` after constructing this pair changes the supposedly snapshotted close price. Mutating the value returned by `pair.series()` or `aligned.getSeries(asset)` has the same effect. `AlignedPortfolioSeries.getBar(...)` additionally returns the internal bar directly, so fixing only input copying would still leave a mutation path.

This lets the same aligned input/executor produce different results after an unrelated update to the original series or a returned view; the defensive-looking API does not actually own its price history.

Detach the retained bar contents at the ownership boundary and prevent mutable owned bars from escaping through the public accessors, while preserving source indexes/timestamps/numeric values. Reuse a suitable existing snapshot primitive if available rather than creating a public framework. Add regressions that mutate the original series and each exposed series/bar view after construction, then verify that aligned prices and repeat executions are unchanged. The replacement `PortfolioSeries.snapshotSeries` and `getBar` in #1583 retain this same issue.

ta4j-core/src/main/java/org/ta4j/core/portfolio/PortfolioExecutionResult.java
```diff
@@ -0,0 +1,139 @@
...
+                .withNumFactory(initialCash.getNumFactory())
+                .build();
+        PortfolioAsset firstAsset = series.assets().getFirst();
+        Num zero = initialCash.getNumFactory().zero();
+        for (PortfolioSnapshot snapshot : snapshots) {
+            Num value = snapshot.portfolioValue();
+            valueSeries.barBuilder()
+                    .timePeriod(series.getBar(firstAsset, snapshot.index()).getTimePeriod())
+                    .endTime(snapshot.endTime())
+                    .openPrice(value)
```
TheCookieLab: [P2] Preserve the initial-capital baseline in the performance bridge

Every exported point is already post-rebalance, so this series cannot represent the first snapshot's return from `initialCash`. On a flat two-asset run with initial cash 1000, fully invested weights, a 1% proportional transaction cost, and `atStart()`, `totalReturn()` is about **−0.990099%**, but the exported closes are approximately `[990.0990099, 990.0990099]`. Close-to-close return and drawdown analysis of that curve both see zero; the initial trading loss has disappeared. A one-bar run has no interval from which it can recover that loss at all.

Either include an explicit initial-capital observation with a well-defined time convention, or provide a baseline-aware return/analysis adapter and document that this raw snapshot series alone excludes the initial interval. Avoid silently inventing a market bar or double-charging costs in a synthetic trading record.

Add an integration regression proving that compounded exported returns, including the initial interval, reconcile with `totalReturn()` for nonzero initial fees, and that the initial loss is visible to drawdown analysis. The existing export test checks only count/end time/final close with zero costs, so it misses this discrepancy.

ta4j-core/src/main/java/org/ta4j/core/portfolio/PortfolioExecutor.java
```diff
@@ -0,0 +1,353 @@
...
+    private Num transactionCost(Num price, Num gross) {
+        if (gross.isNegativeOrZero()) {
+            return series.numFactory().zero();
+        }
+        return transactionCostModel.calculate(price, gross.dividedBy(price));
+    }
+
+    private Num valueTolerance(Num reference) {
+        Num epsilon = series.numFactory().epsilon();
+        return epsilon.max(reference.abs().multipliedBy(epsilon));
```
TheCookieLab: [P2] Do not use a dimensionless epsilon as an absolute cash tolerance

The `max(epsilon, ...)` floor makes the solve depend on the units in which prices/capital are expressed. The API accepts any positive finite DecimalNum capital, but when it is smaller than `1e-12`, the post-cost solve can stop after its first midpoint regardless of the relative error.

For example, with starting cash `1e-13`, constant positive prices `1e-14`, weights 0.6/0.4, and a 1% proportional fee, the current arithmetic chooses target value `5e-14` and leaves about **49.75% of actual NAV in cash**, despite the fully invested target. Multiplying prices and capital by a common factor to ordinary-sized values produces the intended near-zero cash weight instead.

Scale monetary tolerances to the positive reference amount and handle numeric stagnation/precision limits explicitly, rather than flooring them at a fixed monetary amount. The same helper governs buy/sell affordability, so keep those semantics consistent. Add a DecimalNum scaling regression that rescales prices and starting capital together and verifies unchanged asset weights, cash weight, and relative transaction costs.

TheCookieLab (commented):
Reviewed `501e67053e40e919501bae84b81067f4cce5a22b` for accounting correctness, snapshot ownership, ta4j composition, and first-use ergonomics. **I would not release this version yet.** Five new actionable findings are attached inline; the broader API/convergence recommendations are in https://github.com/ta4j/ta4j/pull/1557#issuecomment-5819785901.

The most consequential problems are recurring fixed fees on a flat portfolio and shared mutable bars behind the snapshot API. The sell pass also makes the economic outcome depend on asset encounter order, and the performance export drops the initial-capital baseline. There is an additional scale-dependent stopping-tolerance issue for small monetary units.

The existing retained-index overflow finding remains applicable at `AlignedPortfolioSeries.indexesByEndTime`; I have not duplicated that thread: https://github.com/ta4j/ta4j/pull/1557#discussion_r3776652296.

I also inspected the replacement `PortfolioSeries`, `PortfolioSeriesManager`, and `PortfolioExecutionResult` in #1583 at `6b692aa73f57612c28e5c40c625d7a07da99dd86`: these five underlying issues survive there, so apply the fixes and regressions to the converged API rather than only to classes being removed. This is not a full review of #1583's analytics/reporting additions.

**Validation:** source review plus independent Python Decimal reproductions of the executor's arithmetic for flat fixed-fee rebalancing, reversed asset order, and small-unit scaling. These are source-equivalent numerical probes, **not executions of the Java implementation or JUnit tests**. GitHub reports the existing Run Verify and associated validation runs successful for this head (Run Verify: https://github.com/ta4j/ta4j/actions/runs/30222004633); I did not run a fresh local Maven/JUnit build. The requested regressions should be run against the implementation after the fixes and stack reconciliation.

PR #1558, opened 2026-07-08
Add FXMacroData read data integration

roberttidball:
## Summary
- Expands the FXMacroData integration beyond release-calendar-only helpers to the public read/data API surface.
- Covers catalogue discovery, macro announcements/latest/changes, predictions, release calendar, FX spot, COT, commodities/latest, curves/rates/forwards, market sessions, risk sentiment, news, press releases, and GraphQL where the host repo exposes a generic client shape.
- Keeps the existing calendar helper as a compatibility wrapper and supports `FXMACRODATA_API_KEY` / `FXMD_API_KEY` for authenticated datasets.

## Validation
- `git diff --check` passed across the external integration batch.
- Changed Python files were formatted with `black` and checked with `python -m py_compile`.
- Live smoke checks passed against production for catalogue, calendar, announcements, predictions, FX, COT, commodities latest, market sessions, and risk sentiment using a local API key without printing it.

## Local environment notes
- Java and .NET builds were not run locally because this machine does not have `javac`, `JAVA_HOME`, or `dotnet` configured.
- Some full package imports still require repo-specific generated files or optional dependencies in an installed environment.

### DIFF
--- a/ta4j-examples/src/main/java/ta4jexamples/datasources/FxMacroDataClient.java
+++ b/ta4j-examples/src/main/java/ta4jexamples/datasources/FxMacroDataClient.java
@@ -1,0 +1,243 @@
+/*
+ * SPDX-License-Identifier: MIT
+ */
+package ta4jexamples.datasources;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URLEncoder;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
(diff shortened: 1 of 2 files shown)


TheCookieLab:
Thanks, while I appreciate the contribution intent the best place for this is in your client application where FXMacroData (presumably) is an intentional part of your particular analysis and trading workflow. I'm not a FXMacroData user so I'm not comfortable with maintaining this code going forward. The existing data provider API code in ta4j-examples is incidental, it exists solely to drive examples rather than real API support.

roberttidball:
Thanks for the clear feedback. I agree this belongs in an application-level example rather than adding a provider you would have to maintain, so I am withdrawing this PR.

PR #1581, opened 2026-07-23
Share deterministic indicator state by series

TheCookieLab:

 - **Transparent series-scoped indicator sharing**: Standard bar series now share calculation state across structurally equivalent deterministic indicators
- Backtest managers, executors, walk-forward results, and position-sizing contexts expose snapshot-safe read-only series views that still delegate indicator sharing to their retained series snapshots. 
- In non-toy 4000 bar series JVM fleet sweeps now run ~18–25x faster (from ~105s to 6s) and with roughly 20x less retained heap (from ~500 MB to 31.4 MB).

- [x] I have run `mvn -B clean license:format formatter:format verify install` (or `scripts/run-full-build-quiet.sh`) to confirm this change does not break the build and keeps code style/license formatting consistent.
- [x] I have added an entry to `CHANGELOG.md` for this change.


TheCookieLab:
@coderabbitai full review

PR #1593, opened 2026-08-02
Add transparent GPU acceleration for Monte Carlo backtests

TheCookieLab:
Linear: CF-336, CF-107.

- transparently accelerate eligible existing `BarSeriesManager` Monte Carlo forecast backtests with omitted/`off` or `auto`, preserving scalar fallback and exposing a primitive-only provider SPI
- keep this PR scoped to the acceleration runtime and eligible forecast contracts; named-rule, drawdown-alias, and CLI work remains outside this PR

Public API rationale: `OperationPlanner`, `OperationDecoder`, and `PlannedOperation` bridge the core-owned forecast package and acceleration runtime package, and let optional provider-module qualification exercise that same lowering/decoding pipeline without duplicating it. They are public for this cross-package extension seam; providers still consume only immutable primitive `KernelRequest` values, never indicators or domain objects. The existing `MonteCarloKernel` is reused for separate raw-state advancement and output mixing so native implementations can reproduce successive version-1 draws without a second random-stream abstraction.

The `ta4j-cli` module and its native Metal/CUDA/OpenCL providers continue in #1610; merge this PR first, then retarget that one to `master`.

## Summary by CodeRabbit

- **New Features**
  - Added opt-in acceleration for supported indicator calculations, with automatic fallback for unsupported, stale, or memory-intensive workloads.
  - Added deterministic per-path Monte Carlo randomness through `ta4j.forecast.rngVersion=1`; default behavior remains compatible with existing results.
  - Added safeguards against excessive host-memory usage during accelerated simulations.
  - Generated trading bars now maintain strictly increasing timestamps.

- **Bug Fixes**
  - Improved Monte Carlo input validation and handling of invalid or incomplete forecast data.
  - Preserved stable behavior when acceleration providers cannot complete a calculation.

Exact price-summary regression ownership: `stronglyNonnormalPathsMatchDirectPriceSampleSummary` compares mean, median, standard deviation and quantiles exactly in the active `Num` domain against the transformed price samples. Removed the duplicate rounded-literal fixture: `exp(log(0.9))`/`exp(log(1.1))` produce standard deviation 10.000000000000007 (DoubleNum) / 10.00000000000001 (DecimalNum), not exactly 10. No tolerance was widened; the stronger exact-sample regression remains.


TheCookieLab:
@coderabbitai full review

TheCookieLab:
@coderabbitai review

TheCookieLab:
@coderabbitai review

TheCookieLab:
@coderabbitai full review

TheCookieLab:
@coderabbitai full review

TheCookieLab:
@coderabbitai full review

TheCookieLab:
@coderabbitai full review

TheCookieLab:
@coderabbitai full review

ta4j-core/src/main/java/org/ta4j/core/acceleration/AccelerationRuntime.java
```diff
@@ -0,0 +1,481 @@
...
+ * the property, or setting it to {@code off}, retains the ordinary scalar path
+ * without provider discovery.
+ *
+ * <p>
+ * This deliberately is not a general indicator graph compiler. Providers must
+ * reject every graph family they do not understand, and scalar
+ * {@link Indicator#getValue(int)} evaluation remains the fallback oracle.
+ *
+ * @since 0.23.1
+ */
```
TheCookieLab, line 48: [Major] Move the supported provider SPI out of an `internal` package

Location:
`AccelerationRuntime` and its public nested `Provider`, `Request`, `Result`, `Backend`, `Status`, `Diagnostic`, and `Scope` contracts.

Violated principle/invariant:
One concept must have one obvious home, and public extension points must communicate that they are supported contracts.

What happens:
The type is public, documented with `@since`, consumed by a separate provider module, and deliberately exported from the OSGi bundle, yet its package is `org.ta4j.core.internal.acceleration` and the class Javadoc says applications do not call it directly. The PR description also calls the SPI “internal,” which is incompatible with an externally implemented and exported ServiceLoader contract.

Operator/downstream consequence:
Provider authors must import an API explicitly named internal. Tools and consumers may exclude it, and any later move to a supported package becomes a source, binary, service-descriptor, and OSGi compatibility break.

Evidence:
`ta4j-cli` implements `AccelerationRuntime.Provider` through `META-INF/services`, and `AccelerationOsgiExportTest` prevents this package from being hidden.

Recommended remediation:
Place the provider-facing types in a supported package such as `org.ta4j.core.acceleration.spi` (or move the whole supported facade to `org.ta4j.core.acceleration`) and keep only implementation mechanics under `.internal`. Align Javadoc, OSGi exports, and the service descriptor with that boundary before release.

Regression proof:
Compile and ServiceLoader-run a minimal external provider against the published core artifact, and add an API/OSGi check that no externally implemented public SPI is exported from an `.internal` package.

ta4j-core/src/main/java/org/ta4j/core/acceleration/AccelerationRuntime.java
```diff
@@ -0,0 +1,481 @@
...
+            CachedBatch cached = batches.get(indicator);
+            if (cached == null) {
+                cached = evaluate(indicator, index);
+                if (cached == null) {
+                    scalarFallback.add(indicator);
+                    return Optional.empty();
+                }
+                batches.put(indicator, cached);
+            }
+            Object value = cached.value(index);
```
TheCookieLab: [Critical] Invalidate cached batches when the series changes

Location:
`AccelerationRuntime.Context.value(...)` after the first provider result is stored in `batches`.

Violated principle/invariant:
A cached forecast must describe the current series revision. Supported mutation must never reuse stale values.

What happens:
`CachedBatch` contains only its start index and values. After the first accelerated lookup fills the remaining range, every later lookup returns `cached.value(index)` without checking `BarSeries#getBarSeriesChangeSnapshot`, history revision, retained range, or a provider snapshot. The provider-side `SeriesStamp` in #1610 protects capture and publication only; a replacement, clear/rebuild, prune, or concurrent mutation after publication but before a later cached read is invisible here.

Operator/downstream consequence:
A strategy can operate on forecasts computed from obsolete prices while the scalar indicator cache has already invalidated, producing a different and factually stale trading record under `auto`.

Recommended remediation:
Bind each `CachedBatch` to one coherent series change snapshot. Before every cached read, compare the current revision/range snapshot; on any relevant change, discard the batch and re-evaluate or fall back to scalar. For series that report unsupported revision tracking, use a conservative policy rather than assuming the batch remains valid.

Regression proof:
Populate a multi-index batch, then replace the current bar, replace a retained historical bar, clear/rebuild reused indexes, and advance a maximum-bar-count series between two reads. In every case, assert that the old batch is not served and that `auto` produces the same decisions/trading record as `off`. Include a concurrent mutation probe around publication and the next cached read.

ta4j-core/src/main/java/org/ta4j/core/indicators/forecast/MonteCarloSimulation.java
```diff
@@ -88,6 +101,36 @@ int getHorizon() {
         return settings.horizon();
     }
 
+    private static boolean legacyStreamRequested() {
+        String configured = System.getProperty(RNG_VERSION_PROPERTY);
+        if (configured == null || configured.isBlank()) {
+            return false;
```
TheCookieLab: [Major] Preserve released scalar forecasts when acceleration is omitted

Location:
`legacyStreamRequested()` and the default branch selected by `project(...)`.

Violated principle/invariant:
An optional transparent acceleration change must not silently alter the scalar/off-mode oracle.

What happens:
When `ta4j.forecast.rngVersion` is absent or blank, this returns `false`, so every existing seeded Monte Carlo forecast switches from the released shared `SplittableRandom` sequence to the new per-path stream. This happens even when `ta4j.acceleration` is omitted or explicitly `off`; the changelog therefore records intentionally different results for an ostensibly transparent acceleration PR.

Operator/downstream consequence:
Existing fixed-seed forecasts, quantiles, threshold decisions, regression fixtures, and trading records change merely by upgrading ta4j, even for users who never enable acceleration.

Recommended remediation:
Keep version 0 as the compatibility default. Require an explicit RNG-version opt-in for the per-path stream, and let `auto` fall back to scalar while version 0 is selected unless a backend can reproduce it. Treat changing the default stream as a separate, clearly migrated behavioral change rather than an implementation prerequisite for acceleration.

Regression proof:
Pin the pre-change golden forecast and a threshold-sensitive trading record with both properties absent and with `ta4j.acceleration=off`; require exact equality. Separately prove version 1 scalar/native parity when it is explicitly selected.

TheCookieLab (commented):
[Major] Complete the scope split before merging the acceleration core

Location:
PR-wide: `NamedAssetRegistry` aliases, the new public `NamedRule` registry/scanner, example named rules, and drawdown-criterion alias fixes.

Violated principle/invariant:
Focused-change discipline and public-API restraint. The transparent `BarSeriesManager` acceleration layer should be independently reviewable and shippable, with CLI-enabling surface owned by the stacked CLI work.

What happens:
This PR still introduces an unrelated public workflow based on global registration and classpath scanning. None of the acceleration runtime, manager, forecast adapter, or compatibility code depends on `NamedRule` or the example rules; those additions support the rule/catalog workflows in #1610.

Operator/downstream consequence:
The core optimization remains coupled to a second permanent API and its discovery/serialization behavior. A defect or design disagreement in named-rule support can block acceleration, and merging #1593 commits CLI-specific surface even when #1610 is deferred.

Evidence:
The changed-file set contains the named-rule implementation and examples alongside the acceleration code, while #1610 is stacked on this exact head and owns the CLI rule workflows.

Recommended remediation:
Move `NamedRule`, the `NamedAssetRegistry` alias additions, the example named rules/tests, and the drawdown alias fixes into #1610 or a small prerequisite PR. Keep #1593 limited to the manager scope, acceleration contract/runtime, eligible forecast adapter, compatibility behavior, tests, and release notes for only what this PR ships.

Regression proof:
Rebase #1610 onto the reduced #1593 head and run the canonical full gate. The public API diff for #1593 should contain only the deliberate acceleration contract, and its core tests should pass without any named-rule or CLI artifacts.

— ChatGPT

TheCookieLab:
@coderabbitai resume

TheCookieLab:
@coderabbitai review

TheCookieLab:
@coderabbitai full review

TheCookieLab:
Scope split complete (re: PRR_kwDOBiw0jc8AAAABJrOn1w): `NamedRule`, the registry scanner, and example named rules already live in the stacked CLI PR #1610; the residual read-only `NamedAssetRegistry.aliases(...)` view moved there too in 351b66124, so this branch's net `named/` delta is a single +14-line regression test pinning master's drawdown alias-to-package mapping. Proof: `git diff origin/master...HEAD -- '**/named/'` at cd4db834f shows only that test file.

TheCookieLab:
[Major] Replace concrete-indicator probing with an operation-level acceleration plan

Location:
`AccelerationRuntime.Provider`, `AccelerationRuntime.Request`, and `MonteCarloPriceForecastIndicator#getValue()` / `accelerationSpec()`.

Violated principle/invariant:
Public extension points must be minimal and capability-based, and an optimization boundary must define its semantic, numeric, workload, and fallback contracts before provider execution.

What happens:
The transport API is generic, but the calculation contract is not. `Request<T>` contains only an arbitrary `Indicator<T>` plus an index range, while `Provider#evaluate` returns fully reconstructed ta4j values. A provider therefore has to recognize the concrete indicator/graph, obtain indicator-specific public state, repeat series/numeric/RNG/window validation, decide whether offload is worthwhile, lower inputs, execute, and reconstruct domain values. The follow-on #1610 implementation already has to cast to `MonteCarloPriceForecastIndicator`, call `accelerationSpec()`, duplicate these checks, and maintain another Monte-Carlo-specific provider/capability model.

A public `Accelerable` marker would not solve this. Acceleratability is not a stable property of an indicator class: it depends on the semantic operation, requested range and batch shape, provider/device, numeric representation and precision, transfer and reconstruction cost, resident data, branch/dependency structure, scratch memory, determinism/tolerance contract, and the best available CPU algorithm. The useful unit is an operation over a batch, for example paths × horizons × decision indexes, pairs × windows × distance-matrix cells, queries × candidates × features, lags × samples, alignments × DP cells, or fit candidates × observations.

Operator/downstream consequence:
Every backend must independently understand ta4j object graphs and can disagree about eligibility or result semantics. Each new target tends to add another permanent public `*Spec`/accessor family solely for providers, while custom/composed indicators remain unportable. Correctness logic and output reconstruction are duplicated outside the owning calculation, making provider parity and future evolution hard to verify.

Evidence:
At exact head `6c4207e5176a74d5582645117738ff6cd512b56b`, the provider receives `Request(indicator, fromInclusive, toInclusive)`, and Monte Carlo opts in through its own `getValue()` override and public `accelerationSpec()`. Structurally different high-value targets already demonstrate why one marker is insufficient: distance correlation is a pairwise O(W²) matrix/reduction; analog forecasting is a batched nearest-neighbor/top-k search; lead/lag correlation scans independent lags; DTW is dependency-bound unless many alignments are batched; and LPPL exposes independent calibration candidates rather than an intrinsically “accelerable indicator.”

Recommended remediation:
Do not add a public `Accelerable` interface. Introduce a core-owned lowering layer that converts a supported calculation plus exact range into an internal `PreparedAcceleration<T>` / `AccelerationPlan<T>`. It should own:

- a versioned semantic operation identifier;
- exact input, range, output-shape, missing-value, ordering/tie, RNG, precision, and determinism contracts;
- immutable source snapshots and workload dimensions;
- input/output/scratch-memory estimates;
- the scalar reference path and a core-owned decoder that reconstructs `Num`, `Forecast`, or other domain values.

Providers should consume a prepared, indicator-agnostic kernel request and return raw kernel output; they should never cast concrete indicators or construct ta4j domain values. Keep the lowerers, decoder, and plan internal. If the provider boundary must remain public because native providers live in another module, expose only a minimal immutable, versioned `KernelRequest` / `KernelResult` SPI containing operation ID, shapes, numeric encoding, and read-only primitive buffers—not indicator-specific specs or graph types.

Prove the abstraction with Monte Carlo and at least one structurally different operation (preferably distance correlation or analog neighbor search) before freezing the public SPI. Candidate implementation order after measurement should be: (1) distance-correlation range/pair batches, (2) analog nearest-neighbor range batches, (3) LPPL coarse-grid calibration, and (4) wide/batched lead-lag profiles. DTW and Bayesian change-point detection should wait for an explicit many-alignment/many-series batch envelope; ordinary moving averages, RSI/MACD/ATR/Bollinger, scalar Kalman/CUSUM, and other small recursive calculations should not be targeted individually.

Regression proof:
- Monte Carlo and a pairwise/neighbor-search operation use the same provider-facing request/result contract without provider-side indicator casts.
- Unsupported graph, precision, RNG, missing-value, or deterministic-output requirements reject with a typed reason before native initialization.
- Accelerated/scalar parity covers the unstable boundary, nonzero begin index, non-finite input, deterministic seeds or ties, and current/historical series mutation.
- Declared peak memory is rejected before allocation, and a stale snapshot falls back safely.
- Every-index end-to-end benchmarks include lowering, host/device transfer, kernel launch, output reconstruction, and snapshot validation—not only kernel time.

— ChatGPT

TheCookieLab:
[Major] Select CPU and accelerator backends through a non-initializing workload assessment

Location:
`AccelerationRuntime.providers()` and `AccelerationRuntime.Context#evaluate(...)`.

Violated principle/invariant:
Backend selection must be deterministic from explicit capability and complete workload cost, not incidental implementation naming; resource limits must be checked before expensive initialization or allocation.

What happens:
Providers are sorted by provider class name, then queried in that order until one returns `EXECUTED`. The result model can describe `CPU_FASTER`, provider unavailability, and multiple accelerator backends, but the provider contract has no separate support/cost/memory assessment stage.

Operator/downstream consequence:
When Metal, CUDA, OpenCL, or multiple implementations coexist, lexical class order—not predicted end-to-end cost or device capability—can select the backend. Renaming or repackaging a provider can change execution behavior, a slower provider can win, and the runtime cannot compare CPU against accelerator crossover or reject excessive device memory in one consistent place. The follow-on implementation consequently invents its own capability and crossover model outside this runtime.

Evidence:
At exact head `6c4207e5176a74d5582645117738ff6cd512b56b`, `providers()` sorts loaded providers by `provider.getClass().getName()`, and `Context#evaluate` accepts the first valid executed result. No provider priority, supported-operation inventory, predicted total duration, data-residency state, or peak-memory estimate is available to the runtime before execution.

Recommended remediation:
Split provider interaction into assessment and execution:

```java
Assessment assess(KernelRequest request);   // must not initialize native code
KernelResult execute(KernelRequest request);
```

`Assessment` should provide an explicit decision/reason, backend/device identity, supported operation/version and numeric contract, predicted end-to-end accelerated cost, applicable CPU baseline/crossover, peak device memory, and determinism/accuracy capability. Rank supported providers by predicted total cost with a stable documented tie-break; choose CPU when no provider clears a safety margin; execute the best candidate; and fall back to the next assessed provider after an isolated execution failure while retaining per-attempt diagnostics.

The cost model must include lowering, host/device copies, launch/synchronization, kernel work, result transfer, reconstruction, and snapshot validation. It should distinguish cold execution from warm/resident execution instead of publishing one global minimum batch size.

Regression proof:
- Fake providers whose lexical order opposes their predicted cost select the faster provider.
- Equal assessments use a stable documented tie-break.
- A failed/unavailable best provider falls back to the next candidate without reusing partial output.
- CPU wins below the measured crossover and acceleration wins above it.
- Assessment does not initialize native code or allocate device buffers.
- Peak-memory rejection happens before allocation, and diagnostics identify every assessed/skipped/failed backend.

— ChatGPT

TheCookieLab:
[Major] Establish the operation-level ABI before merging the core acceleration boundary

Location:
`AccelerationRuntime`, the Monte Carlo lowering path, and the SPI consumed by #1610.

The earlier operation-planning and non-initializing assessment findings are the accepted architecture direction. Do not add `Accelerable` to `Indicator`, `MonteCarloMethod`, or concrete indicators: acceleration eligibility belongs to a versioned operation over a particular workload, not to a class.

Required shape before merge:

1. Core-internal operation planners recognize supported domain calculations and lower them into immutable, versioned kernel requests.
2. Cross-module providers receive an operation key/version, workload shape, numeric/accuracy/determinism contracts, primitive inputs, output shape, and pre-allocation memory estimate—not `Indicator`, `BarSeries`, `NumFactory`, or `Forecast`.
3. Provider assessment is separate from execution and must not initialize native state. Assess all usable providers, then select by predicted end-to-end time with an explicit stable tie-break; provider class name and host OS must not become the routing policy.
4. Core owns semantic eligibility, exact series snapshot/revision validation, raw-result validation, scalar fallback, and reconstruction through the owning `NumFactory`.
5. Unknown/custom graphs and unsupported `MonteCarloMethod` implementations fall back to scalar without probing or loading native code.
6. Failure isolation and quarantine are keyed by provider + device + operation/version, not by provider globally.
7. The first v1 operation is the qualified Monte Carlo shock-path kernel. Additional operations are admitted only after a better-CPU-baseline audit and measured end-to-end crossover evidence.

Ownership split:

- **#1593:** runtime, operation contracts, core lowering/decoding, cache/snapshot/fallback behavior, and provider-selection tests.
- **#1610:** optional Metal/CUDA/OpenCL kernels, packaging, device qualification data, and CLI diagnostics.
- **#1617:** keep the public Monte Carlo method seam accelerator-neutral; expose only internal canonical method-graph information needed by the core lowerer.

Regression proof:

- unsupported and below-threshold requests do not initialize native code;
- two available providers are assessed and the lower predicted total-cost provider wins independently of registration/class-name order;
- one provider execution failure falls through to the next qualified provider before scalar fallback;
- malformed or partial raw output never reaches domain consumers;
- append, current-bar replacement, historical replacement, clear/rebuild, and retained-head advance invalidate affected prepared work;
- a custom `MonteCarloMethod` remains source-compatible and scalar-correct while declining native lowering.

— ChatGPT

TheCookieLab:
@coderabbitai incremental review

TheCookieLab:
@coderabbitai review

TheCookieLab:
@coderabbitai review

TheCookieLab:
@coderabbitai review

TheCookieLab:
@coderabbitai full review

TheCookieLab:
@coderabbitai review

TheCookieLab:
@coderabbitai full review

TheCookieLab:
@coderabbitai full review

TheCookieLab:
@coderabbitai review

TheCookieLab:
The current 931369f45 scope contains the core operation-level runtime, planner/decoder contracts and forecast integration. Native classifier packaging and named-rule/CLI reconstruction are owned by stacked PR #1610, not this PR; those fixes are being validated there. The snapshot-before-planning finding is fixed here and its three regression cases failed on the pre-fix runtime and pass on this head (full gate: 9900 tests, no failures/errors/skips). The canonical summary still attributes those stack-owned issues to this branch although no review threads remain open.

TheCookieLab:
@coderabbitai resume

TheCookieLab:
@coderabbitai full review

Please reconcile the current-head disposition with the scope clarification above: native packaging and named-rule/CLI reconstruction belong to stacked PR #1610.

ta4j-core/src/main/java/org/ta4j/core/acceleration/OperationDecoder.java
```diff
@@ -0,0 +1,32 @@
...
+ * <p>
+ * Decoders run inside the acceleration scope after the runtime validates raw
+ * shapes and series snapshots. They map primitives through the owning
+ * {@link NumFactory} and must return {@code null} only via an exception: a
+ * {@code null} return is treated as malformed provider output.
+ *
+ * @since 0.25.1
+ */
+@FunctionalInterface
+public interface OperationDecoder {
```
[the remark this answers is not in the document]
TheCookieLab: Public visibility is required by the existing cross-package core boundary: MonteCarloShockPathPlanner is package-private in org.ta4j.core.indicators.forecast while the runtime is in org.ta4j.core.acceleration. Package-private OperationDecoder/PlannedOperation would not compile for that core caller. Added this rationale to both Javadocs; native providers still receive only KernelRequest primitives, never planners/decoders/domain graphs.

scripts/tests/test_review_examples_compile.sh
```diff
@@ -0,0 +1,80 @@
...
+        else if (c == "\\") escaped = 1
+        else if (c == "\"") in_string = 0
+      } else if (c == "\"") {
+        in_string = 1; arg = arg c
+      } else if (c == "(") {
+        depth++; arg = arg c
+      } else if (c == ")") {
+        depth--; arg = arg c
+        if (depth == 0) {
+          if (arg !~ /"/) { print "statement at line " NR " has no string argument"; failed = 1 }
```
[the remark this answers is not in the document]
TheCookieLab: Removed this source-text scanner rather than extending a partial Java parser. Maven test compilation already rejects missing contains arguments and unbalanced Java syntax before tests execute; the handwritten scanner duplicates that compiler check and falsely constrains valid source forms. The full contributor gate remains the executable proof.

TheCookieLab:
@coderabbitai review

Head 2a99d3093 now rejects the reported 25,000 × 5,000 host workload before allocation (the 128 MiB smoke previously failed with OutOfMemoryError and now falls back). Public cross-package core types are documented; the duplicate source parser was removed in favor of compiler verification. Please update the current-head disposition; all matching threads are addressed.

TheCookieLab:
@coderabbitai Please reconcile the canonical Merge Risk summary with your verification at https://github.com/ta4j/ta4j/pull/1593#issuecomment-5584366039. You confirmed all current-head findings addressed, and no review threads remain open, but the summary still cites unresolved native packaging and named-rule reconstruction from the separate stacked PR #1610. Please identify a concrete remaining finding in this PR or update its current-head disposition.

ta4j-core/src/main/java/org/ta4j/core/indicators/forecast/MonteCarloShockPathPlanner.java
```diff
@@ -0,0 +1,214 @@
...
+        double[] variances = new double[size];
+        double[] windows = new double[windowSize];
+        Indicator<Num> priceIndicator = forecast.kernelPriceIndicator();
+        ReturnForecastStateIndicator<? extends ReturnMomentState> stateIndicator = forecast.kernelStateIndicator();
+        ReturnIndicator returnIndicator = stateIndicator.getReturnIndicator();
+        for (int row = 0; row < size; row++) {
+            int index = fromInclusive + row;
+            if (!snapshotRow(index, priceIndicator, returnIndicator, stateIndicator, factory, settings, prices, means,
+                    drifts, variances, windows, row)) {
+                return null;
```
TheCookieLab, line 119: [Major] Preserve the scalar first-stable index before lowering a batch

**Location:** the row loop / `snapshotRow`, compared with `MonteCarloSimulation.project()`.

**Invariant:** enabling acceleration must not turn an unstable scalar forecast into a stable forecast.

**Evidence and consequence:** scalar simulation rejects `index < max(stateWarmup, returnWarmup + lookback - 1)`, but this planner only checks state stability, retained-window length, and finite return values. With ordinary `LogReturnIndicator` (`getCountOfUnstableBars() == 1`, but `getValue(0) == 0`), a 252-return lookback, and an already-warm EWMA state, origin **251** passes this planner: it includes the synthetic return at index 0. Scalar origin 251 is unavailable; its first stable origin is 252. Since `MonteCarloPriceForecastIndicator.getValue()` asks the runtime before scalar evaluation, starting a scope/backtest at 251 can publish a stable native forecast and change trading decisions one bar early.

**Remediation:** apply the canonical simulation/forecast eligibility boundary before capturing rows. Do not treat a finite warm-up sentinel as a real observed return, and preserve unavailable prefixes when selecting an eligible range.

**Regression proof:** use the real `LogReturnIndicator`, a 252-bar lookback and a warm state; compare acceleration off/on at 251 and 252 with a deterministic test provider. Assert the exact stability boundary, not only finite sample equality at later indexes.

— ChatGPT

ta4j-core/src/main/java/org/ta4j/core/acceleration/AccelerationRuntime.java
```diff
@@ -0,0 +1,859 @@
...
+                    diagnostic = new Diagnostic(DiagnosticCode.INVALID_RESULT, candidate.providerId,
+                            failureMessage(exception));
+                    continue;
+                }
+                effectiveBackend = candidate.assessment.backend();
+                providerInUse = candidate.providerId;
+                diagnostic = new Diagnostic(DiagnosticCode.ACCELERATED, providerInUse,
+                        candidate.assessment.backend().name().toLowerCase(Locale.ROOT) + "/"
+                                + candidate.assessment.deviceId() + " executed " + request.operation());
+                return new CachedBatch(request.fromInclusive(), decoded, after);
```
TheCookieLab, line 708: [Major] Revalidate the snapshot after decoding and before publishing

**Location:** `Context.evaluate()` after `decodeAll()`.

**Invariant:** every published accelerated value must still belong to the captured series revision.

**Evidence and consequence:** `after` is captured and checked immediately after native execution, then `decodeAll()` constructs every result (including per-forecast sample normalization/sorting). A historical replacement, head advance, or clear/rebuild during that work is not checked again. The returned `CachedBatch` retains the earlier `after` stamp, and `Context.value()` immediately stores and returns its first value without another `matchesCurrentSeries()` call. Only a subsequent read detects the mutation, so the first stale forecast can already affect the trading record.

**Remediation:** check the captured revision/range again after successful decoding, before returning the batch. Discard stale decoded work with `STALE_SERIES`; a source mutation must not quarantine a healthy provider. Keep the check in the shared publication boundary rather than duplicating it in native backends.

**Regression proof:** install a deterministic provider and a decoder that replaces a retained bar during its first decode. The initial runtime read must decline/recompute, never return the pre-mutation result. Also test head advance and clear/rebuild, and verify later reads recover. This deterministic decoder hook needs no timing-dependent race or GPU.

— ChatGPT

ta4j-core/src/main/java/org/ta4j/core/acceleration/AccelerationRuntime.java
```diff
@@ -0,0 +1,859 @@
...
+            CachedBatch cached = batches.get(indicator);
+            if (cached != null && !cached.matchesCurrentSeries(indicatorSeries)) {
+                batches.remove(indicator);
+                cached = null;
+            }
+            if (cached == null) {
+                cached = evaluate(indicator, index);
+                if (cached == null) {
+                    scalarFallback.add(indicator);
+                    return Optional.empty();
```
TheCookieLab, line 595: [Major] Keep a transient warm-up miss from disabling the whole run

**Location:** `Context.value()` and its scope-long `scalarFallback` set.

**Invariant:** the ordinary constructor-first backtest path should accelerate later eligible work without requiring callers to manually skip an indicator's warm-up.

**Evidence and consequence:** every null result from `evaluate()` permanently adds the indicator to `scalarFallback`. A normal `new BaseStrategy(entryRule, exitRule)` starts with zero strategy-level unstable bars. Its first forecast read can therefore occur at index 0; the Monte Carlo planner declines the incomplete lookback, and every later read in that `BarSeriesManager` scope bypasses planning—even after the forecast becomes fully stable. A temporary invalid row or stale snapshot has the same effect. This makes `auto` silently ineffective for a realistic normal user journey; manually synchronizing strategy warm-up becomes incidental acceleration plumbing.

**Remediation:** distinguish permanent unsupported/unavailable decisions from index/revision-dependent declines. Handle warm-up locally or retain an earliest retry index/revision, rather than caching all declines for the scope. Avoid blindly rebuilding the entire remaining batch on every unstable bar.

**Regression proof:** run a default `BaseStrategy` whose entry rule reads a forecast from the beginning of a sufficiently long series. Assert scalar behavior during warm-up, then a provider execution after stability in the same run. Add a temporary-invalid-window/recovery case and verify attempts remain bounded.

— ChatGPT

ta4j-core/src/main/java/org/ta4j/core/acceleration/AccelerationRuntime.java
```diff
@@ -0,0 +1,859 @@
...
+                    diagnostic = new Diagnostic(DiagnosticCode.PROVIDER_FAILURE, fallbackId, failureMessage(exception));
+                    LOG.debug("Provider {} assessment failed: {}", fallbackId, exception.getMessage());
+                    continue;
+                } finally {
+                    suspended = false;
+                }
+                if (assessment == null || !assessment.supported() || !assessment.deterministic()
+                        || assessment.predictedTotalNanos() < 0 || Math.max(request.peakDeviceBytesEstimate(),
+                                assessment.peakDeviceBytes()) > memoryLimitBytes) {
+                    continue;
```
TheCookieLab, line 736: [Major] Preserve the provider's actionable decline diagnostic

**Location:** `Context.assess()` and the empty-candidate branch in `evaluate()`.

**Invariant:** distinguish an absent provider from an installed provider declining precision, workload, or memory; operators must be able to recover without reading implementation code.

**Evidence and consequence:** an `Assessment.unsupported(...)` is discarded here without recording `assessment.diagnostic()`. If all providers decline, `evaluate()` replaces the reason with `NO_PROVIDER / no provider supports ...`. In #1610 the installed Metal provider deliberately returns a message naming `ta4j.acceleration.approximateTolerance` when an exact request cannot run, and another actionable message for a missing native classifier. Both are lost by this runtime. The README's documented recovery message therefore cannot reach the runtime diagnostic even though the provider supplied it.

**Remediation:** retain typed assessment/attempt reasons and select a truthful final fallback diagnosis. Reserve `NO_PROVIDER` for empty discovery; preserve provider/device attribution and the relevant property/library/memory detail. A small bounded attempt record or decisive-decline policy is sufficient—no new operator framework is needed.

**Regression proof:** install one fake provider returning `PROVIDER_UNAVAILABLE` with an accuracy opt-in message. Assert that `lastDiagnostic()` preserves that code, provider and message rather than reporting `NO_PROVIDER`. Repeat for missing-library and memory declines, plus a failing provider followed by a successful sibling.

— ChatGPT

TheCookieLab (commented):
## Operator-experience and production-readiness re-review

**Profile verdict: applied** (`cookielab-review-profile-v2.md`). Reviewed `ta4j/ta4j`, base `master@240deb4c236dc296bc301bb3c4801f3d86a58366`, head `2a99d30939bea70be7fe30355ebba4cfee74828c`; head rechecked before submission. Changed-file inventory: 19 files. Deep review focused on the runtime/provider boundary, core Monte Carlo lowering/decoding, RNG compatibility, normal indicator/strategy call sites, mutation handling, and owning tests. Applied root, core, and indicator-scoped repository guidance and the current CF-336/CF-107 PRDs.

**Verdict:** correctness and production readiness **not ready**; API architecture **aligned**; joy to use **almost**; accelerated/scalar behavioral compatibility **conditional on the inline fixes**.

The main architectural correction is present: providers now receive primitive operations, core owns reconstruction, and custom methods can remain scalar. The public planner/decoder bridge has an explicit cross-package rationale; this review does not ask for another framework, an `Accelerable` marker, or more public configuration. The remaining problems are in the actual first-use, recovery, and publication paths, not the overall direction.

| Surface / journey | Assessment |
|---|---|
| Normal `MonteCarloPriceForecastIndicator` → rule → `BaseStrategy` → `BarSeriesManager.run` | Existing domain entry points remain intact, but an initial warm-up miss disables acceleration for the rest of the run. |
| `OperationPlanner` / `PlannedOperation` / `OperationDecoder` | Appropriate ownership for lowering and reconstruction; the planner must preserve the scalar first-stable boundary. |
| `KernelRequest` / `Provider` / `KernelResult` | Primitive-only extension boundary retained; native numerical qualification is reviewed separately in #1610. |
| Failure and diagnostics | Typed provider rejection details are discarded, making normal setup failures misleading. |
| Mutation/concurrency | Snapshot capture before planning is fixed; mutation during decoding still permits a stale first result to escape. |
| Compatibility/custom composition | Legacy/default RNG and custom-method fallbacks were inspected in source. No external compilation or full serialization matrix was run here. |

Four concrete findings are inline, each with a targeted regression probe. Earlier inline conversations were already resolved when this review began; I have not reopened remediated architecture or API-visibility findings.

**Verification:** source-based review, not a local build or GPU qualification run. GitHub reports `Run Verify` run `34219712243` successful for this head. The review environment could not obtain a local checkout because repository DNS access failed; the proposed regression probes have not been executed against ta4j here. The existing planner tests use a zero-warm-up fixed return source, so they do not exercise the ordinary `LogReturnIndicator` boundary described below.

— ChatGPT

TheCookieLab (commented):
## Operator-experience and production-readiness review

Profile verdict: **applied** (`cookielab-review-profile-v2.md`). Reviewed `ta4j/ta4j`, base `master@240deb4c236dc296bc301bb3c4801f3d86a58366`, head `2a99d30939bea70be7fe30355ebba4cfee74828c`; head rechecked immediately before submission. Scope: the acceleration runtime, planner/decoder/provider contracts, Monte Carlo interception and scalar compatibility, and the corresponding normal, warm-up, mutation, and fallback journeys. This is a source review, not a claim of a locally executed full gate or native qualification.

The primitive-only provider boundary and core-owned reconstruction address the earlier architectural direction. The remaining operator friction is behavioral, not a reason to add another abstraction: normal warm-up must not disable the feature permanently, and a CPU fallback must explain the actual reason.

**Correctness:** conditional. **Production readiness:** not ready for a dependable automatic-acceleration experience. **API architecture:** aligned with the operation-level direction. **Joy to use:** almost. **Compatibility:** runtime/native parity remains unverified by execution in this review.

### [Major] Keep temporary window declines out of the permanent scalar-fallback set

**Location:** `ta4j-core/src/main/java/org/ta4j/core/acceleration/AccelerationRuntime.java`, `Context.value()` and `Context.evaluate()`; `MonteCarloShockPathPlanner.plan()` / `snapshotRow()`; `MonteCarloPriceForecastIndicator.getValue()`.

**Violated invariant:** The ordinary value-at-index path must recover after warm-up or temporary unavailable inputs. A request-specific decline is not proof that the indicator is permanently unsupported.

**What happens:** With RNG version 1 selected, `MonteCarloPriceForecastIndicator.getValue()` asks the runtime before the scalar calculation. The planner returns null when a row has an unstable state or an incomplete return lookback. `Context.value()` treats every null evaluation identically and adds the indicator to `scalarFallback`, which is then consulted before any later evaluation in the same scope. Consequently, the first forecast read during warm-up disables acceleration for every later stable index in that backtest. A temporary invalid row in the requested tail has the same scope-wide effect.

**Operator consequence:** An otherwise eligible backtest can remain entirely scalar merely because its rules evaluated the forecast before the forecast's own warm-up completed. Users would need to manually coordinate strategy/run warm-up with an internal acceleration requirement, contrary to the transparent common path.

**Evidence:** `Context.value()` unconditionally executes `scalarFallback.add(indicator)` after `evaluate()` returns null; `snapshotRow()` rejects unavailable moments and a lookback whose start precedes the retained begin index; the `getValue()` interception only bypasses removed indexes and the legacy RNG mode, not warm-up.

**Recommended remediation:** Distinguish permanent structural declines from temporary index/window declines. Keep warm-up reads on the scalar path without poisoning later eligibility, and retry after a meaningful eligibility/index/revision change. For a bad row later in the range, use a bounded eligible batch or another documented recovery policy rather than permanently rejecting the whole indicator. Avoid replacing the sticky decline with a full-tail rescan on every bar.

**Regression proof:** In one enabled scope, install a counting fake provider and read a warm-up index followed by a stable index. Assert the first result preserves scalar instability and the later eligible range executes through the provider. Add a same-scope invalid-window/recovery case. Independent calls to `planner.plan()` are insufficient to exercise this runtime state bug.

### [Major] Preserve the reason an installed provider declined the request

**Location:** `ta4j-core/src/main/java/org/ta4j/core/acceleration/AccelerationRuntime.java`, `Context.assess()` and the empty-candidate branch of `Context.evaluate()`.

**Violated invariant:** Optional fallback must remain actionable and distinguish missing providers from unsupported precision, insufficient memory, unqualified workload, and execution failure.

**What happens:** `Assessment` carries a typed diagnostic, but `assess()` immediately continues for `!assessment.supported()` without retaining that diagnostic. When no candidate survives, `evaluate()` commonly replaces the outcome with `NO_PROVIDER` / `"no provider supports ..."`. The original provider ID and specific remedy are lost. Supported assessments rejected for determinism or memory also disappear without their own explanation.

**Operator consequence:** A correctly installed provider can decline for a specific accuracy or qualification requirement, yet normal diagnostics resemble a provider-installation problem. This makes the central 'why did auto stay on CPU?' workflow unnecessarily opaque and can send users toward reinstalling a working backend.

**Evidence:** The unsupported-assessment branch does not read `assessment.diagnostic()`; the empty-candidate branch synthesizes `NO_PROVIDER` unless the accumulated code is `CPU_FASTER` or `PROVIDER_FAILURE`.

**Recommended remediation:** Retain bounded, provider-attributed assessment outcomes and expose a factual fallback summary through the existing diagnostic path. Reserve the no-provider-discovered outcome for an empty discovery result. Preserve actionable unsupported/memory/accuracy explanations and distinguish malformed assessments from ordinary unsupported requests. Do not introduce a second diagnostic framework.

**Regression proof:** Install a fake provider returning a distinctive unsupported diagnostic and verify that the completed scope reports that reason and provider rather than `NO_PROVIDER`. Keep a separate zero-provider fixture, and test multiple declines so selection/order does not erase their explanations.

### Verification limits

Repository review guidance and the supplied review profile were applied to the inspected surfaces. No local build, complete test suite, GPU execution, or crossover benchmark was run during this pass; source-level counterexamples above are the basis for the findings. Existing unrelated discussions are not implicitly cleared by this review.

— ChatGPT

ta4j-core/src/main/java/org/ta4j/core/indicators/forecast/MonteCarloShockPathPlanner.java
TheCookieLab: Boundary now enforced explicitly: the planner lowers only `[max(fromInclusive, forecast.getCountOfUnstableBars()), toInclusive]`, so origin 251 stays scalar while the eligible suffix from 252 accelerates.

One correction to the premise: on this head `LogReturnIndicator.getValue(0)` returns `NaN`, not `0` (`calculate(0)` guards `index < getCountOfUnstableBars()`), so the previous planner declined origin 251 through its finite-return check rather than publishing a stable forecast. The gap was still real for any return source whose unstable prefix is finite, so the regression test uses that shape (`getCountOfUnstableBars() == 1` with finite values, 252-return lookback, real `LogReturnIndicator` + warm state): origin 251 is excluded from the batch and only 252+ is lowered. Both the planner test and the runtime boundary test fail without the clamp.

Addressed in 7a33ed12c.

TheCookieLab:
@coderabbitai incremental review

Current head covers the four operator-experience findings (scalar first-stable boundary, post-decode snapshot revalidation, transient-decline recovery, provider decline diagnostics), the non-synchronized test override, and a merge of `master` into the branch. Please review the exact current head.

TheCookieLab:
@coderabbitai incremental review

Head 73f03bff6 applies a fresh-eyes improvement pass: accelerated forecasts now populate the indicator cache, memory-bounded chunking with continuation, truthful planner decline reasons and post-run `lastDiagnostic()`, resilient provider discovery, build-time `rngVersion` resolution, and consolidated Monte Carlo tests. The V1 provider wire contract is unchanged.

TheCookieLab:
@coderabbitai incremental review

TheCookieLab:
@coderabbitai incremental review

TheCookieLab:
@coderabbitai incremental review

This head changes the unreleased `MONTE_CARLO_SHOCK_PATHS_V1` contract: kernels return cumulative log-returns (core applies exp and terminal guards), the Silverman factor travels as a parameter, and inputs use one contiguous returns buffer. A reference kernel test pins bitwise parity with the scalar lane across all shock models and volatility modes.

PR #1613, opened 2026-08-23
Speed up equity-curve criteria and add opt-in batch evaluation via EquityBundle

TheCookieLab:
Fixes #

**Why this matters:** any workflow that evaluates several criteria over the same trading record — report generation, walk-forward leaderboards, parameter scans, custom dashboards — currently rebuilds the identical equity cash flow once per criterion. Four ratio/drawdown criteria means four identical sweeps of the same series. This PR removes that redundancy, both inside the existing classes (helping every existing caller, no API changes needed) and through a new opt-in batch entry point for callers who evaluate criteria as a set.

- **Multi-criterion evaluation gets measurably faster**: after the single-sweep rework, evaluating a 9-criterion set over a 1096-bar record on `DecimalNum` drops from 6 ms to 4 ms (~1.5×) with the batch facade; the saving grows linearly with the number of equity-curve criteria evaluated together. The underlying class-level fixes are larger still: the standard metrics bundle went from 679 ms to 23 ms on `DecimalNum` (29.5×) and 58 ms to 23 ms on `DoubleNum`, benefiting all existing call sites unchanged.
- **One-call batch API**: `CriteriaEvaluation.evaluateAll(series, tradingRecord, criteria)` returns an insertion-ordered criterion-to-value map, computing each distinct curve configuration exactly once via a shared `EquityBundle`. Criteria participate by implementing `EquityBundleAware` (all eight equity-curve criteria in core do); everything else falls back to its existing behavior, so arbitrary criterion mixes just work.
- **Identical results, proven**: batch output equals individual evaluation for every migrated criterion across closed/open/empty records on both Num factories, and the SHA-256 digest harness is byte-identical to pre-change baselines. Full build green (ta4j-core 7338 tests, ta4j-examples 753, SpotBugs clean).
- **One flagged contract change**: `CashFlow`/`CumulativePnL`/`Returns` capture their series snapshot at construction and return it from every `getBarSeries()` call, instead of tracking later live-series replacement. Existing mutation-isolation semantics are retained.

Implementation detail lives in the two CHANGELOG entries and commit messages.


ta4j-core/src/main/java/org/ta4j/core/criteria/CriteriaEvaluation.java
```diff
@@ -0,0 +1,94 @@
...
+ * every equity-curve-based criterion. Batch evaluation builds each distinct
+ * curve once and shares it across all participating criteria, which support
+ * this transparently. Criteria without shared-curve support are evaluated as if
+ * {@link AnalysisCriterion#calculate(BarSeries, TradingRecord) calculate} had
+ * been called directly, so arbitrary criterion mixes produce exactly the same
+ * values as individual evaluation.
+ * </p>
+ *
+ * @since 0.24.2
+ */
```
TheCookieLab: [Major] Integrate shared scoring into the established execution-result workflow

**Violated principle:** one workflow/orchestration surface per problem family; a new public API should improve the demonstrated common journey.

**What happens:** this facade is not used by production code. `BacktestExecutionResult#getTopStrategies(...)` and `TradingStatementExecutionResult#rankTradingStatements(...)` still run nested loops that call `criterion.calculate(barSeries, tradingRecord)` independently. Those are the report/leaderboard workflows cited as motivation, so the common path gets no shared-curve reuse unless callers abandon or duplicate the established workflow and discover this separate class. This facade also omits the existing `AnalysisWindow`/`AnalysisContext` variants, creating an overlapping but incomplete criterion-evaluation API.

**Operator/downstream consequence:** ta4j acquires two orchestration paths with divergent feature coverage, while the existing path users already call misses the incremental batch optimization.

**Recommended remediation:** create one internal evaluation scope per `(BarSeries, TradingRecord)` and route the existing scoring/ranking/report paths through it. Then remove this public top-level runner (or retain only a package-private helper). If standalone batch evaluation remains a demonstrated requirement, expose it from the same domain owner used by those workflows rather than adding a parallel utility.

**Regression proof:** exercise `getTopStrategies(...)` and `rankTradingStatements(...)` with multiple equity criteria and assert identical values/order plus one construction per distinct curve key; include a custom `AnalysisCriterion` proving the ordinary `calculate(...)` fallback runs once and retain the existing windowed-analysis fixtures.

ta4j-core/src/main/java/org/ta4j/core/criteria/EquityBundleAware.java
```diff
@@ -0,0 +1,40 @@
...
+ * removes redundant curve construction when several criteria are evaluated
+ * together via
+ * {@link CriteriaEvaluation#evaluateAll(BarSeries, TradingRecord, AnalysisCriterion...)}.
+ * Criteria that do not implement this interface fall back to their regular
+ * two-argument calculation in batch evaluation.
+ * </p>
+ *
+ * @since 0.24.2
+ */
+public interface EquityBundleAware {
```
TheCookieLab: [Major] Collapse the optimization SPI instead of publishing bundle-awareness

**Violated principle:** public API is permanent debt; implementation mechanics should remain internal, and inputs should not be restated when one object already owns them.

**What happens:** the optimization adds three public top-level concepts (`CriteriaEvaluation`, `EquityBundle`, and `EquityBundleAware`), two runner overloads, the bundle lifecycle/cache/accessor API, and a public three-argument `calculate(...)` overload on all eight participating criteria. Yet `EquityBundle` explicitly says its typical use is indirect. This method also repeats `series` and `tradingRecord` already captured by the bundle, making contradictory calls representable and requiring `requireInputsFor(...)` solely to reject them.

**Operator/downstream consequence:** cache invalidation, concurrency, snapshot/freeze behavior, identity checks, and dual calculation paths become permanent extension contracts even though users only asked to evaluate criteria efficiently. Every future equity criterion must now understand and maintain this plumbing.

**Recommended remediation:** keep an internal `EquityCurveCache` and preserve `AnalysisCriterion` as the public extension contract; built-ins can use an internal/protected evaluation hook, while arbitrary custom criteria fall back to their existing two-argument calculation. If standalone batch analysis truly warrants a public concept, collapse all three types into one domain object, preferably:

```java
TradingRecordAnalysis analysis = TradingRecordAnalysis.of(series, tradingRecord);
Map<AnalysisCriterion, Num> scores = analysis.evaluate(criteria);
```

That object owns the inputs and evaluation, while curve caching remains hidden. `CriterionEvaluationContext` is a reasonable lower-level alternative; `EquityAnalysisContext` works only if curve access is intentionally public. If an interface is retained despite this, `EquityCurveCriterion` is more accurate than `EquityBundleAware`, but removing the interface is the cleaner design.

**Regression proof:** add a public-API inventory/compile fixture showing only the chosen single concept (or no new concept), verify existing custom criteria compile unchanged, and retain exact sequential-versus-shared result tests for all built-ins.

TheCookieLab (commented):
Public API architecture: redesign required; see inline findings.

— ChatGPT

PR #1563, opened 2026-07-12
Preserve absolute indices when restoring bar series

TheCookieLab:
Fixes #.

Changes proposed in this pull request:
- add `withBeginIndex(int)` to base and concurrent bar-series builders
- preserve absolute retained indices through appends, pruning, subseries, and serialization
- support intentional series clearing for fail-safe strategy reinitialization

- [x] I have run `mvn -B clean license:format formatter:format verify install` (or `scripts/run-full-build-quiet.sh`) to test my changes per the [contributing guidelines](.github/CONTRIBUTING.md)
- [x] I have added an entry with applicable ticket number(s) to the appropriate unreleased section of `CHANGELOG.md`

## Summary by CodeRabbit

* **New Features**
  * Added support for restoring bar series with a configurable starting index, preserving absolute indices during appends, pruning, sub-series creation, and persistence.
  * Added `clear()` support to reset retained bars and indices while preserving series configuration.
  * Added equivalent functionality for concurrent bar series.

* **Bug Fixes**
  * Prevented read-only indicator views from clearing their underlying series.

* **Documentation**
  * Updated the changelog with the new retained-series behavior.

PR #1618, opened 2026-08-25
Preserve series begin index in BarSeriesManager snapshots

TheCookieLab:
Fixes #.

- Preserve the series begin index in `BarSeriesManager.snapshotSeries()` by restoring `.withBeginIndex(series.getBeginIndex())`, so series constructed with a non-zero begin index (including terminal `Integer.MAX_VALUE` ranges) keep their logical index range and index-addressed strategies/rules fire as expected instead of silently producing no positions.
- Add terminal-index overflow regression tests for the snapshot path (entry-only assertions via `getOpenPositions()`).
- Silence intentionally-triggered failure-path logging in test output: pin four noisy loggers to `error` in `log4j2-test.xml`, make `TraceTestLogger` restore the logger config it overrides, and replace a test `printStackTrace` with a rethrow.

## Summary by CodeRabbit

- **New Features**
  - Analysis results support rolling and offset series, seeded returns, trailing exits, and same-bar returns.
  - Cash-flow, return, cumulative P&L, and invested-interval results expose materialized values through streams.
  - Backtest and walk-forward results provide stable, immutable snapshots.
  - Backtesting and analysis reflect live series updates while preserving captured result windows.

- **Bug Fixes**
  - Improved concurrent-series handling, retention windows, logical analysis bounds, and holding-cost calculations.
  - Prevented overflow-related failures in terminal ranges and position scans.

- **Tests**
  - Added regression coverage for rolling windows, terminal indexes, overflow protection, and immutable results.

### Latest review fixes and verification

Batch execution holds one read lease across baseline, strategy runs, and final result capture. ConcurrentBarSeries work runs on the lease-owning thread to avoid queued-writer/worker-reader deadlocks; ordinary-series scheduling remains parallel.

Walk-forward results snapshot retained history. Materialized analysis streams now use captured windows (including trailing raw exits), and terminal cash-flow consumers use nonwrapping iteration. Shared buffer reuse avoids per-position raw-list copies; seeded undefined returns and empty logical/raw-backed windows are covered. Stream regression assertions fail on the previous source and pass after the fix.

Windows full gate: Tests run 9988; Failures 0; Errors 0; Skipped 0. Log: `C:/Users/David/Workspace/ta4j-org/ta4j/.agents/worktrees/pr1618/.agents/logs/full-build-20260908-032404.log`.

### Test-support reuse and visibility

`ConstrainedSeriesSupport` is public only because analysis, criteria, manager, Elliott, and Wyckoff tests in separate packages share its constrained-series factories and deterministic lock probes; it is test-only, not a production API. Reuse audit: `ConcurrentBarSeriesTest.RecordingReadWriteLock` is private and observes replacement write-lock ownership, not gaps between read leases; extending the existing shared support avoids duplicating fixtures or exposing production constructors.


TheCookieLab:
@coderabbitai full review

TheCookieLab (commented):
## Coverage

- **Profile verdict:** `applied`
- **Exact target:** `ta4j/ta4j` base `5816c1a2eea5ebf8860c51634e071354af68e898` → head `399a9dc1144b846efb531ff60739f7d2d7f604ea`
- **Scope:** all 29 changed files, with emphasis on the `BarSeries`/`BaseBarSeries`/`ConcurrentBarSeries` snapshot contract, manager/context/result ownership, constrained execution ranges, offset-indexed analysis, and downstream criteria.
- **Repository instructions:** root plus scoped core-main/core-test `AGENTS.md` files.
- **Verification:** exact head and current review threads were rechecked immediately before submission. All hosted checks on this head are green. I did not run a separate local checkout/build; the semantic findings below use exact-head source, the introducing commit history, and deterministic counterexamples.

## Executive verdict

- **Correctness:** not ready
- **Production readiness:** not ready
- **API architecture:** redesign required
- **Joy to use:** no
- **Compatibility:** migration required
- **Risk summary:** this branch is fixing the consequences of a false ownership premise. A `BarSeries` is an identity-bearing collaborator already embedded throughout the strategy/rule/indicator graph, not a value owned by `BarSeriesManager`. Copying only the manager side creates two market-data realities. Reusing that copy mechanism for stable result capture then spreads index-window, mutability, concurrency, and terminal-arithmetic concerns across unrelated classes. The continuing churn is evidence of the missing boundary, not a need for a third copier.

**Recommendation: stop extending this branch and reset the implementation around the contracts below.** The original manager/analysis copies were introduced by CF-236 commits `1624827` and `bfa7ed2` to eliminate SpotBugs representation-exposure warnings. They were not a domain requirement. PR #1563's absolute-index support exposed the latent inconsistency, but did not create it.

## Findings

### [Blocker] Restore one coherent series graph

**Location:** `BarSeriesManager`, `PositionSizer.Context`, and any execution callback that receives a strategy plus series.  
**Violated principle/invariant:** every signal decision, execution price, sizing decision, and result must describe one series identity/revision.  
**What happens:** the strategy passed to `run()` remains bound to the caller's original series, while execution and sizing use a copied series. A replacement after manager construction can make a rule fire from the new bar while the fill is priced from the old bar.  
**Operator/downstream consequence:** a backtest can record a trade at a price that did not produce its signal—a wrong financial result. Custom strategies cannot be generically rebound to repair this after copying.  
**Evidence:** the exact-head helper still supplies a snapshot only to the manager side; `TradeOnCurrentCloseModel` prices from that supplied series, while indicator-backed rules retain their construction-time series.  
**Recommended remediation:** restore borrowed-reference semantics at execution seams and document that the series must remain stable for a run. Use a narrowly scoped `EI_EXPOSE_REP`/`EI_EXPOSE_REP2` justification: shared identity is required for semantic correctness. If deterministic frozen execution is later needed, make it a separate API that accepts a strategy factory/descriptor and constructs the strategy against the same frozen observation; accepting an already-bound `Strategy` cannot guarantee rebinding.  
**Regression proof:** construct the manager with close `20`, replace the source bar with close `100`, then run a source-bound rule that enters only above `50`. Signal observation and execution price must come from the same bar/revision.

### [Blocker] Separate stable result capture from execution ownership

**Location:** `BarSeries.snapshot()`, `BaseBarSeries.snapshotOf`, `CashFlow`, `CumulativePnL`, `Returns`, and result carriers.  
**Violated principle/invariant:** a frozen result input must be deeply immutable and provenance-coherent; an execution collaborator must preserve identity. These are different contracts.  
**What happens:** the current method copies the list but shares mutable `Bar` elements, while returning another mutable `BaseBarSeries`. It also reconstructs every implementation as a time-bar series and cannot generally preserve custom subtype or concurrency semantics.  
**Operator/downstream consequence:** source `addPrice`/`addTrade` calls can mutate the alleged snapshot without advancing its own revision, leaving cached indicators or already-computed result values stale against changed bars.  
**Recommended remediation:** remove the generic public snapshot path. Result objects that demonstrably require stable ownership should capture one package-private immutable `FrozenBarSeries`/`SeriesObservation`: deep immutable bar values, immutable logical metadata, and an atomic built-in capture where supported. Return that same safe value rather than recopying it on each accessor. For custom series, require a stable source during capture rather than promising an unenforceable universal atomic snapshot.  
**Regression proof:** capture, mutate the source through both `addPrice` and `addTrade`, and prove every captured bar/value/revision remains unchanged; mutation methods on captured bars/series must reject the operation.

### [Major] Model execution limits as a window, not as hidden series storage

**Location:** constrained `BaseBarSeries` shapes and `BarSeriesManager`'s post-range close scan.  
**Violated principle/invariant:** a series should have one obvious addressable domain; an operation's allowed range is orchestration state, not hidden raw-bar state.  
**What happens:** `getBeginIndex()`/`getEndIndex()` describe one logical window while `getBarData()` and `getBar(index)` can expose leading/trailing bars outside it. The manager deliberately executes on trailing bars, while analysis later trusts the narrower logical end.  
**Operator/downstream consequence:** an exit can be recorded at an index the result's series says does not exist, after which cash flow/returns criteria clamp or reprice it. The current unresolved trailing-exit thread is one manifestation.  
**Recommended remediation:** let the source series represent its full retained, addressable sequence, with end derived from first index plus retained size. Carry an explicit run window—conceptually `signalStart`, `signalEnd`, and `executionEnd/closeThrough`—through the manager and result. Limit strategy evaluation with that window rather than constructing a series whose raw storage disagrees with its advertised domain. A true subseries should contain only its own bars; a view should be represented as a view, not encoded as orphan storage.  
**Regression proof:** run signals through index `1`, permit closing through index `2`, and assert the result records both the signal window and effective execution range while every criterion values the actual index-2 exit price.

### [Blocker] Centralize overflow-safe absolute-index mapping

**Location:** `OffsetNumBuffer` and all inclusive index loops/successor calculations.  
**Violated principle/invariant:** disjoint ranges must remain disjoint and terminal indexes must never wrap.  
**What happens:** local `int` subtraction/clamping and repeated `+ 1` logic make each consumer rediscover `Integer.MAX_VALUE` behavior. A disjoint `MIN_VALUE..MIN_VALUE` range against a `MAX_VALUE`-offset buffer can wrap the computed last position positive and mutate the retained window. Existing open threads show the same defect family in `CashFlow` and `Returns`.  
**Operator/downstream consequence:** analysis values can be silently multiplied/added at unrelated indexes, or terminal calculations can throw/loop/misvalue trades.  
**Recommended remediation:** introduce one package-private checked absolute-range/value abstraction used by the eager analyses. Validate disjoint/inverted ranges before offset conversion; compute spans, offsets, successors, and loop counters in `long`; checked-convert only after membership is established. Let each analysis choose its documented unavailable/neutral behavior rather than making unsafe storage arithmetic silently plausible.  
**Regression proof:** exercise `MIN_VALUE` ranges against a `[MAX_VALUE - 1, MAX_VALUE]` buffer, exact terminal successors, inverted/disjoint ranges, and one/two-bar terminal windows with exact unchanged values.

## Recommended architecture

### 1. Execution uses borrowed identity

`BarSeriesManager`, strategy rules/indicators, execution models, and the synchronous sizing callback should all observe the exact same `BarSeries`. Do not clone collaborators to satisfy a value-object warning. Preserve the pre-CF-236 interaction model, add precise ownership Javadoc, and suppress only the specific SpotBugs warning with the semantic justification. Mutation during a backtest should be unsupported/documented; built-in revision tracking may additionally detect and reject a changed run rather than returning a mixed-revision result.

### 2. Windows are first-class orchestration state

Keep data availability and run policy separate. The manager already accepts start/end bounds; internalize them as one validated window and carry the effective execution end into the result. This removes the need for "trailing raw bars beyond logical end," eliminates most special scan arithmetic, and gives analysis an exact represented domain.

### 3. Frozen captures are internal immutable values

Only result/analysis objects that promise stable historical values should own a capture. That capture should not pretend to be a generally appendable `BaseBarSeries`; it should be a read-only implementation with immutable bars and the exact effective range/provenance used by the run. Built-in concurrent capture can happen under one lock. No new public API is needed for the current user journey.

### 4. Indexed analysis storage owns range arithmetic once

Replace scattered `index - begin`, `end + 1`, absolute-sized lists, and terminal loop forms with one tested package-private range mapping. `CashFlow`, `Returns`, `CumulativePnL`, and `InvestedInterval` can then focus on financial semantics—including positions entering before the retained window—rather than independently managing coordinate conversion.

## Public API inventory

| API | Existing equivalent / need | Journey role | Verdict |
| --- | --- | --- | --- |
| `BarSeries.snapshot()` | Introduced to service internal defensive-copy warnings; no demonstrated operator journey | None required for normal construction, execution, or analysis | **Remove** |
| `ConcurrentBarSeries.snapshot()` | Exists only as the override of the new method | None once internal capture owns concurrency | **Remove with interface method** |
| Internal immutable capture | No current coherent equivalent | Stable result provenance only | **Package-private, narrowly scoped** |
| Explicit run/effective range | Existing `run(start,end)` parameters already express most of it | Keeps signals, closes, and analysis coherent | **Internalize first; expose only if a user journey requires it** |

## Journey results

| Journey | Result | Friction or defect | Proof |
| --- | --- | --- | --- |
| Shortest normal | Fails semantic coherence | strategy and fills may observe different series | replace-bar counterexample |
| Advanced/ranged run | Not coherent | logical end differs from bars used to close | trailing-exit counterexample |
| Composition/custom | Cannot guarantee isolation | arbitrary custom rules cannot be rebound to a copied series | custom indicator-bound rule |
| Failure/recovery | Patch-driven | terminal and retained-window edge cases are distributed across consumers | current unresolved threads |
| Mutation/concurrency | Not detached or one-revision | shallow bars plus split series identities | `addPrice`/replace-bar probes |
| Result/analysis | Conditional/wrong | represented range may omit actual execution and retained baselines | existing open P1s |
| Discovery/API | Misleading | `snapshot()` promises a generic detached semantic it cannot uphold | public contract versus mutable bars |

## Suggested implementation sequence

1. Split the unrelated logging cleanup from this architectural repair.
2. Revert the CF-236 defensive-copy semantics at execution/callback seams; use exact shared references with explicit ownership documentation and focused SpotBugs justification.
3. Add the signal/fill same-revision regression before any further index work.
4. Represent run/effective execution ranges explicitly and remove trailing-hidden-storage dependence from the manager path.
5. Introduce one checked package-private indexed-value/range primitive and migrate eager analysis curves, including retained-window baseline semantics.
6. Add a deep immutable result capture only where stable result ownership is a demonstrated contract; do not add a generic public `BarSeries` copier.
7. Re-run the proof matrix across empty, nonzero begin, retained head, `MAX_VALUE - 1..MAX_VALUE`, replace/add-price/add-trade mutation, trailing closes, and concurrent built-in capture.

A green build confirms the patches integrate; it does not make the current ownership model coherent. The existing unresolved P1 comments remain valid local symptoms and are not duplicated as additional inline findings here.

— ChatGPT

ta4j-core/src/main/java/org/ta4j/core/backtest/BarSeriesManager.java
```diff
@@ -659,12 +666,7 @@ private ExecutionTarget fallbackSizingTarget(int index) {
     }
 
     private static BarSeries snapshotSeries(BarSeries barSeries) {
-        BarSeries series = Objects.requireNonNull(barSeries, "barSeries");
-        return new BaseBarSeriesBuilder().withName(series.getName())
-                .withNumFactory(series.numFactory())
-                .withBars(series.getBarData())
-                .withMaxBarCount(series.getMaximumBarCount())
-                .build();
+        return barSeries.snapshot();
```
TheCookieLab: [Blocker] Stop copying only the series side of the strategy graph.

`Strategy` and its indicators remain bound to the caller's original `BarSeries`; this helper gives execution and sizing a different series instance. After constructing the manager, replace the source's last bar. An indicator-backed rule can see the replacement and enter, while `TradeOnCurrentCloseModel` prices the fill from the stale copy. That is a wrong financial result. The behavior is additionally mutation-dependent because list replacement is isolated while in-place `Bar` mutation is shared.

Restore borrowed/reference semantics here (with a precise SpotBugs representation-exposure justification), or introduce a separate frozen-run API that captures the series and builds/rebinds the strategy against that same capture. Add a regression where source close `20` is replaced by `100` between manager construction and `run()`: signal and fill must observe the same close/revision.

ta4j-core/src/main/java/org/ta4j/core/BaseBarSeries.java
```diff
@@ -111,6 +111,43 @@ public BaseBarSeries(final String name, final List<Bar> bars) {
...
+     *
+     * @param series the series to copy; must not be {@code null}
+     * @return a new {@link BaseBarSeries} with the same logical view and raw data
+     * @throws NullPointerException if {@code series} is {@code null}
+     * @see BarSeries#snapshot()
+     * @since 0.24.2
+     */
+    static BaseBarSeries snapshotOf(final BarSeries series) {
+        Objects.requireNonNull(series, "series");
+        BaseBarSeries copy = new BaseBarSeries(series.getName(), series.getBarData(), series.getBeginIndex(),
```
TheCookieLab: [Blocker] This is not a detached snapshot.

The list is copied, but its `Bar` elements are not; `Bar.addPrice` and `Bar.addTrade` mutate those shared instances. After the concurrent source lock is released, a source bar mutation changes this copy without advancing the copy's own history revision. Cached indicators and result objects can therefore remain stale while their underlying prices change.

If a frozen capture remains, deep-copy each bar's observable fields into immutable/read-only bar values under the same coherent observation and prove `addPrice`/`addTrade` isolation. My recommendation is to remove the generic snapshot path and reserve deep immutable capture for explicit result/value ownership.

ta4j-core/src/main/java/org/ta4j/core/BarSeries.java
```diff
@@ -203,6 +203,24 @@ default void clear() {
...
+     * The snapshot is rebuilt from this series's public accessors, so subclasses
+     * that override {@link BarSeries} accessors to advertise a shifted logical view
+     * are reproduced exactly. Concurrent implementations override this method to
+     * observe the source atomically under their read lock.
+     * </p>
+     *
+     * @return a new {@link BaseBarSeries} with the same logical view and raw data
+     * @since 0.24.2
+     */
+    default BarSeries snapshot() {
```
TheCookieLab: [Major] Keep this internal ownership workaround out of the public API.

A default `BarSeries.snapshot()` cannot uphold the documented contract for arbitrary implementations: it cannot guarantee atomicity, deep immutability, custom builder/subtype semantics, or one coherent revision. The current fallback also reconstructs everything as `BaseBarSeries` with a time-bar builder. This permanently expands the primary domain interface without improving a normal user journey.

Remove this method and use a package-private immutable capture only in result paths that truly need stable ownership. Execution paths should preserve the original series identity instead of copying it.

ta4j-core/src/main/java/org/ta4j/core/analysis/OffsetNumBuffer.java
```diff
@@ -0,0 +1,184 @@
...
+     * Adds the given delta over the absolute index range, clamped to the window.
+     * Ranges entirely outside the window are ignored.
+     *
+     * @param fromIndex the first absolute index of the range
+     * @param toIndex   the last absolute index of the range
+     * @param delta     the delta to add
+     */
+    void addRange(int fromIndex, int toIndex, Num delta) {
+        int firstPosition = Math.max(fromIndex, startIndex) - startIndex;
+        int lastPosition = Math.min(toIndex, endIndex) - startIndex;
```
TheCookieLab: [Blocker] Reject disjoint ranges before converting them to offsets.

For a buffer covering `[Integer.MAX_VALUE - 1, Integer.MAX_VALUE]`, `addRange(Integer.MIN_VALUE, Integer.MIN_VALUE, delta)` computes `firstPosition == 0` and wraps `lastPosition` to `2`; the loop then mutates both retained slots even though the requested range is entirely disjoint. `multiplyRange` has the same defect.

Check inverted/disjoint ranges first, then compute clamped offsets in `long` and checked-convert only after membership is established. Add exact unchanged-value regressions for `MIN_VALUE..MIN_VALUE` against a terminal-offset buffer.

ta4j-core/src/main/java/org/ta4j/core/backtest/BarSeriesManager.java
```diff
@@ -146,14 +154,17 @@ public BarSeriesManager(BarSeries barSeries, CostModel transactionCostModel, Cos
...
+            + "copying desynchronized signal and fill pricing.")
     public BarSeriesManager(BarSeries barSeries, CostModel transactionCostModel, CostModel holdingCostModel,
             TradeExecutionModel tradeExecutionModel, TradingRecordFactory tradingRecordFactory) {
         Objects.requireNonNull(barSeries, "barSeries");
         Objects.requireNonNull(transactionCostModel, "transactionCostModel");
         Objects.requireNonNull(holdingCostModel, "holdingCostModel");
         Objects.requireNonNull(tradeExecutionModel, "tradeExecutionModel");
         Objects.requireNonNull(tradingRecordFactory, "tradingRecordFactory");
-        this.barSeries = snapshotSeries(barSeries);
+        this.barSeries = barSeries;
```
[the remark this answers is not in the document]
TheCookieLab: Revalidated against head `09bdcb5aecdb042f5098718bb7404a76865180f0`: this remains open. `BarSeriesManager` now correctly borrows the strategy's series, but the private run loop is not enclosed by `ConcurrentBarSeries.withReadLock(...)`; each individual accessor locks and unlocks independently. A bounded append can still evict index `i` after `strategy.shouldOperate(i, ...)` and before the execution model prices index `i`, at which point legacy `getBar(i)` aliases the new first retained bar. The smallest safe fix is a run-scoped read lease/lock; if ingestion must remain nonblocking, use a versioned immutable read lease rather than globally archived bars. Add a latch-controlled signal→append→fill regression.

— ChatGPT

ta4j-core/src/main/java/org/ta4j/core/analysis/CashFlow.java
```diff
@@ -60,7 +57,8 @@ public class CashFlow implements PerformanceIndicator {
      */
     public CashFlow(BarSeries barSeries, TradingRecord tradingRecord, int finalIndex, EquityCurveMode equityCurveMode,
             OpenPositionHandling openPositionHandling) {
-        this(barSeries, tradingRecord, 0, barSeries.getEndIndex(), finalIndex, equityCurveMode, openPositionHandling);
+        this(barSeries, tradingRecord, 0, Math.max(barSeries.getEndIndex(), tradingRecord.getEndIndex(barSeries)),
```
TheCookieLab, line 60: [Blocker] Anchor default materialization at the retained series head.

This still passes literal `0` as the buffer start. A three-bar rolling series at indexes `1_000_000_000..1_000_000_002` therefore asks `OffsetNumBuffer` to allocate roughly one billion slots, defeating the rolling-series memory bound and likely exhausting the heap. The terminal test at `Integer.MAX_VALUE` does not prove compactness because the current oversized-span fallback creates an empty buffer and `getValue()` returns neutral one.

Use `max(requestedStart, barSeries.getBeginIndex())` as the effective materialized start (and pass `getBeginIndex()` on the normal path). Add a non-neutral buy/sell regression at a large but nonterminal offset and prove the materialized size is the retained/trailing window size, not the absolute end index.

ta4j-core/src/main/java/org/ta4j/core/analysis/OffsetNumBuffer.java
```diff
@@ -0,0 +1,212 @@
...
+    static OffsetNumBuffer of(BarSeries series, Num initialValue, Num neutral) {
+        return new OffsetNumBuffer(series.getBeginIndex(), series.getEndIndex(), initialValue, neutral);
+    }
+
+    private static int sizeOf(int startIndex, int endIndex) {
+        if (startIndex < 0 || endIndex < startIndex) {
+            return 0;
+        }
+        long span = (long) endIndex - startIndex + 1;
+        return span > Integer.MAX_VALUE ? 0 : (int) span;
```
TheCookieLab, line 75: [Blocker] Fail loud instead of turning an impossible span into an empty neutral curve.

For `[0, Integer.MAX_VALUE]`, this returns size zero. `CashFlow`, `CumulativePnL`, or `Returns` can then return plausible neutral values even though none of the requested domain was materialized; this is why terminal-offset tests can pass without testing a stored value.

After callers clamp to the retained/materialized start, such a span should be unreachable. Validate the invariant and throw a specific exception (or use checked span conversion) rather than silently returning zero. Add an exact constructor/range regression asserting rejection.

ta4j-core/src/main/java/org/ta4j/core/backtest/BacktestExecutionResult.java
```diff
@@ -39,8 +38,10 @@ public record BacktestExecutionResult(BarSeries barSeries, List<TradingStatement
      * @param runtimeReport     runtime statistics for the execution
      * @param strategyFailures  strategies skipped because execution failed
      */
+    @SuppressFBWarnings(value = "EI_EXPOSE_REP2", justification = "The result borrows the caller's live series so "
```
TheCookieLab, line 41: [Blocker] Separate live execution ownership from completed-result ownership.

Borrowing the exact series is correct during execution, but this result evaluates arbitrary criteria later in `getTopStrategies(...)`. If the rolling series advances or a retained historical bar is replaced after the run, recorded fills remain from revision R while drawdown/returns criteria read revision R+1. The result is no longer a coherent account of one backtest, and the current justification's “same instance” does not imply the same revision.

At the result boundary, either compute all requested criteria eagerly or retain one immutable/versioned run-data lease shared by every statement. Keep this internal; it does not require restoring the public generic snapshot API. Regression proof: execute a position spanning a middle bar, capture its score, replace that middle bar/advance retention, and prove the completed result's ranking and score remain unchanged (with an explicit separate revaluation path if desired).

ta4j-core/src/main/java/org/ta4j/core/analysis/Returns.java
```diff
@@ -264,15 +290,37 @@ public int getCountOfUnstableBars() {
...
+     * leading no-prior-close placeholder. This is the case when an entry predating
+     * the window seeds the whole entry-to-first-retained-close move, or when a
+     * position enters and exits on the first retained bar itself. Tail-risk
+     * criteria use this distinction to decide whether the leading slot joins the
+     * return distribution.
+     *
+     * @return {@code true} if the first retained slot holds a real return
+     * @since 0.24.2
+     */
+    public boolean isFirstSlotSeeded() {
```
TheCookieLab, line 322: [Major] Keep placeholder bookkeeping out of the public API.

`isFirstSlotSeeded()` exposes an internal buffer-layout decision solely so core criteria can decide whether to skip slot zero. It does not improve a demonstrated external `Returns` user journey and becomes permanent API debt.

Internalize sample extraction instead—for example, a package-private materialized-return view/helper, or derive availability from the first raw value and established `NaN` convention. Then remove this public method before release. Add criterion tests for seeded, placeholder, and genuinely unavailable first values.

TheCookieLab (commented):
## Follow-up review

### Coverage

- **Profile verdict:** `applied`
- **Exact target:** `ta4j/ta4j` base `e4340715c7b4a098993cecc5e21fa53ef56ebfc4` → head `09bdcb5aecdb042f5098718bb7404a76865180f0`
- **Scope:** all 37 changed files, current review threads, the execution/result ownership boundary, rolling/constrained index semantics, eager performance analyses, and tail/drawdown consumers.
- **Verification:** all hosted checks on the exact head are green. I could not run an independent local checkout, so the findings below are based on exact-head source and deterministic counterexamples.

### Executive verdict

- **Correctness:** not ready
- **Production readiness:** not ready
- **API architecture:** needs consolidation
- **Compatibility:** behavioral migration still needs definition

The major previous concern was fixed correctly: the temporary public `BarSeries.snapshot()` API is gone, and execution/sizing now use the same series identity as the strategy. The overflow-safe range work is also materially better. Four issues remain: the concurrent run is still not revision-stable (existing thread reopened), default `CashFlow` can still allocate by absolute index, impossible buffer spans silently become plausible neutral curves, and completed backtest results still depend on future mutations of a borrowed live series. I also flagged one unnecessary new public method.

## Foundational assessment

`OffsetNumBuffer` is **not archiving rolled-off bars**. It stores one derived `Num` per materialized cash-flow/PnL/return point. Once its start index is correct, its memory is `O(materialized window)`, not `O(absolute index)` and not a second copy of OHLCV history. Some compact indexed-value representation is unavoidable whenever the caller asks for an exact random-access curve.

Keeping old `Bar` objects “archived” inside the series while expecting the garbage collector to reclaim them will not work: a strong reference keeps them alive. Weak references would make domain results depend on GC timing. Keeping null/tombstone slots forever avoids retaining bars but leaks pointer/index storage and eventually hits the same absolute-index problem.

The sound version of that idea is:

1. **Active storage:** use a circular/segmented bar buffer with a monotonic first-retained index. Eviction clears or overwrites the old slot, so the old `Bar` becomes GC-eligible immediately. Appends and evictions become `O(1)` instead of shifting an `ArrayList` head.
2. **Unavailable history:** retain only a range tombstone (`firstRetainedIndex`), not archived bar objects. Core reads for an evicted index should eventually fail as unavailable rather than aliasing the first retained bar. The legacy aliasing is what makes concurrent eviction capable of recording index `i` at index `i+1`'s price.
3. **Evaluation range:** move signal/end/close-through bounds out of hidden raw storage and into an explicit run window. A `BarSeries` should have one obvious addressable domain; `addressableEndIndex()` and trailing raw bars beyond the logical end are symptoms of conflating storage with orchestration.
4. **Stable reader lifetime:** immediately, hold `ConcurrentBarSeries.withReadLock(...)` for a whole run (and ideally a whole multi-strategy batch). If ingestion must never block, the scalable alternative is an internal versioned read lease over immutable chunks: an in-flight reader pins the old chunk version, the writer publishes a new version, and the old chunks become GC-eligible when the lease closes. That—not a permanent archive—is the GC-friendly interpretation of retaining bars temporarily.
5. **Optional history archive:** exact reanalysis after eviction requires either the original bars or sufficient derived state. Put long-term bars in an explicit disk/database/off-heap `BarStore`, loaded deliberately for backtests. Do not make normal live `getBar()` transparently consult archival storage.
6. **Analysis output:** keep a small package-private indexed-value primitive (`startIndex + values[]`) for exact curves, but use streaming reducers for scalar metrics. Cumulative PnL, peaks/max drawdown, moments, and many ratios can be maintained incrementally; exact curves/quantiles inherently require retained values or external persistence. Share one materialized equity/return curve across criteria rather than creating parallel arrays per criterion.

The clean dependency order is therefore: fix the remaining blockers in this PR; then follow with a focused active-storage/run-window refactor. Do not fold a persistent archive or another public snapshot API into this already broad repair.

The PR title and description still describe the deleted snapshot implementation and should be rewritten before merge.

— ChatGPT

TheCookieLab:
@coderabbitai review

TheCookieLab:
@coderabbitai review

TheCookieLab:
@coderabbitai review

TheCookieLab:
@coderabbitai full review

ta4j-core/src/main/java/org/ta4j/core/backtest/BarSeriesManager.java
```diff
@@ -245,6 +259,10 @@ public TradingRecord run(Strategy strategy, TradeType tradeType, int startIndex,
      * @return the trading record coming from the run
      */
     public TradingRecord run(Strategy strategy, TradeType tradeType, Num amount) {
+        if (barSeries instanceof ConcurrentBarSeries concurrentBarSeries) {
+            return concurrentBarSeries.withReadLock(
+                    () -> run(strategy, tradeType, amount, barSeries.getBeginIndex(), barSeries.getEndIndex()));
```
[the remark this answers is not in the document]
TheCookieLab: I am consolidating the manager-level lease delegation and preserving read scopes through indicator views. The non-reentrant-lock premise is not part of the supported external ConcurrentBarSeries API: custom ReadWriteLock injection is package-private test infrastructure; every public constructor uses ReentrantReadWriteLock. Reentrancy also predates this change (for example, getSubSeries holds its read lock while calling locked getName, and withReadLock has existed since 0.22.2). The shared BarSeries read-scope contract now states same-thread nesting explicitly. I will not introduce a generalized lock adapter or change production locking semantics to support an otherwise unsupported test lock.

TheCookieLab:
@coderabbitai full review

Please re-evaluate head 3a3fa5272: read-only views now forward the common read lease and concurrency capability; manager overloads acquire one lease; invested intervals use addressable raw bounds. The corresponding outdated threads are resolved after regression and full-gate verification.

TheCookieLab:
@coderabbitai review

Please update the disposition for head 12009abb5. Logical benchmark/mark-to-market ends are restored; only offset analyses derive raw trailing activity bounds. Manager, walk-forward and batch writer regressions use the same pre-lock observation seam, and the unused buffer factory is removed. All five matching threads are resolved and the final full gate passed 10,004 tests.

TheCookieLab:
@coderabbitai review Please evaluate head aaf35b0b9c81e27820f6a04f89ad00d43eca2232. The captured trailing-return and coherent live-facade findings are fixed; their regressions fail on the previous implementation and pass on this head. Full gate: 10,023 tests, 0 failures/errors. Please refresh the canonical summary from the earlier head.

